PR #429, opened 2022-11-01
Initial code dump from transition-legacy

tahini:
This brings to this repo the codes from the 6 typescript packages, for chaire-lib and transition. The application does not work yet, but `compile`, `format`, `lint` and `test` do. Changes are minimal from their current state in transition-legacy, only to make the afore-mentioned scripts work. All contributors to the original code base are added as co-authors to the initial commit.

This is one of the first step of the Transition repo migration. See https://github.com/chairemobilite/transition-legacy/issues/1943 for all steps

Co-authored-by: @kaligrafy 
Co-authored-by: @tahini
Co-authored by: @jufaua
Co-authored-by: @alexis-viallard
Co-authored-by: @vincentchab
Co-authored-by: @khadidjaSeridi
Co-authored-by: @greenscientist 
Co-authored-by: @amDeimos666
Co-authored-by: @YasmineMoumou


kaligrafy (approved):

tahini:
@greenscientist  How to review this PR: Look at the general directory/file structure. The code in the first commit has not been changed from their original version in private repo. The second commit shows what needed to be changed to make the main commands work. It can be reviewed independently if one wishes.

packages/chaire-lib-backend/src/utils/processManagers/OSRMServicePreparation.ts
```diff
@@ -12,12 +12,14 @@ import { fileManager } from '../filesystem/fileManager';
 import { getOsrmDirectoryPathForMode, defaultDirectoryPrefix, getDirectoryPrefix } from './OSRMServicePath';
 import { RoutingMode } from 'chaire-lib-common/lib/config/routingModes';
 
-const defaultImportOsmFilePath = fileManager.directoryManager.getAbsolutePath('imports/network.osm');
```
greenscientist: Is it moved somewhere else?

greenscientist (approved):
Only small comments/questions

I don't have much opinion on the directory structure. Other than we have a lot of small ones. I wonder if we could (someday, not now) consolidate a little bit. I feel this make it a bit hard to find where the code should be.

.env.example
```diff
@@ -0,0 +1,62 @@
...
+MAIL_TRANSPORT_SMTP_HOST=smtp.example.org
+MAIL_TRANSPORT_SMTP_PORT=587
+# Whether to use TLS 
+MAIL_TRANSPORT_SMTP_SECURE=false
+# SMTP server credentials
+MAIL_TRANSPORT_SMTP_AUTH_USER=myUser
+MAIL_TRANSPORT_SMTP_AUTH_PWD=password
+
+# From email
+MAIL_FROM_ADDRESS=admin@transition.city
```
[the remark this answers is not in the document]
tahini: good point, this is public now!

packages/chaire-lib-backend/src/utils/processManagers/OSRMServicePreparation.ts
tahini: No, it's gone, that was not even a real default (maybe once, but not anymore, it's not the name the tasks give it).

tahini:
> I don't have much opinion on the directory structure. Other than we have a lot of small ones. I wonder if we could (someday, not now) consolidate a little bit. I feel this make it a bit hard to find where the code should be.

I was talking about the high level repository directory structure, not inside the packages (I won't change that here, or anytime soon for that matter), but yeah, I took example with other react applications, but between services and utils, sometimes the difference is minimal. Anyway...

PR #432, opened 2022-11-03
Make the transition client and server work

tahini:


tahini:
Based on #429 so still a draft

examples/config.js
```diff
@@ -0,0 +1,134 @@
+module.exports = {
```
greenscientist, line 1: We so many copies of this or part of this in the tests. Maybe we should use something to generate all of these automatically instead of relying on copy paste. (When we need to change some default stuff)

greenscientist:
Looks fine to me so far

examples/config.js
tahini: You may have been referring to something that was in the previous PR, because I don't see many copies here. But yeah, for unit tests, there are 2 configuration files, one used throughout unit tests, the other to test with different configuration values. That second one could be mocked instead. But the purpose is to test the code path with the file.

This file is the only "real" (not test) configuration file

tahini: You may have been referring to something that was in the previous PR, because I don't see many copies here. But yeah, for unit tests, there are 2 configuration files, one used throughout unit tests, the other to test with different configuration values. That second one could be mocked instead. But the purpose is to test the code path with the file.

This file is the only "real" (not test) configuration file

greenscientist: Yes, I meant in general, not necessarly in this PR

examples/README.md
```diff
@@ -0,0 +1,36 @@
+This folder contains an example configuration file, as well as necessary files to run a demo of Transition. This demo corresponds to the territory of the Réseau de transport de Longueuil area.
```
greenscientist, line 1: maybe add the URL for the RTL, to give more contexte to people (https://www.rtl-longueuil.qc.ca/)

examples/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+# Running the demo
+
+1. Make sure all the pre-requisites are installed. For complete installation instructions of all dependencies, see the [instructions for Ubuntu](../../docs/transition/installationUbuntu20.04.md).
```
greenscientist, line 5: Not for now, but we'll need to update this. This is a bit of an outdated ubuntu version. We might need more generic instructions

examples/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+2. Update the .env file and set the `PROJECT_CONFIG` to point to the config file in this directory
```
greenscientist, line 7: Update or copy the .env.example file??

greenscientist: Maybe say "Or set the environment variable..." ????
or is this too much details/options?

examples/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+5. Get and prepare the road network for `osrm` to route. This step is optional, but required to create new lines that properly follow the road network. The first line will download the Open Street Map network data from the overpass API. The second line will prepare the data for the `osrm` servers. Data is prepared differently for different modes of tranportation. Selecting `driving` and `walking` are mandatory modes, as `driving` is the default mode for vehicles and `walking` is required to calculate access, transfer and egress times from transit.
+
+```shell
+yarn babel-node --max-old-space-size=4096 transition-app/chaire-lib/backend/lib/scripts/osrm/downloadOsmNetworkData.task.js --polygon-file examples/transition/polygon_rtl_area.geojson
+
+yarn babel-node --max-old-space-size=4096 transition-app/chaire-lib/backend/lib/scripts/osrm/prepareOsmNetworkData.task.js
+```
+
+6. Start the nodejs server with `yarn start:transition`.
```
greenscientist, line 25: unsure, but was this changed to remove the transition part?

examples/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+7. Navigate to `http://localhost:8080` and log into the application.
+
+8. Import [transit data for the Réseau de Transport de Longueuil](https://transitfeeds.com/p/reseau-de-transport-de-longueuil/37), that will work for the area of this demo. Lines and paths can be edited, added within this territory.
+
+# Running the application for another territory
+
+Copy or edit the config.js file and update for the area. The `projectDirectory` and `mapDefaultCenter` are the main fields to update. Be sure to point the `PROJECT_CONFIG` environment variable to this new file.
```
greenscientist, line 33: Why is projectDirectory implicated in changing the geographical region?

greenscientist (commented):
I had several comments about the doc, but other code looks ok. 

maybe @kaligrafy  want to have a look

examples/README.md
tahini: I assume they followed the previous instructions in their current state, so they have a .env at this point. I haven't tested with just setting environment variable, and had bad luck when trying with `yarn start:json2capnp`, so I wouldn't say it here. I know that's what you do in terraform. But if you're reading this, you want it as simple as possible.

examples/README.md
tahini: good catch, indeed it was

examples/README.md
tahini: I'ts where the local osrm data will be stored by default, so if you don't want to overwrite previous data, you need to change the path. But that's only if you plan to support many regions... But at least, if we tell them to change the project directory and they forgot to download/prepare the osm network, it will tell them the servers are not running instead of running on the wrong region.

greenscientist: PRobably just add why we need to change the project directory.

greenscientist (approved):
Probably ok :)

PR #434, opened 2022-11-03
Copy the json2capnp service from legacy

tahini:
Note that the cache path directory must now be part of the command line since it can't be deduced from .env anymore.

Depends on #432 so still a draft


greenscientist:
Looks fine

greenscientist:
There's no change to other part of the documentation with the command line change?

tahini:
> There's no change to other part of the documentation with the command line change?

There's no documentation in this repo yet ;-)

greenscientist (approved):
Ok, Let's not forget to update it before comitting it then

PR #436, opened 2022-11-03
Add the Dockerfiles for this repo

tahini:
Add a Dockerfile and an example docker-compose


tahini:
Based on all the other ones before. @greenscientist  if you have the opportunity to try this out and tell me if it seems to work for your setup. I tried and it builds and runs the docker build, but I'm not sure it completely works, there's probably something missing re the .env files and the path for json2capnp. Or may it just works too...

greenscientist:
Yeah, I'm wondering if the default config and stuff like that that we used to have still work fine. I'll do a test run and see.

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+    ports:
+      - 8080:8080
+    environment:
+      PG_CONNECTION_STRING_PREFIX: postgres://postgres:pass@postgres:5432/
+    depends_on:
+      - postgres
+    #HACK, needed for running trRouting. See https://github.com/kaligrafy/trRouting/issues/23
+    tty: true
+    volumes:
+      # Fine-tune these paths to the target container's runtime directories
```
greenscientist, line 27: We need to change those directory to something that make sense now. The example directory maybe?

greenscientist (changes requested):

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+#From OSRM dockerfile, make sure tools work
+RUN /usr/local/bin/osrm-extract --help && \
+    /usr/local/bin/osrm-routed --help && \
+    /usr/local/bin/osrm-contract --help && \
+    /usr/local/bin/osrm-partition --help && \
+    /usr/local/bin/osrm-customize --help
+
+# Start json2capnp -> Relies on manually creating cache directory before
+# Start Node app
+CMD sh -c "cd services/json2capnp && pwd && ./json2capnp -- 2000 /app/examples/runtime/cache/demo_transition > /app/json2capnp.log &" && yarn build:prod && yarn start
```
greenscientist, line 64: I don't think we need the -- here, they are for cargo run, not ruinning the binary directly

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+#RUN yarn compile && yarn build:dev
+RUN yarn compile
+#RUN yarn compile && yarn build:prod
+
+#TODO We probably need to do something different for the projects configuration directories
+# the docker-compose file have an example of using volume for part of a project
+
+
+# Copy in json2capnp
+COPY --from=json2capnpbuild /app/services/json2capnp/target/debug/json2capnp services/json2capnp/
```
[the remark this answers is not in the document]
greenscientist: Because we build  the debug build.

greenscientist:
It kinda work after addng
COPY .env.example /app/.env
to the docker to have some basic default. 
But I think it's missing some other good default values to have something really working.

greenscientist:
So, I'll like to have some default env set so it can at least run the example without more setup. 
We can either: 
1- Copy the .env.example file
or 
2- Set some of the environnement variables individually in the Dockerfile. 

Any option you'd think is better?

tahini:
> So, I'll like to have some default env set so it can at least run the example without more setup. We can either: 1- Copy the .env.example file or 2- Set some of the environnement variables individually in the Dockerfile.
> 
> Any option you'd think is better?

iiuc, the best option is to have a config file in some predefined path and put all configs there. But we are not there yet, though I'd like to prioritize that soon. In the meantime, we could copy the .env.example (or .env if it exists?) I guess, but really, I have no preference...

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+#RUN yarn setup
+#RUN yarn setup && yarn migrate && yarn create-user
+#RUN yarn compile && yarn build:dev
+RUN yarn compile
+#RUN yarn compile && yarn build:prod
+
+#TODO We probably need to do something different for the projects configuration directories
+# the docker-compose file have an example of using volume for part of a project
+
+
```
greenscientist, line 43:
```suggestion
# Setup the example as a default configuration for the image
COPY .env.example /app/.env

```

greenscientist:
ok, I'have added the suggestion change, tested here.

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  postgres:
+    image: postgis/postgis
+    volumes:
+      - transition-postgis-data:/var/lib/postgresql/data
+    environment:
+      POSTGRES_PASSWORD: pass
+
+volumes:
+  transition-postgis-data:
```
tahini, line 41: If I want to use this docker-compose file, would I need to put the local paths in here? What happens with this file here if we put nothing? How does it know what to mount by default?

greenscientist: No, those will be automatically created by docker in its "local volume" repo. (Usually something like /var/lib/docker/volumes)

greenscientist (approved):
Should be ok. 

(I can imagine a lot more improvements, but that's a good base)

tahini:
> (I can imagine a lot more improvements, but that's a good base)

@greenscientist Dockerfiles are your territory! Please open issues with improvement suggestions.

Issue #442, opened 2022-11-08
How to manage our database migrations?


tahini:
In the legacy, there was one migration directory, which contains tables for transition, evolution, osc2ar and octavi. The easy migration path is to just take it all in this repo now. But we have tables for evolution, osc2ar and octavi that are not necessary.

But if we only migrate related migrations, we get errors in current databases for the missing migrations.

Also see this https://stackoverflow.com/questions/57174914/knex-js-can-i-have-multiple-migrations-to-save-in-only-one-migration-table 

We could have multiple migration tables for each project: chaire-lib, transition, evolution and octavi/osc2ar migrations. Or the post suggests that each project could have its own database (or schema) with their own migrations. That means chaire-lib would be responsible for the users table. Transition and Evolution are quite different, but osc2ar and octavi need some transition stuff and they are more or less related.

So what do we do?

1. Single database, multiple migrations tables
2. Multiple databases for the application

Maintenance-wise, would 2 be more complicated? By the time a single application requires data from more than one app at a time, we probably will have some micro-service architecture to get data from the appropriate location...

cc @kaligrafy @greenscientist  thoughts?

greenscientist:
I think we should consider this new repo as a new app and drop all the old migration script and start new. 

If people have been running a legacy app, they would need to upgrade with the legacy app and then migrate to the public app. (We don't have many internal users, so we can deal with that easily.

tahini:
I'm not sure how I went from A to B, but I think your comment implies the 1- approach above (multiple migration tables (one per package) in a single database _if_ the "web application" contains more than the). We add one migration, which adds the new migration table, adds the migrations, then drops the old table...

greenscientist:
btw, yeah, multiple database would be awful. 
My thought: each application should be responsible for their migration. They could import something from chaire-lib to do the "common" bits. But some how the responsabilités will reside in the app.

PR #443, opened 2022-11-08
Make setup and migrate work

tahini:


package.json
```diff
@@ -10,7 +10,6 @@
         "packages/transition-frontend"
     ],
     "scripts": {
-        "create-project": "yarn workspace transition-legacy run create-project",
```
greenscientist, line 13: Is there code to remove as well or this was broken already?

tahini:
Tested with a new database and it works

greenscientist (commented):
You can see my comment in #442, I don't think we should have all the legacy migration script in the new repo. 
Maybe we add them and remove them after, but it's old stuff that we don't need to maybe, as the app don't exist publicly.

tahini:
I agree it should not all be here. But we need to rethink our architecture for the multiple app support (evolution is quite independent, but osc2ar and octavi use some of the transition concepts). So either multiple migration tables or multiple databases/schemas.

To ease the transition from the legacy repo though, the easiest way is this PR, then have transition-legacy support the external chaire-lib/transition, _then_ fix #442, in which case, the current migrations will be removed and replaced with something else

greenscientist:
We don't have to care about octavi and oscar. They don't exist in the world here. They are legacy only app. WE don't need to support them here.

tahini:
I was hoping to postpone fixing #442 to later when we have the new infra up... but ok, I can do it now.

tahini:
@greenscientist So these are brand new migrations with the table creation. chaire-lib-backend and transition-backend both have migration tables so they are all independent. chaire-lib's must be run before transition-backend.

Tested both on new project and migration from transition-legacy.

packages/chaire-lib-backend/src/scripts/config/setup.task.ts
```diff
@@ -8,11 +8,10 @@ import 'chaire-lib-common/lib/config/shared/dotenv.config'; // Unused, but must
 import config from '../../config/server.config';
 import knex from 'knex';
 
-// These files should not be published, they are for application preparation and should never be called by actual app
```
greenscientist, line 11: Reasking here, why is this comment not applicable anymore? Just curious

greenscientist (approved):
Sounds good

PR #453, opened 2022-11-14
chaire-lib: Make setup work and remove create-project

tahini:
The create-project command is not required anymore in this repo, one just has to provide a configuration file in the .env

Also remove comments and do proper import of knexfiles in the setup task.


tahini:
@greenscientist This is the commit from #443 that is not migration-related. I'll do the migrations separately.

packages/chaire-lib-backend/src/scripts/config/setup.task.ts
```diff
@@ -8,11 +8,10 @@ import 'chaire-lib-common/lib/config/shared/dotenv.config'; // Unused, but must
 import config from '../../config/server.config';
 import knex from 'knex';
 
-// These files should not be published, they are for application preparation and should never be called by actual app
```
greenscientist, line 11: What about this comment that is removed?

Issue #459, opened 2022-11-17
.env and rootDir are only good when run from this repo


tahini:
They use relative paths that only apply to the current repository. When run as a submodule in transition-legacy for instance, these paths are wrong: the desired .env file does not exist and the rootDir is not the expected one.

It's time to fix #425, with a default relative path (for applications running from this repo and docker) and the possibility to set this path in an environment variable (but not in .env).

And the configuration options set in the .env should now be part of the config.js file.

Issue #452, opened 2022-11-14
Have sane default to allow the app to run without a .env file


greenscientist:
We should be able to run an application with only the right settings in a config file.

tahini:
I wanted to work on this, to put the .env content in a config file, but #97 will have to be solved first. There are some sensible information, like database password, that should not be exposed, and right now, the whole config file content goes to the client.

Issue #97, opened 2022-04-06
chaire-lib-common: Redesign preferences


kaligrafy:
- Preferences should be separated by modules
- User preferences saved in db should only store diff with default one
- Clarify the difference between pref (user pref) and config (server/project config)

PR #461, opened 2022-11-18
config: Add default paths for project config file

### LINKED ISSUE #425: chaire-lib: Add a default path for the project config file
The PROJECT_CONFIG environment variable needs to point to the configuration file. We should have a default path where to look for that file if it is not set, for example first CWD, then maybe $HOME/.config/transition/ then /etc/transition.

tahini:
Fixes #425

First, we check if the PROJECT_CONFIG environment variable contains the
path to the config file. Otherwise, it is expected to be in the
$homedir/.config/transition/config.js or /etc/transition/config.js.

The path in the .env now has to be absolute. ${rootDir} works only in
the current repository, but chaire-lib is meant as a library, so rootDir
will not work for those.

Add a start:demo script in the package.json which sets the
PROJECT_CONFIG environment variable to the example config.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,4 @@
-PROJECT_CONFIG=${rootDir}/examples/config.js
+PROJECT_CONFIG=/absolute/path/to/transition/config.js
 #OSRM_DIRECTORY_PREFIX=demo_osrm (default: empty)
 HOST=http://localhost:8080
 PG_DATABASE_DEVELOPMENT=tr_dev
--- a/.github/workflows/transition.yml
+++ b/.github/workflows/transition.yml
@@ -15,6 +15,8 @@ jobs:
     strategy:
       matrix:
         node-version: [16.x]
+    env:
+      PROJECT_CONFIG: ${{ github.workspace }}/examples/config.js
     steps:
     - uses: actions/checkout@v2
     - name: copy env file
--- a/examples/README.md
+++ b/examples/README.md
@@ -7,7 +7,7 @@ This folder contains an example configuration file, as well as necessary files t
 2. Update the .env file and set the `PROJECT_CONFIG` to point to the config file in this directory
 
 ```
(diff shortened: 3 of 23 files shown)


kaligrafy (approved):

tahini:
I added the fix of #459 to this PR

TLDR:

1. The PROJECT_CONFIG environment variable can be set, either in .env or some other way, but if not set, it will look in $HOME/.config/transition/config.js or /etc/transition/config.js
2. The TRANSITION_DOTENV environment variable can be set, defaults to <this repo's root>/.env. Thus it does not need to be set in this repo, but in transition-legacy yes (tested and it works now)

The doc has not been updated yet, I'll re-work it in the next PRs, with complete startup documentation.

@greenscientist Can you take a look with deployment and infra in mind? (knowing #452 and #97 still need fixing)

greenscientist:
One quick comment. PROJECT_CONFIG path being absolute is fine, but one alternative way is to treat a string starting with a "/" as absolute and one without it as relateive to the CWD. But we can always replicate the later by setting the string to $(CWD)/something. 

For TRANSITION_DOTENV, " defaults to <this repo's root>/.env. " what does happen when you run the code from outside a git repo, like a TGZ that packaged everything? Or a docker image? (Probably need an answer to this before commiting)

tahini:
I was expecting that any repo NOT running from here would require those environment variables, but true, I could use node's provided envirionment variables for CWD. I'll test that it's the correct one in every case and update if so

tahini:
So... CWD and PWD happen to be packages/chaire-lib-backend because the `yarn workspace chaire-lib-backend` apparently chagned the working directory. But INIT_CWD is the directory from which the original command was run, ie the root of the repos that interest us. So the default relative paths are now set to INIT_CWD. It works both with git submodule in transition-legacy and here.

greenscientist (approved):
ok, sounds good

PR #466, opened 2022-11-22
Add lib/*.lua files for the osrmProfiles

tahini:
They were not noticed as their path is in the .gitignore


greenscientist:
Can you also fix the .gitignore or you'll create an issue for it?

PR #471, opened 2022-11-28
Use node instead of babel-node for CLI commands

tahini:
Typescript compile in commonjs, which is directly understandable by the `node` command and does not need babel-node.


kaligrafy:
It does not need a "type": "module" in the package.json file?

tahini:
apparently not...

tahini:
I think since it's commonjs, it does not need it. The type: module is for ES modules, which is not the case here.

tahini:
From https://nodejs.org/api/packages.html

"If the nearest parent package.json lacks a "type" field, or contains "type": "commonjs", .js files are treated as [CommonJS](https://nodejs.org/api/modules.html).
"

kaligrafy:
ok, fine!

kaligrafy (approved):

greenscientist:
I imagine this reduce the list of dependencies?

tahini:
> I imagine this reduce the list of dependencies?

Indeed I think we don't need any babel packages anymore! I'll keep the ones in webpack, as the babel-loader may be required for the client packages. But all other packages don't need it.

greenscientist (approved):
cool

PR #475, opened 2022-11-29
chaire-lib: Bump passport from 0.4.1 to 0.6.0

tahini:
And fix the call to logout, which now requires a callback.


greenscientist (approved):
Looks sane

PR #485, opened 2022-12-02
batchRouting: Reduce the amount of logging in batch calculations

tahini:
Each odTrip calculated was logged when read and when calculated. For large input files, that is a lot of log lines, that are useless for debugging and any actual error logged during the execution is kicked out of the log buffer.

Log entries at read time are shown only if the debug option is set to true. Execution logs are shown for every 1% progress.


greenscientist (approved):
ok. 

One day, we will have a proper logging infra :)

PR #490, opened 2022-12-07
chaire-lib: Use a proper typeguard for the `isSaveable` method

tahini:
The typeguard avoids having to cast the object before using it as Saveable.


tahini:
@greenscientist  Here's one that won't hurt your eyes anymore.

greenscientist (approved):
ok. I had to check a bit higher in the code to see where isSaveable was used.

PR #503, opened 2022-12-14
Improve the docker startup workflow and documentation

tahini:
* Copy the .env.example file to .env before the rest, to not override the .env file

* Add edit suggestions to the `docker-compose.yml` file for common use cases

* Update the docker section of the README with additional documentation, as it does not work out of the box and needs some tweaking.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,9 +26,11 @@ COPY packages/transition-backend packages/transition-backend
 COPY packages/transition-frontend packages/transition-frontend
 RUN yarn install
 
-
 # Copy the rest. (node_modules are excluded in .dockerignore)
 COPY . /app
+
+# Setup the example as a default configuration for the image
+COPY .env.docker /app/.env
 
 #TODO evaluate if any of those commands are necessary
 #RUN yarn setup
@@ -39,9 +41,6 @@ RUN yarn compile
 
 #TODO We probably need to do something different for the projects configuration directories
 # the docker-compose file have an example of using volume for part of a project
-
-# Setup the example as a default configuration for the image
-COPY .env.example /app/.env
 
 # Copy in json2capnp
 COPY --from=json2capnpbuild /app/services/json2capnp/target/debug/json2capnp services/json2capnp/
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -26,6 +26,8 @@ COPY packages/transition-backend packages/transition-backend
 COPY packages/transition-frontend packages/transition-frontend
 RUN yarn install
 
+# Setup the example as a default configuration for the image
+COPY .env.example /app/.env
```
tahini, line 30: @greenscientist  Can this be a security issue? The local .env may contain sensible information and now has precedence over the .env.example.

But currently, the .env.example does not work, as the PROJECT_CONFIG points to nowhere. Maybe the .env.example could just have the internal docker path? `/app/examples/config.js` instead of `/absolute/path/to/config.js`?

But then how does one run docker with a custom .env file?

kaligrafy (approved):

Dockerfile
greenscientist: You have to add the PROJECT_CONFIG to the environnement of your docker execution and not rely on what's in the docker file.

README.md
```diff
@@ -102,12 +102,18 @@ You can easily launch the whole transition system using Docker and thus not havi
 ### Building the image
 `docker build -t testtransition .`
 (You can replace testtransition with your prefered image name. Don't forget to update any other command and compose file if you do so)
-To run the application directly, you'll need to add an .env as previously described before building. 
+To run the application directly, you'll need to add an .env as previously described before building.
```
greenscientist, line 105: Maybe we should say that these are some default, but we need to have a few mandatory one. 

We could also have a docker specific .env file.

greenscientist (changes requested):
There's some rework to do about how we manage the .env for the docker environment.

Dockerfile
```diff
@@ -26,10 +26,12 @@ COPY packages/transition-backend packages/transition-backend
 COPY packages/transition-frontend packages/transition-frontend
 RUN yarn install
 
-
 # Copy the rest. (node_modules are excluded in .dockerignore)
 COPY . /app
 
+# Setup the example as a default configuration for the image
+COPY .env.docker /app/.env
```
greenscientist, line 33: Why not use the .env.example file and just sed the lines that needs to be different for docker?
Or
we could use ENV entries in the Dockerfile to override those entries with other values.

greenscientist (commented):
I'm not super happy with this, since managing environnement variable for docker outside of the standard docker mecanism is a bit clumsy. 
The current proposal work, but not hyper clean.

greenscientist (approved):
I'll approve, as it's ok. It's just harder on future maintenance. (2 different example files)
We can make changes later

tahini:
Thanks. Hopefully in the future, the only environment variable potentially required will be the path to the configuration file. But there's so many priorities and only so little time...

greenscientist:
exactly

PR #507, opened 2022-12-21
trRouting: Prepare routes to support v2/route endpoint

tahini:


greenscientist:
Do we really want to support the old code path??

tahini:
For now yes, so we can slowly do small reviewable PRs instead of a huge one (I had started the big one and it's all or nothing, here we can do the route first and accessibility next).

greenscientist:
If I understand this properly, we are adding a new Transition API which have a TrRouting version. Can't we just hide that behind the Transition API?

tahini:
Oh yeah and the client was preparing the query string as a string instead of the server... which makes it much much harder to support v2 here.

greenscientist (approved):
ok, let's do this then

PR #509, opened 2023-01-09
TrRouting: Use the new `v2/route` endpoint of trRouting

tahini:


kaligrafy (approved):

tahini:
> It's ok, but we seem to do a lot of work to keep the V1 stuff around. Yes it makes for simpler commit, but there will be a lot of scafolding to remove afterward.

We have to keep it, the v1 is used for everything, including the simulation odTrip,s calculation, so it will be around for a while still... And i don't think there's that much scaffolding to remove later.

You're probably right about keeping v2 in backend, but we'll have to handle other types of responses from the backend as well I guess and provide a single API for errors. Not trivial, but not hard either, I'll open an issue for it, before doing the accessibility endpoint.

greenscientist:
I really hope we don't have to keep V1 for a while. I really hope to kill it in TrRouting as soon as possible!

tahini:
Depends on you or me implementing the v2/odTrips in trRouting... ;-)

PR #516, opened 2023-01-12
Send a email to new users telling them that the account confirmation …

greenscientist:
…is pending.

This is to provide some feedback that the account creation worked.

Updated test and tested locally

### DIFF
--- a/locales/en/server.json
+++ b/locales/en/server.json
@@ -3,6 +3,8 @@
     "confirmByAdminSubject": "Transition: New User Confirmation",
     "confirmByUserText": "Hi {{name}},\n\nYou have requested access to transition.\n\nTo confirm your access, please click on the link below, or copy-paste the following URL {{userConfirmationUrl}}.\n\nThank you",
     "confirmByUserSubject": "Transition: New User Confirmation",
+    "pendingConfirmByAdminSubject": "Transition: account creation confirmation pending",
+    "pendingConfirmByAdminText": "Hi {{name}},\n\nyour access request to transition will be confirmed by an administrator shortly.\n\nThank you",
     "resetPasswordEmailSubject": "Forgotten password",
     "resetPasswordEmailText": "Hi {{name}},\n\nYou are receiving this because you have requested the reset of the password for your account.\n\nPlease click on the following link, or paste this into your browser to complete the process\n\n{{resetPasswordUrl}}\n\nIf you did not request this, please ignore this email and your password will remain unchanged.\n\nThank you",
     "magicLinkEmailText": "Welcome!\n\nTo connect to the application, click on the following link: {{magicLinkUrl}}.\n\nThank you",
--- a/locales/fr/server.json
+++ b/locales/fr/server.json
@@ -3,6 +3,8 @@
     "confirmByAdminSubject": "Transition: Confirmation d'un nouvel utilisateur",
     "confirmByUserText": "Bonjour {{name}},\n\nVous avez demandé l'accès à transition.\n\nPour confirmer votre accès, veuillez cliquer sur le lien ci-dessous, ou copier-coller l'URL dans votre navigateur {{userConfirmationUrl}}.\n\nMerci",
     "confirmByUserSubject": "Transition: Confirmation d'un nouvel utilisateur",
+    "pendingConfirmByAdminSubject": "Transition: compte en attente de confirmation par un administrateur",
+    "pendingConfirmByAdminText": "Bonjour {{name}},\n\nVotre accès à transition sera confirmé par un administrateur prochainement.\n\nMerci",
     "resetPasswordEmailSubject": "Mot de passe oublié",
     "resetPasswordEmailText": "Bonjour {{name}},\n\nVous recevez ce courriel car vous avez demandé de recréer un mot de passe pour votre compte.\n\nPour compléter la demande, veuillez cliquer sur le lien suivant:\n\n{{resetPasswordUrl}}\n\nSi vous n'avez pas effectué cette requête de changement de mot de passe, ignorez ce message et votre mot de passe demeurera inchangé.\n\nMerci",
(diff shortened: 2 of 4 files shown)


greenscientist:
Adding a related commit to add an email notification after the admin validate the account.

locales/fr/server.json
```diff
@@ -3,6 +3,8 @@
     "confirmByAdminSubject": "Transition: Confirmation d'un nouvel utilisateur",
     "confirmByUserText": "Bonjour {{name}},\n\nVous avez demandé l'accès à transition.\n\nPour confirmer votre accès, veuillez cliquer sur le lien ci-dessous, ou copier-coller l'URL dans votre navigateur {{userConfirmationUrl}}.\n\nMerci",
     "confirmByUserSubject": "Transition: Confirmation d'un nouvel utilisateur",
+    "pendingConfirmByAdminSubject": "Transition: compte en attente de confirmation par un administrateur",
+    "pendingConfirmByAdminText": "Bonjour {{name}},\n\nVotre accès à transition sera confirmé par un administrateur prochainement.\n\nMerci",
```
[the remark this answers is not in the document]
greenscientist: "Votre accès à transition doit d'abord être validé par un administrateur"  
Ok, ca me semble bon. 

Le coté anglais est correct?

locales/en/server.json
```diff
@@ -3,6 +3,10 @@
     "confirmByAdminSubject": "Transition: New User Confirmation",
     "confirmByUserText": "Hi {{name}},\n\nYou have requested access to transition.\n\nTo confirm your access, please click on the link below, or copy-paste the following URL {{userConfirmationUrl}}.\n\nThank you",
     "confirmByUserSubject": "Transition: New User Confirmation",
+    "pendingConfirmByAdminSubject": "Transition: account creation confirmation pending",
+    "pendingConfirmByAdminText": "Hi {{name}},\n\nyour access request to transition will be confirmed by an administrator shortly.\n\nThank you",
+    "confirmedByAdminEmailSubject": "Transition: account confirmed",
+    "confirmedByAdminEmailText": "Hi {{name}},\n\nyour access request have been confirmed. You can now use your account.",
```
tahini, line 9: has* been confirmed

locales/en/server.json
```diff
@@ -3,6 +3,10 @@
     "confirmByAdminSubject": "Transition: New User Confirmation",
     "confirmByUserText": "Hi {{name}},\n\nYou have requested access to transition.\n\nTo confirm your access, please click on the link below, or copy-paste the following URL {{userConfirmationUrl}}.\n\nThank you",
     "confirmByUserSubject": "Transition: New User Confirmation",
+    "pendingConfirmByAdminSubject": "Transition: account creation confirmation pending",
+    "pendingConfirmByAdminText": "Hi {{name}},\n\nyour access request to transition will be confirmed by an administrator shortly.\n\nThank you",
```
tahini, line 7: "will be reviewed by an administrator" instead of confirmed?

packages/chaire-lib-backend/src/services/auth/user.ts
```diff
@@ -58,13 +59,16 @@ export default class UserModel extends bookshelf.Model<UserModel> {
         return bcrypt.hashSync(password, bcrypt.genSaltSync(10));
     };
 
-    static confirmAccount = async (token: string): Promise<'Confirmed' | 'NotFound'> => {
+    static confirmAccount = async (token: string, confirmedByAdmin?: boolean): Promise<'Confirmed' | 'NotFound'> => {
```
tahini, line 62: To avoid a circular dependency, here, you could pass an optional confirmation callback that receives a user in parameter and that is called after the save. It also gives more flexibility to what to do after user confirmation.

packages/chaire-lib-backend/src/services/auth/user.ts
```diff
@@ -10,6 +10,7 @@ import moment from 'moment';
 import { BaseUser } from 'chaire-lib-common/lib/services/user/userType';
 import { serializePermissions } from './authorization';
 import { getHomePage } from './userPermissions';
+import { sendConfirmedByAdminEmail } from './userEmailNotifications';
```
tahini, line 13: This unfortunately creates a circular dependency between email notifications and user...

packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
```diff
@@ -95,6 +103,16 @@ export const sendConfirmationEmail = async (
...
 };
 
+export const sendConfirmedByAdminEmail = async (user: UserModel): Promise<void> => {
+    const email = {
+        mailText: 'server:confirmedByAdminEmailText',
+        mailSubject: 'server:confirmedByAdminEmailSubject',
+        toUser: user
+    };
+    await sendEmail(email, {});
+    console.log('Email sent confirmation by admin');
```
tahini, line 113: Email sent for* confirmation. Otherwise it's the email who sent a confirmation by admin, doesn't make sense :p

packages/chaire-lib-backend/src/services/auth/user.ts
greenscientist: But it seems to work...

tahini: Indeed it does! The language probably allows this, especially that they are in the same package. But it's usually something to avoid.

packages/chaire-lib-backend/src/services/auth/user.ts
greenscientist: Ideally, we would have a set of pre registered callbacks instead of passing it to the function, so that no matter where it's called from, the callback is active. But I will not do that now.

greenscientist:
Now with callback

packages/chaire-lib-backend/src/services/auth/user.ts
tahini: I think this is now an unused import

tahini (changes requested):
Just remove the unused import please, and this is good to merge

tahini (approved):

PR #527, opened 2023-01-19
gtfs: Fix format of color in GTFS import/export

### LINKED ISSUE #505: GTFS export: colours should not include #
When exporting to GTFS, the values in the `route_color` and `route_text_color` should not contain the `#` symbol.

tahini:
Fixes #505

In Transition, color fields are prefixed with '#', but in the gtfs specification, colors are the 6 hexadecimal digits of the colors. A helper method takes care of converting the color in the right format during import and exports removes the '#' prefix.


kaligrafy:
Would it be better to accept both (with or wihtout the sharp character)? I'm pretty sure I've seen gtfs with the sharp in the color fields. Google may already accept both with or without the #.

tahini:
We do accept both, as the unit tests show. But we produce without, as per the specification

kaligrafy (approved):

Issue #412, opened 2022-09-28
Add other direct "write" functions to OsmOverpassDownloader


greenscientist:
We added a WriteXml function so that we don't have to first convert the download to a string and then write to a file. 
the other Download* functions of the class could be converted and their usage also migrated. 
With a direct file download, we might need to migrate the class to backend instead of common

PR #537, opened 2023-01-20
Add WriteJson and writeGeojson fuction to OsmOverpassDownloader

greenscientist:
The writeXml was added in a previous commit. WriteJson re-use the same logic, but write geojson is more complexe as it needs to parse and convert the input file.

Fixes: #412


yarn.lock
```diff
@@ -11438,6 +11469,23 @@ osmtogeojson@^3.0.0-beta.4:
   optionalDependencies:
     "@types/geojson" "^1.0.2"
 
+osmtogeojson@^3.0.0-beta.5:
```
tahini, line 11472: beta.4 is still fetched because it is in the transition-common's package.json. It does not appear used there. could you remove it from there?

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
```diff
@@ -119,6 +141,38 @@ class OsmOverpassDownloaderImpl implements OsmOverpassDownloader {
...
+        filename: string,
+        boundPoly: GeoJSON.Polygon | GeoJSON.FeatureCollection | GeoJSON.Feature,
+        overpassXmlQueryString: string = allWaysAndRelationsXmlQuery
+    ): Promise<boolean> {
+        const response = await this.downloadData(boundPoly, overpassXmlQueryString, 'json');
+        const streamPipeline = promisify(pipeline);
+
+        console.log('Writing osm geojson data to ' + filename);
+
+        // Inspired by a conversion with ChatGPT :)
```
tahini, line 161: lol! also, conversation*

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
```diff
@@ -119,6 +141,38 @@ class OsmOverpassDownloaderImpl implements OsmOverpassDownloader {
...
     }
 
+    /**
+     * Download data from openstreetmap API, convert it to geojson and write it to
+     * a geojson file
+     * If the file exist, it will be overwritten
+     *
+     * See downloadGeojson for params
+     */
+    public async writeGeojson(
```
tahini, line 151: For more clarity, maybe call the methods that download and write date downloadAndWrite<Format>? or fetchAndWrite

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
```diff
@@ -119,6 +141,38 @@ class OsmOverpassDownloaderImpl implements OsmOverpassDownloader {
...
+    ): Promise<boolean> {
+        const response = await this.downloadData(boundPoly, overpassXmlQueryString, 'json');
+        const streamPipeline = promisify(pipeline);
+
+        console.log('Writing osm geojson data to ' + filename);
+
+        // Inspired by a conversion with ChatGPT :)
+        const writeStream = fs.createWriteStream(filename);
+        const jsonStream = JSONStream.parse('elements.*');
+        jsonStream.on('data', (element) => {
```
tahini, line 164: How and why is this write method different from the other write methods, who just call this.write. Maybe comment why you need a separate implementation here.

tahini (changes requested):

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
greenscientist: If I copy the desciption from the commit message, is it enough?

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
greenscientist: That would imply also renaming all the one that existed before right?

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
tahini: yes, so that's what osmToGeojson was for...

packages/chaire-lib-common/src/utils/osm/OsmOverpassDownloader.ts
tahini: In a world where method names should be self-explaining, then I guess so.

greenscientist:
@tahini Still need to rename the functions, I'll add an extra commit for that, but to avoid working on a stack of commit, wanted to have this part checked first.

tahini (approved):
You can do the rename of functions in a separate PR

PR #539, opened 2023-01-20
Fix projection not set issue in batch routing

tahini:


packages/transition-backend/src/services/accessMapLocation/AccessMapLocationProvider.ts
```diff
@@ -29,19 +29,22 @@ export interface accessMapLocationOptions {
     debug?: boolean;
 }
 
+const DEFAULT_PROJECTION_SRID = 4326;
```
tahini, line 32: Now that I see this, it should be a preference itself...

greenscientist, line 32: This random number needs a comment!

packages/chaire-lib-common/src/config/defaultPreferences.config.ts
```diff
@@ -544,7 +544,7 @@ const defaultPreferences: PreferencesModel = {
         routing: {
             batch: {
                 withGeometry: false,
-                projection: '+title=WGS 84 (long/lat) +proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees',
+                projection: 4326,
```
greenscientist, line 547: Again, needs a comment.

greenscientist (changes requested):
Request changes for the comment. (NEeds to know where this number comes from. 

It feels the code is somewhat duplicated between the 2 sections. Would be useful to have some common functions to deals with projections, default and conversion. (But not a requirement for this PR)

packages/transition-backend/src/services/accessMapLocation/AccessMapLocationProvider.ts
greenscientist: True, since a user will probably always use the same projection.

tahini: This is actually a server configuration, as the FIXME I'll add explains

tahini: > It feels the code is somewhat duplicated between the 2 sections. Would be useful to have some common functions to deals with projections, default and conversion. (But not a requirement for this PR)

later...

packages/chaire-lib-common/src/config/defaultPreferences.config.ts
```diff
@@ -542,6 +542,9 @@ const defaultPreferences: PreferencesModel = {
             defaultColor: '#0086FF'
         },
         routing: {
+            // FIXME This value is a configuration, not a preference, it should match the projection expected by the calculator (here trRouting, whence 4326)
+            // Refers to EPSG:4326 projection, WGS84 World Geodetic System, used in GPS https://epsg.io/4326
+            defaultProjection: 4326,
```
greenscientist, line 547: I'll be picky here, but should the name be more something like "serverProjection", "calculatorProjection" or something along those lines?

tahini: It's not even a preference, it should be a constant. All Transition use geojson, which uses the WSG84 coordinates system. Only data coming from the outside can be in a different projection system, but we need to convert at import. Now, where does this constant go...

kaligrafy (approved):

packages/chaire-lib-common/src/config/defaultPreferences.config.ts
```diff
@@ -604,11 +605,7 @@ const defaultPreferences: PreferencesModel = {
         }
     },
     proj4Projections: {
-        '4326': {
-            srid: 4326,
-            label: 'WGS84 EPSG:4326 (latitude/longitude)',
-            value: '+title=WGS 84 (long/lat) +proj=longlat +ellps=WGS84 +datum=WGS84 +units=degrees'
-        },
+        [String(constants.geographicCoordinateSystem.srid)]: constants.geographicCoordinateSystem,
         '2950': {
```
greenscientist, line 609: Maybe not for now, but probably all projections definition should in in constants somewhere and not in a defaultPreferences file.

packages/transition-backend/src/services/odTrip/odTripProvider.ts
```diff
@@ -36,27 +37,27 @@ const extractOdTrip = (
...
     },
     options: OdTripOptions,
-    projections: any
+    projection: { srid: number; value: string }
 ): BaseOdTrip => {
     const internalId = line[options.idAttribute];
     const originGeography =
-        options.projection === '4326'
+        projection.srid === constants.geographicCoordinateSystem.srid
             ? turfPoint([parseFloat(line[options.originXAttribute]), parseFloat(line[options.originYAttribute])])
```
greenscientist, line 45: Not a bit deal either, but this check + conversion could be in a fonction and reused for origin and destination.

greenscientist (approved):
probably ok, but still have comments :)

packages/transition-backend/src/services/odTrip/odTripProvider.ts
tahini: It's issue #540

packages/chaire-lib-common/src/config/defaultPreferences.config.ts
tahini: I'll add an issue, there's no reason to support only MTM zone 8 and WSG84. There's a lot more geographic coordinates systems. Preferences could reduce the supported subset of an instance maybe, but Transition should support them all.

Issue #549, opened 2023-01-25
auth: Email should be case insensitive


tahini:
It is possible to add 2 users with emails test@test.org and TEST@TEST.ORg. Emails should be saved as lowercase and always lower cased

PR #555, opened 2023-02-07
Fix/email case

nik498:
**Draft**

This would fix issue #549.

This has been tested through the command line with yarn create-user and through the registration form. Both reject the creation of a new user when emails or usernames are identical, regardless of the case.

To test : What happens when you try to create a new user with an email that's already the username of another user, regardless of the case?

TODO : Add a migration to delete duplicates and sanitize all emails to lowercase.


tahini (changes requested):
The current solution requires to change many code paths: the task, the passport for creation, then you'll have to add the toLowerCase to queries so that uppercase email is ok. There's no guarantee you cover it all, or that eventually someone won't add a faulty code path.

Ideally, since all queries to database pass by the users.db.queries.ts file, this is the only file that should be modified to fix this. You can add a function that cleans the attributes (we have that for example in other db files, like transitNodes.db.queries.ts, attributesCleaner) and by which you pass all the parameters you receive in all functions.

And of course, add db tests to make sure it all works! Database tests can be run with `yarn test:sequential`, but it's tricky to set up... let us know if it doesn't work for you.

PR #563, opened 2023-02-10
Scenario Analysis: Move the batch routing feature to a Scenario analysis section

tahini:
* Add a new section, "Scenario analysis", which is part of the "supply/demand analysis" module.

* The current batch routing functionality is moved to this new section as follows:
  - The main panel displays by default the list of batchRouting tasks that belongs to the user. It also contains a button to do a new analysis
  - The new analysis button opens a 3-steps wizard, where the user can 
    1- select the demand CSV file and map the fields, 
    2- Specify the calculation parameters (travel times, scenario, withAlternatives) 
    3- Visualize the various parameters and either go back to modify or click Calculate. Clicking calculate triggers the calculation and the user is back to the job list and can see the progress.

* The current batch calculation is replaced by a message telling the user to go to the "Scenario analysis" section instead.

**Coming improvements:**

- Add texts to the forms
- Move the "detailed results" and "with geometries", which used to be part of the batch routing form, to a calculation options object, and maybe put them in the form where the user can visualize the parameters (step 3) and decide on the calculation option.
- Review how/where the buttons are handled. Now, each individual form makes its own validation of the data, but has to tell the parent form to enabled the next/previous button. It looks awkward.
- Replace the job list component by something prettier, like in the other list, with buttons, where the user can select one and expand it in a div below

### DIFF
--- a/locales/en/main.json
+++ b/locales/en/main.json
@@ -25,6 +25,8 @@
   "Delete": "Delete",
   "loading": "loading",
   "Loading": "Loading",
+  "Next": "Next",
+  "Previous": "Previous",
   "DiscardChanges": "Discard Changes",
   "ConfirmBackModal": "There are unsaved changes. Discard current changes and leave this section?",
   "UnsavedChanges": "There are unsaved changes on the current objects. Please save them before leaving this section.",
--- a/locales/en/main.json
+++ b/locales/en/main.json
@@ -31,7 +31,7 @@
   "connectionIsOffline": "You are offline, please verify your connection.",
   "Routing": "Routing",
   "AccessibilityMap": "Accessibility map",
-  "OdRouting": "OD batch routing",
+  "BatchCalculation": "Batch calculations",
   "RestartTrRouting": "[Re]start routing engine",
   "SaveAllData": "Save all data for the routing engine",
   "PrepareData": "Prepare data",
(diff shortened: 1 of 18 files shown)


tahini:
Here are some screenshots for this feature

List calculation tasks

scenarioAnalysisList

Select the demand, from a local CSV file

newScenarioAnalysisStep1

Configure the scenario parameters:

newScenarioAnalysisStep2

Confirm the parameters and start calculations

newScenarioAnalysisStep3

tahini:
To put it in context, this is the first step to get the complete scenario analysis and comparison functionality:

- The demand specification step can later be bonified with a choice of CSV file or a pre-uploaded/pre-calculated demand in database, and/or add field mappings and filters
- The parameters step can later be bonified with comparison options and/or add options
- The task result view that doesn't exist yet can eventually be used for various visualizations and/or let the user create a new simulation from a current one, where he can change the parameters or scenario.

kaligrafy (approved):

greenscientist:
I'm just curious if moving it is the right thing to do? (or if the name of the new section right for it.. 

How would a user who want to do a routing in batch think of going to that new section?

tahini:
I added a button where there used to be the batch calculation to directly navigate to the new "batch calculation" form. Also, everything is now called "batch calculation", or sometimes just "calculations" instead of "scenario analysis".

tahini:
@greenscientist is there anything else to add to this PR? Otherwise, let's get it in the hands of the users and see if they notice and complain!

greenscientist (approved):
It's probably good. Let's how does it feel in use and we can adjust later.

PR #567, opened 2023-02-20
Add documentation for getting started with Docker Desktop

tahini:
This documents a default user desktop installation using docker for desktop. It is aimed at end users who do not want to clone this repository and explains the steps to have a running local Transition instance in Docker.


kaligrafy (approved):
maybe we should also add a documentation section to show how to install nginx to serve the transition docker instance to the outside world.

tahini:
> maybe we should also add a documentation section to show how to install nginx to serve the transition docker instance to the outside world.

I'd say no. Opening to the outside world can be a security issue, I don't want to encourage this for people who "just want to install on their laptops". To serve the outside world, one needs to know what they are doing and thus don't need our documentation.

greenscientist:
> > maybe we should also add a documentation section to show how to install nginx to serve the transition docker instance to the outside world.
> 
> I'd say no. Opening to the outside world can be a security issue, I don't want to encourage this for people who "just want to install on their laptops". To serve the outside world, one needs to know what they are doing and thus don't need our documentation.

Also, they don't need nginx to serve to the outside world! That part is really a general hosting questions and we should not bother with it.

README.md
```diff
@@ -10,6 +10,10 @@ See http://transition.city/
 
 For Ubuntu users: [complete step-by-step development environment setup procedure](docs/setupDevEnvironmentUbuntu20.04.md)
 
+## User installation
```
greenscientist, line 13: I would change the title to something like
"Stand alone Desktop installation"

user installation is not really clear at what it is. Could be interpreted as "add user to the platform)

README.md
```diff
@@ -10,6 +10,10 @@ See http://transition.city/
...
+
+For users only interested in running Transition, without manually installing all the dependencies, follow these [step-by-step instructiont to set up Transition using Docker](docs/runWithDocker.md).
```
greenscientist, line 15: using Docker Desktop

docs/runWithDocker.md
```diff
@@ -0,0 +1,203 @@
+# Install complete Transition with Docker
+
+This installation documentation is targeted in end-users who want to try Transition locally, who do not have access to an infrastructure for web server deployment. It runs the server and its dependencies using `docker`.
+
+### 1. Install Docker for desktop
+
+Follow [the docker for desktop](https://www.docker.com/products/docker-desktop/) for instructions to install the docker for desktop application locally.
+
+If you are already familiar with docker and can run container and scripts  easily, you don't have to install docker for desktop.
```
greenscientist, line 9: extra space between scripts and easily

docs/runWithDocker.md
```diff
@@ -0,0 +1,203 @@
...
+        // confirmEmail: true
+      }
+    },
+  
+    mapDefaultCenter: {
+      lat: 45.5092960,
+      lon: -73.4769080
+    },
+  
+    languages: ['fr', 'en'],
```
greenscientist, line 82: Thinking about it, we really don't need a bilingual example for a desktop setup

greenscientist (changes requested):
So I'm wondering, do we want to have instructions where people install postgis directly on their machine instead of having to deal with a docker-compose for it? (Would it be simpler?)

docs/runWithDocker.md
tahini: At the Chaire Mobilité, given our partners and funders, though we don't need a bilingual setup, it has to be available in french. so we might as well keep it bilingual.

tahini: Though most of the bilingualism support shouldn't be in the config.js file... But when we revisit the preferences and configuration, we can change that.

greenscientist: We can provide a translation of the page in french also instead.

tahini: En fait, il faudrati juste que l'application soit bilingue et ne pas nécessiter de config pentoute...

greenscientist: Il va probablement quand même falloir traduire la page d'installation éventuellement.

greenscientist (approved):

PR #568, opened 2023-02-21
jobs: Allow to restart running and pending jobs when server restarts

### LINKED ISSUE #476: transition: jobs should be restarted with the server
Now, running and pending jobs when the server stops are not restarted when the server restarts. This leaves them in a state which gives the impression that they will eventually finish, but no, they won't...

For now, the running jobs should just be restarted and the pending ones added to the queue

tahini:
Fixes #476

This enqueues any executable jobs that are in progress or pending when
the server starts so they can be run again. In progress jobs are simply
restarted, running jobs are enqueued first, then the pending ones
ordered by the creation time, earliest first.

Add unit tests for resume functions and queries with multiple sort options

### DIFF
--- a/packages/transition-backend/src/models/db/jobs.db.queries.ts
+++ b/packages/transition-backend/src/models/db/jobs.db.queries.ts
@@ -9,7 +9,7 @@ import knex from 'chaire-lib-backend/lib/config/shared/db.config';
 import { truncate, destroy } from 'chaire-lib-backend/lib/models/db/default.db.queries';
 import TrError from 'chaire-lib-common/lib/utils/TrError';
 import { Knex } from 'knex';
-import { JobAttributes, JobDataType } from 'transition-common/lib/services/jobs/Job';
+import { JobAttributes, JobDataType, JobStatus } from 'transition-common/lib/services/jobs/Job';
 
 const tableName = 'tr_jobs';
 
@@ -36,6 +36,7 @@ const collection = async (
     options: {
         userId?: number;
         jobType?: string;
+        statuses?: JobStatus[];
         pageIndex: number;
         pageSize: number;
         sort?: { field: string; direction: 'asc' | 'desc' };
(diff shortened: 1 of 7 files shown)


kaligrafy (approved):

packages/transition-backend/src/models/db/jobs.db.queries.ts
```diff
@@ -49,6 +50,19 @@ const collection = async (
             if (options.jobType !== undefined) {
                 query.where('name', options.jobType);
             }
+            if (options.statuses !== undefined) {
+                const statuses = options.statuses;
+                if (statuses.length === 1) {
+                    query.where('status', options.statuses[0]);
+                } else if (statuses.length > 1) {
+                    query.where(function () {
+                        this.where('status', statuses[0] as JobStatus);
```
greenscientist, line 59: I don't know much about knex, but instead of orWhere, is it possible to do a WHERE iN (LIST of status)

I would think that would be more optimal. I don't know if the optimisation is done automatically)

packages/transition-backend/src/socketServerApp.ts
```diff
@@ -42,6 +42,14 @@ const setupSocketServerApp = async function (server, session) {
     serviceLocator.addService('socketEventManager', new events.EventEmitter());
     allSocketRoutes(serviceLocator.socketEventManager);
 
+    // Enqueue/resume running and pending tasks
+    // FIXME This implies a single server process for a given database. We don't
+    // know if the job is enqueued in another process somewhere and may be
+    // executed twice. For pending jobs, we could have a first run, first serve,
+    // but for jobs in progress, we don't know if they are actively being run or
+    // not
+    await ExecutableJob.enqueueRunningAndPendingJobs(serviceLocator.socketEventManager);
```
greenscientist, line 51: AS per you comment mention, we could have case where you won't want this. 
It should be behind a parameter setting so we can disable it. Enabled by default. 
(I can foresee in particular people running on their desktop not wanting to have this restart if something went bad

greenscientist:
Everything I see something related to our config management I have a bit of a chill

tahini:
> Everything I see something related to our config management I have a bit of a chill

I guess you are wondering why I'm putting it in .env and not in config.js... So am I... anyhow...

greenscientist:
> > Everything I see something related to our config management I have a bit of a chill
> 
> I guess you are wondering why I'm putting it in .env and not in config.js... So am I... anyhow...

I did not want to ask!

tahini:
@greenscientist If that is not a major objection (it's using the same mechanism as for the cache creation at startup, which makes sense, and we'll handle all the config/env stuff eventually), I'll merge as this + the trRouting updates make the batch calculations more robust.

PR #578, opened 2023-02-25
add rail lua profile

kaligrafy:


packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/rail.lua
```diff
@@ -0,0 +1,138 @@
+-- Copyright 2017-2019 Railnova SA <support@railnova.eu>, Nikita Marchant <nikita.marchant@gmail.com>
+-- Code under the 2-clause BSD license
+
+api_version = 4
```
tahini, line 4: Where does this file come from? Is it copy-pasted from anywhere? If from osrm, then why do we need it in Transition and don't use the default from there? If it is modified from a copy-paster of osrm, could you add a comment at the top of the file explaining why you need to modify the default rail profile (so that we know right away when opening the file later) and also a comment in the file itself where the modification was done (to clearly highlight the change)

tahini (changes requested):
The commit message also needs to specify where the file comes from and why it is added here.

packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/rail.lua
kaligrafy: osrm does not include any rail profile, ans this one was wroking well. Including the copyright with license is not enough?

tahini: This copyright does not tell where it comes from. This code was not produced for Transition, these people published it somewhere. And when we need to update it 2 years from now, we want to be able to track it.

kaligrafy:
I will add the link:
https://github.com/railnova/osrm-train-profile

tahini:
> I will add the link: https://github.com/railnova/osrm-train-profile

Perfect, also add what you said in your previous comment in the commit message, that osrm does not include rail profile.

packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/rail.lua
kaligrafy: added source + commit msg

tahini (approved):

PR #596, opened 2023-03-14
Fix case insensitive emails

nik498:
This would fix issue #549 .

This has been tested through the command line with yarn create-user and through the registration form. Both reject the creation of a new user when emails or usernames are identical, regardless of the case.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
+++ b/packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
@@ -55,7 +55,6 @@ beforeAll(async () => {
     await truncate(knex, 'users');
     await knex('users').insert(user);
     await knex('users').insert(user2);
-    await knex('users').insert(userCaps);
 });
 
 afterAll(async() => {
@@ -166,7 +165,7 @@ test('Update converts new emails to lowercase', async () => {
 
 test('Collection', async () => {
     const collection = await dbQueries.collection();
-    expect(collection.length).toEqual(4);
+    expect(collection.length).toEqual(3);
     const dbUser = collection.find((u) => u.id === user2.id);
     expect(dbUser).toBeDefined();
     expect(dbUser).toEqual(expect.objectContaining(user2));
@@ -174,7 +173,7 @@ test('Collection', async () => {
 
 describe('list users', () => {
 
-    const nbUsers = 4;
+    const nbUsers = 3;
 
     test('Get the complete list', async () => {
         const { users, totalCount } = await dbQueries.getList({ filters: {}, pageIndex: 0, pageSize: -1 });
@@ -188,16 +187,16 @@ describe('list users', () => {
     });
 
     test('Get paginated users list', async () => {
-        // 2 pages, the last one has 2 elements
+        // 2 pages
         const pageSize = 2;
-        // First page
+        // First page, should have 2 elements
         const { users: page1, totalCount: totalCount1 } = await dbQueries.getList({ filters: {}, pageIndex: 0, pageSize });
         expect(totalCount1).toEqual(nbUsers);
         expect(page1.length).toEqual(pageSize);
-        // Second page, should have 2 elements
+        // Second page, should have 1 element
         const { users: page2, totalCount: totalCount2 } = await dbQueries.getList({ filters: {}, pageIndex: 1, pageSize });
         expect(totalCount2).toEqual(nbUsers);
-        expect(page2.length).toEqual(2);
+        expect(page2.length).toEqual(1);
         const inOtherPage = page2.find((user) => page1.find((user2) => user2.id === user.id) !== undefined);
         expect(inOtherPage).toBeUndefined();
         // There is no third page
@@ -213,13 +212,13 @@ describe('list users', () => {
     });
 
     test('Get users list with email filter', async () => {
-        // Use 'test' string, should return 2 users as one email is modified prior
+        // Use 'test' string, should return 1 user as one email is modified before
         const { users: testUsers, totalCount: totalCountTest } = await dbQueries.getList({ filters: { email: 'test' }, pageIndex: 0, pageSize: -1 });
-        expect(totalCountTest).toEqual(2);
-        expect(testUsers.length).toEqual(2);
+        expect(totalCountTest).toEqual(1);
+        expect(testUsers.length).toEqual(1);
 
-        // Use 'test.org' string, should return 1 user
-        const { users: testOrgUsers, totalCount: totalCountTestOrg } = await dbQueries.getList({ filters: { email: 'test.org' }, pageIndex: 0, pageSize: -1 });
+        // Use 'example.com' string, should return 1 user
+        const { users: testOrgUsers, totalCount: totalCountTestOrg } = await dbQueries.getList({ filters: { email: 'example.org' }, pageIndex: 0, pageSize: -1 });
         expect(totalCountTestOrg).toEqual(1);
         expect(testOrgUsers.length).toEqual(1);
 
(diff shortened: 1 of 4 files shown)


nik498:
Could 
```
is_valid: params.isTest === true ? false : true, 
is_test: params.isTest,
```
be changed to 
```
is_valid: !params.isTest,
is_test: params.isTest,
```
?

tahini:
> In passport,utils.ts, could
> 
> ```
> is_valid: params.isTest === true ? false : true, 
> is_test: params.isTest,
> ```
> 
> be changed to
> 
> ```
> is_valid: !params.isTest,
> is_test: params.isTest,
> ```
> 
> ?

Not really, isTest could accidentally have a value that is not `true` but that would be considered `true` by params.isTest, like any string or non-zero number. With this argument, it's rather the is_test: that should be params.isTest === true ? true : false

tahini (changes requested):
This PR seems to include commits that were already merged in the main branch. You should rebase your branch on upstream's main branch, instead of merging it for the PR. Just checkout your branch, and do a `git rebase main` and solve any conflict along the way. If the conflicting commit is already merged, just skip it with `git rebase --skip` and if everything seems wrong and you're not sure, you can abort with `git rebase --abort`. Let us know if you need help with this.

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -32,6 +32,15 @@ const user2 = {
     password_reset_token: 'sometokenforpasswordreset'
 };
 
+const userCaps = {
+    id: 3,
+    uuid: uuidV4(),
+    username: 'testCaps',
+    email: 'TEST@TEST.ORG',
```
tahini, line 39: To make sure this is always valid, you can use user.email.toUpperCase(), in case one gets edited, and the other not. Even better, you could extract test@test.org to a variable and user toLowerCase() for user and toUpperCase for userCaps

tahini (changes requested):
Thanks for the PR! This PR makes sure that all new users will be inserted in lowercase, but there are some more tests to do:

* In the `Find user by %s` test, you should add test cases with various cases for valid emails.
* In the list test, the `Get users list with email filter` test should also include various cases

Lastly, what about current emails? If we assume everything is lowercase now, current emails in the DB should be updated to lowercase too. That means adding a migration to put everything in lowercase. Another option would be to keep the current case, but use whereILike for the queries for case insensitive comparisons (see https://knexjs.org/guide/query-builder.html#whereilike). I just found this exists, I never used it before. If you decide to use this, make sure to add tests to validate that exact queries are exact (find) and searches accept partial matches (getList's filters).

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -51,14 +62,17 @@ afterAll(async() => {
...
-    [{ first_name: user.first_name }, user ],
+    [{ first_name: user.first_name }, user],
     [{}, undefined],
     [{ first_name: 'other'}, undefined],
     [{ first_name: user2.first_name, username: user2.username }, user2],
     [{ usernameOrEmail: user2.email }, user2],
     [{ usernameOrEmail: user2.username }, user2],
     [{ password_reset_token: user2.password_reset_token }, user2],
     [{ username: user2.username, email: 'arbitrary' }, undefined],
+    [{ usernameOrEmail: user2.email.toUpperCase() }, undefined],
```
tahini, line 73: I would expect user2 to be found here, no?

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -51,14 +62,17 @@ afterAll(async() => {
...
+    [{ usernameOrEmail: userCaps.email }, userCaps],
+    [{ username: user.username, email: userCaps.email}, undefined]
```
tahini, line 75: Here too, userCaps and user both have the same email, so it should return the user object

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -51,14 +62,17 @@ afterAll(async() => {
...
+    [{ first_name: user.first_name }, user],
     [{}, undefined],
     [{ first_name: 'other'}, undefined],
     [{ first_name: user2.first_name, username: user2.username }, user2],
     [{ usernameOrEmail: user2.email }, user2],
     [{ usernameOrEmail: user2.username }, user2],
     [{ password_reset_token: user2.password_reset_token }, user2],
     [{ username: user2.username, email: 'arbitrary' }, undefined],
+    [{ usernameOrEmail: user2.email.toUpperCase() }, undefined],
+    [{ usernameOrEmail: userCaps.email }, userCaps],
```
tahini, line 74: I think this should return `user`. `userCaps` is not in the database, no?

packages/chaire-lib-backend/src/models/db/users.db.queries.ts
```diff
@@ -68,7 +77,7 @@ const find = async (
         }
         const query = knex(tableName);
         if (whereData.usernameOrEmail !== undefined) {
-            query.where('email', whereData.usernameOrEmail);
+            query.whereILike('email', whereData.usernameOrEmail);
```
tahini, line 80: You probably have to add the whereILike in the below for the email and username cases. The forEach probably needs to be changed to something specific for those 2 fields, like it is done here.

tahini (changes requested):
I think some of the tests are not right. Or I may be mistaken.

tahini (changes requested):
Do the tests pass now? And also, user and userCaps shouldn't be both in the database, as it should not be possible to have the same email with multiple cases.

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -44,22 +55,27 @@ beforeAll(async () => {
     await truncate(knex, 'users');
     await knex('users').insert(user);
     await knex('users').insert(user2);
+    await knex('users').insert(userCaps);
```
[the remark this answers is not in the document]
nik498: userCaps has been removed, the test that tries to insert but fails already existed

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -51,15 +62,19 @@ afterAll(async() => {
...
-    [{ first_name: 'other'}, undefined],
+    [{ first_name: 'other' }, undefined],
     [{ first_name: user2.first_name, username: user2.username }, user2],
     [{ usernameOrEmail: user2.email }, user2],
     [{ usernameOrEmail: user2.username }, user2],
     [{ password_reset_token: user2.password_reset_token }, user2],
     [{ username: user2.username, email: 'arbitrary' }, undefined],
+    [{ usernameOrEmail: user2.email.toUpperCase() }, user2],
+    // When using userCaps as expected value, the email would get sanitized and wouldn't match the expected
+    [{ usernameOrEmail: userCaps.email }, { email: testEmail.toLowerCase() }],
```
tahini, line 75: Then the expected can be simply 'user'?

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -187,20 +212,25 @@ describe(`list users`, () => {
...
 
         // Use 'foo' string, should return nothing
         const { users: fooUsers, totalCount: totalCountFoo } = await dbQueries.getList({ filters: { email: 'foo' }, pageIndex: 0, pageSize: -1 });
         expect(totalCountFoo).toEqual(0);
         expect(fooUsers.length).toEqual(0);
+
+        // Use 'TEST' string, should return 1 user as it is case insensitive
+        const { users: capsUsers, totalCount: totalCountCaps } = await dbQueries.getList({ filters: { email: 'TEST' }, pageIndex: 0, pageSize: -1 });
+        expect(totalCountCaps).toEqual(1);
+        expect(capsUsers.length).toEqual(1);
```
tahini, line 233: Here, you could add `expect(capsUsers).toEqual(testUsers)` as this is expected to be the same user in both queries. And for test flow, the caps case should be just below the initial `test` test.

packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
```diff
@@ -221,10 +251,10 @@ describe(`list users`, () => {
     });
 
     test('Get users list with username and email filter', async () => {
-        // Use 'test' string for both email and username, should return 2 users
+        // Use 'test' string for both email and username, should return 1 user
         const { users: testUsers, totalCount: totalCountTest } = await dbQueries.getList({ filters: { username: 'test', email: 'test' }, pageIndex: 0, pageSize: -1 });
-        expect(totalCountTest).toEqual(2);
-        expect(testUsers.length).toEqual(2);
+        expect(totalCountTest).toEqual(1);
+        expect(testUsers.length).toEqual(1);
```
tahini, line 257: It's ok to have changed an email above, but it has the unfortunate effect that now all `getList` tests return only 1 value. We should have at least one tests which returns more than 1 value to make sure we don't accidentally limit our output. You could just change the email from 'test' to 'org', then both user and user2 would be returned.

tahini (changes requested):
Almost! just a few details and it's ok!

PR #605, opened 2023-03-16
dataSources: Add a data source owner to allow private data sources

tahini:
Add a `owner` column to the data source, which can be null (for global data sources), or a foreign key to the userId. Queries are updated so they take an optional userId parameter to retrieve only allowed data sources (since data sources are currently mostly used in tasks, like simulations, which do not have a userId, we can't make this parameter mandatory for now).

For now, the main intention is to allow users to import private odTrip data, without sharing it with all users (odTrip import feature is currently not working, but when it is brought back, the option to import a global data source (with null owner) will be available)


kaligrafy (approved):

greenscientist:
we should note that we will build a whole better owner ship model for all the data in the future. (To allow multiple project on once instance.)  
We should not do to much adhoc thing list this.

PR #607, opened 2023-03-17
jobs: Remove await from job when restarted at the beginning

tahini:
The await for the enqueue effectively waits for the whole job to be complete! This is not desired.

### DIFF
--- a/packages/transition-backend/src/services/executableJob/ExecutableJob.ts
+++ b/packages/transition-backend/src/services/executableJob/ExecutableJob.ts
@@ -40,7 +40,7 @@ export class ExecutableJob<TData extends JobDataType> extends Job<TData> {
             console.log(`Enqueuing ${runningAndPending.jobs.length} in progress and pending jobs`);
             for (let i = 0; i < runningAndPending.jobs.length; i++) {
                 const job = new ExecutableJob<TData>(runningAndPending.jobs[i] as JobAttributes<TData>);
-                await job.enqueue(progressEmitter);
+                job.enqueue(progressEmitter);
             }
 
             return true;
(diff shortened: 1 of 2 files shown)


packages/transition-backend/src/services/executableJob/__tests__/ExecutableJob.test.ts
```diff
@@ -164,6 +165,7 @@ describe('Test resume running and pending', () => {
         }];
         mockedJobCollection.mockResolvedValueOnce({ jobs: jobsToRun, totalCount: jobsToRun.length })
         expect(await ExecutableJob.enqueueRunningAndPendingJobs(progressEmitter)).toEqual(true);
+        await TestUtils.flushPromises();
```
greenscientist, line 168: Probably need a comment about why it's added

packages/transition-backend/src/services/executableJob/ExecutableJob.ts
```diff
@@ -40,7 +40,7 @@ export class ExecutableJob<TData extends JobDataType> extends Job<TData> {
             console.log(`Enqueuing ${runningAndPending.jobs.length} in progress and pending jobs`);
             for (let i = 0; i < runningAndPending.jobs.length; i++) {
                 const job = new ExecutableJob<TData>(runningAndPending.jobs[i] as JobAttributes<TData>);
-                await job.enqueue(progressEmitter);
+                job.enqueue(progressEmitter);
             }
 
             return true;
```
greenscientist, line 46: What does the return true here means?
DOes the mean clash with the fact that we don't await the enqueuing?

packages/transition-backend/src/services/executableJob/__tests__/ExecutableJob.test.ts
tahini: done

packages/transition-backend/src/services/executableJob/ExecutableJob.ts
tahini: That there was no exception and everything was correct, instead of just returning nothing... we don't do anything with it anyway

greenscientist (approved):
ok

kaligrafy (approved):

Issue #54, opened 2021-04-15
Remove language name and locale translation from the config file and convert the fr/en toggle button by a menu language


greenscientist:
Language, language name and locale are standards, they should not live in the configuration. 
The app itself have the list of languages it support. 
Having a configurable alias for standard locales is a bit unnecessary. 

So we should remove those entries from the config.js file: 
  languages: ['fr', 'en'],
  locales: {
    fr: 'fr-CA',
    en: 'en-CA'
  },
  languageNames: {
    fr: "Français",
    en: "English"
  },

kaligrafy:
We should also change the current English/Français toggle by a menu language which would list all available languages

PR #609, opened 2023-03-20
config: Fetch config from server in frontend

tahini:
Instead of having it built into the application at webpack time, it is fetched at load time from the server.

This adds an express route named `/config` that returns a Status with the project configuration.

The client's entry page now needs to asynchronously fetch it from server using this route and the page is only rendered when the response is back. If there is an error while getting the configuration, the MaintenancePage is displayed.

### DIFF
--- a/packages/chaire-lib-frontend/src/components/pages/index.ts
+++ b/packages/chaire-lib-frontend/src/components/pages/index.ts
@@ -7,3 +7,4 @@
 export { default as LoadingPage } from './LoadingPage';
 export * from './NotFoundPage';
 export { default as LoginPage } from './LoginPage';
+export { default as MaintenancePage } from './MaintenancePage';
--- a/packages/transition-frontend/src/app-transition.tsx
+++ b/packages/transition-frontend/src/app-transition.tsx
@@ -11,13 +11,13 @@ import { I18nextProvider } from 'react-i18next';
 import { createBrowserHistory } from 'history';
 import { Router } from 'react-router-dom';
 
-import i18n from 'chaire-lib-frontend/lib/config/i18n.config';
+import initI18n from 'chaire-lib-frontend/lib/config/i18n.config';
 import TransitionRouter from './components/routers/TransitionRouter';
 import MainMap from './components/map/TransitionMainMap';
 import configureStore from 'chaire-lib-frontend/lib/store/configureStore';
 import { login, logout } from 'chaire-lib-frontend/lib/actions/Auth';
-import { LoadingPage } from 'chaire-lib-frontend/lib/components/pages';
-import config from 'chaire-lib-frontend/lib/config/project.config';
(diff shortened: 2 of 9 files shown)


packages/chaire-lib-frontend/src/components/pages/index.ts
```diff
@@ -7,3 +7,4 @@
 export { default as LoadingPage } from './LoadingPage';
 export * from './NotFoundPage';
 export { default as LoginPage } from './LoginPage';
+export { default as MaintenancePage } from './MaintenancePage';
```
greenscientist, line 10: Is this change relevant?

packages/transition-frontend/src/app-transition.tsx
```diff
@@ -50,28 +50,41 @@ const contributions = [
...
+        ReactDOM.render(jsx(), document.getElementById('app'));
         hasRendered = true;
     }
 };
 
+fetchConfiguration().then((configOk: boolean) => {
+    hasConfig = configOk;
+    renderApp();
+});
+
```
greenscientist, line 84: Should we have a catch or a else to catch (and log) failure here?

greenscientist (dismissed):
I had comments that maybe need fixing, but approving since they might be ok.

packages/chaire-lib-frontend/src/components/pages/index.ts
tahini: Yes, it allows to import the MaintenancePage along with the rest instead of a separate import.

packages/transition-frontend/src/app-transition.tsx
```diff
@@ -50,28 +50,41 @@ const contributions = [
...
     </Provider>
 );
 
 ReactDOM.render(<LoadingPage />, document.getElementById('app'));
 
-let hasRendered = false;
 const renderApp = () => {
-    if (!hasRendered) {
-        ReactDOM.render(jsx, document.getElementById('app'));
+    if (!hasRendered && hasConfig !== undefined && hasFetchedAuth !== undefined) {
```
tahini, line 74: In retrospect, 2 things I don't like here: the multiple flags and the fact that we need to call a i18n initialization function. It makes the currentLanguage potentially undefined and multiplies the possibilities. I think we should fetch the configuration, then import the rest which we are certain will run with a valid configuration. If we can't get the config, we put a [eventually] pretty and generic failure page.

tahini (commented):
Thanks for approving the direction, but I think this can still be improved. Especially that the application "entry" will have to be updated in Evolution also. I'll dismiss your review for now so it doesn't accidentally gets merged and I'll try to make it better soon.

packages/chaire-lib-frontend/src/components/pages/index.ts
greenscientist: ok, missed as import instead of export

tahini:
I'll work on #54 first

greenscientist:

PR #614, opened 2023-03-20
config: Split server and project config

tahini:
Part 1.2 of #97

This ensures that the server specific configuration is not leaked to the
client side and not used in the common/frontend classes and files.

Rename the `server.config.js` file in chaire-lib-backend to simply
`config.js` as it loads both server and project configuraiton.

On the server side, ServerConfiguration is now a completely different
type and object than the ProjectConfiguration. They are both initialized
at the server side's `config.js` first import, from the
application's config file.


tahini:
Includes previous PR

greenscientist (approved):
Split is ok. 
I would debate what should go into server and waht should be in project, but let's do that later when the infra is in place.

greenscientist:

PR #619, opened 2023-03-23
batchRoute: Save results to DB, then generate the output files at the end

tahini:
This adds a `tr_batch_route_results` table to the database, which has 3 fields: jobId, tripIndex and data. The data field is the raw output of the od trip calculation.

The routeOdTrip function now simply returns the result and does not compute the csv and geojson outputs, this raw result is saved to the database. After all trips have been calculated, the result files are generated from the output in the database and the records are deleted.

TODO: Implement a checkpoint mechanism to resume a running task (calculations can be done out of order if there is multiple trRouting instances, so we can't simply count the results)


locales/en/notifications.json
```diff
@@ -73,7 +73,8 @@
   "ParsingCsv": "Parsing CSV file",
   "ParsingCsvWithLineNumber": "Parsing CSV file row",
   "StartingRoutingParallelServers": "Starting parallel routing servers",
-  "StoppingRoutingParallelServers": "Stoping parallel routing servers",
+  "StoppingRoutingParallelServers": "Stopping parallel routing servers",
```
kaligrafy, line 76: Why reversing the name between the attribute and the text?

packages/transition-backend/src/models/db/batchRouteResults.db.queries.ts
```diff
@@ -0,0 +1,162 @@
...
+        throw new TrError(
+            `Cannot insert result ${result.tripIndex} for job ${result.jobId} in table ${tableName} database (knex error: ${error})`,
+            'DBBRR0001',
+            'DatabaseCannotCreateBecauseDatabaseError'
+        );
+    }
+};
+
+const attributesParser = ({
+    job_id,
```
kaligrafy, line 51: Is there a reason to use an integer for job id but uuid everywhere else?

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -219,16 +234,55 @@ class TrRoutingBatch {
...
+                pageSize
+            });
+            totalCount = total;
+            for (let i = 0; i < tripResults.length; i++) {
+                const processedResults = await generateFileOutputResults(
+                    tripResults[i].data,
+                    routing.attributes.routingModes,
+                    {
+                        exportCsv: true,
+                        exportDetailed: this.demandParameters.detailed === true ? true : false,
```
kaligrafy, line 257: ? true : false seems superfluous

locales/en/notifications.json
tahini: What do you mean? I just fixed a typo in stoping here

packages/transition-backend/src/models/db/batchRouteResults.db.queries.ts
kaligrafy: OK :-)

kaligrafy (approved):

PR #622, opened 2023-03-24
Fix migrate-test script missing config parameter

nik498:
This PR adds a fix for the setup of the test DB that was already implemented for the migration.

To test : Sequential tests pass and fail randomly on my end, but seem to always pass on other computers.


tahini (approved):
Thanks!

PR #625, opened 2023-03-24
batchRoute: Allow batch routing jobs to be resumed

tahini:
Add an internal_data field to jobs to keep the checkpoint, and eventually all other internal data that the user is not meant to see.

Implement a checkpoint mechanism to track completion of potentially out of order steps of a task.

Let the batchRoute function use this checkpoint tracker mechanism to resume the task

Note that the results are not deleted from the database anymore, as there may be a point in the job execution where the files are generated, the results are deleted, but the job is not yet marked as completed. If the server is shut down at that moment, the results won't be recoverable for this job. We would need some other mechanism to properly complete the task. Besides, jobs are not meant to be kept forever, they will fill the user quota anyway. Results will be deleted when the job is deleted.

### DIFF
--- a/packages/transition-backend/src/api/__tests__/jobs.socketRoutes.test.ts
+++ b/packages/transition-backend/src/api/__tests__/jobs.socketRoutes.test.ts
@@ -47,7 +47,7 @@ describe('Jobs list', () => {
     });
 
     test('List jobs properly', (done) => {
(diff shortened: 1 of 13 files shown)


tahini:
Based on #619

kaligrafy (approved):

PR #632, opened 2023-03-29
Create scorecard-openssf.yml

greenscientist:
Add a scan for the openssf scorecards (securityscorecards.dev) Gives us an overview of our security stance and can also warn if we add some unsecure deps

### DIFF
--- a/.github/workflows/scorecard-openssf.yml
+++ b/.github/workflows/scorecard-openssf.yml
@@ -1,0 +1,72 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard-openssf.yml
```diff
@@ -0,0 +1,72 @@
...
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  schedule:
+    - cron: '22 17 * * 4'
```
tahini, line 13: Could you comment what this means? I always need to lookup cron syntax, I don't use it enough to remember: at 22h17 the 4th of each month? every thursday?

tahini (changes requested):
Is this something like FOSSA? Our project will be listed in some third-party site investigating our code? If so, could you add some info to the commit message: How do we get to see the results of this once run? And will visitors see the run result directly from github? If so, how?

.github/workflows/scorecard-openssf.yml
greenscientist: minute         0-59
              hour           0-23
              day of month   0-31
              month          0-12 (or names, see below)
              day of week    0-7 (0 or 7 is Sun, or use names)

So this is every wednesday at 17h22

greenscientist:
Just as a general comment, this is the default file that is generated by the tool. I'll add more in the commit message.

kaligrafy:
Any update on this?

greenscientist:
I should update this!

tahini (approved):

PR #635, opened 2023-03-30
config: Allow applications to specify and parse configuration

tahini:
Part 1.3 of #97

When initializing the configuration, the applications can pass functions
that parse the server and project configuration respectively, as
received from the configuration file. These function can handle
validation of the configuration fields, set their default values and are
expected to return a valid configuration object for the application.


tahini:
Based on #614

greenscientist:

PR #638, opened 2023-04-12
batchCalculation: Refactor types

tahini:
* Types related to batch calculation are moved from chaire-lib to transition-common
* Rename types that relate to the demand to contain only the word "Demand" instead of Batch (which is for calculation)
* Move to transition-frontend some types that are used only for the forms in the frontend
* Move the detailed and withGeometries parameters to the BatchCalculationParameters instead of being part of the demand

This is a pre-requisite step to implement OD Trip (demand) import and management

### DIFF
--- a/packages/transition-common/src/services/batchCalculation/types.ts
+++ b/packages/transition-common/src/services/batchCalculation/types.ts
@@ -1,4 +1,5 @@
 import { RoutingOrTransitMode } from 'chaire-lib-common/lib/config/routingModes';
+import { ErrorMessage } from 'chaire-lib-common/lib/utils/TrError';
 import {
     TransitRoutingQueryAttributes,
     validateTrQueryAttributes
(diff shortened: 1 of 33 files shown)


packages/transition-common/src/services/batchCalculation/types.ts
```diff
@@ -1,26 +1,53 @@
...
-    }
-    if (parameters.routingModes.includes('transit')) {
+    } else if (parameters.routingModes.includes('transit')) {
         const { valid: queryAttrValid, errors: queryAttrErrors } = validateTrQueryAttributes(parameters);
         if (!queryAttrValid) {
             parametersValid = false;
             errors.push(...queryAttrErrors);
         }
     }
+    if (typeof parameters.cpuCount !== 'number' && typeof parameters.maxCpuCount === 'number') {
```
kaligrafy, line 22: Shouldn't this be moved to an external function, since it is used at multiple places?

kaligrafy (approved):

packages/transition-common/src/services/batchCalculation/types.ts
kaligrafy: OK, but for some users like me, I would like to keep the CPUCount parameter available.

tahini:
The cpu count field is back in batch calculation fwiw

PR #640, opened 2023-04-17
Aesthetic changes to lines and nodes

florencelauer:
Offset lines when there are overlaps to better see where each line goes. Offset stop nodes to be in the middle of crossing lines to see which lines have a stop there.

Update the visual dynamically depending on the lines and nodes shown on screen. 

TODO: 
- A better sorting algorithm to decide the order of the offsetted lines
- Add a user configurable minimum zoom level for which the algorithm is applied.

Limits: 
- The current offset is applied on a perfect match between the overlapped segment coordinates and the line coordinates. 
- The detected overlaps are currently for perfect matches between two segments. This can be adjusted in turf.lineOverlap function.


tahini (commented):
Here's a screencast to illustrate the bug where the nodes offset ends up too big because of a line error. This is the 368 Mont-Royal. I've seen some other glitches like that in other places. The node is too far away from its original position. I think it should be offset at max 50m ?

[Uploading Capture vidéo du 04-17-2023 02:22:34 PM.webm…]()

packages/chaire-lib-frontend/src/services/map/events/GlobalMapEvents.ts
```diff
@@ -6,10 +6,17 @@
...
 
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import Preferences from 'chaire-lib-common/lib/config/Preferences';
 import { MapEventHandlerDescription } from '../IMapEventHandler';
+import { getLinesInView, offsetOverlappingLines } from 'chaire-lib-common/lib/services/geodata/ManageOverlappingLines';
+import { getNodesInView, manageRelocatingNodes } from 'chaire-lib-common/lib/services/geodata/RelocateNodes';
+
+// TODO: Make zoomLimit modifiable by user
+const zoomLimit = 14; //Zoom levels smaller than this will not apply line separation
+let applyAestheticChangesNonce: Object = new Object();
```
tahini, line 19: To avoid the lint error, you can try using the Record<string, unknown> type instead. If it doesn't work a line above saying `// eslint-disable-next-line @typescript-eslint/ban-types` would be acceptable in this case

florencelauer: I disabled it, as done in TransitRoutingForm.

packages/chaire-lib-frontend/src/services/map/events/GlobalMapEvents.ts
```diff
@@ -61,4 +73,46 @@ const globalEventDescriptors: MapEventHandlerDescription[] = [
     { type: 'map', eventName: 'mousemove', handler: onMouseMove }
 ];
 
+const applyAestheticChanges = async (boundsGL: MapboxGL.LngLatBounds, zoom: number): Promise<void> => {
+    const localNonce = (applyAestheticChangesNonce = new Object());
```
tahini, line 77: Here please add the following block (but comment it when testing locally ;-) ). This will hide the feature behind a preference. I'll add the preference editing code in a separate PR

```
if (!Preferences.get('features.map.prettyDisplay', false)) {
   return;
}
```

tahini:
Replaced with #659, which squashed all the commits together with co-authors. Thanks for the contribution!

PR #643, opened 2023-04-18
Deck.gl proof of concept

florencelauer:
Working layers using geojson data, events and custom shader using deck.gl.

The code is located in deck-gl-pow/app.js. The custom circle shader is in deck-gl-pow/scatter-plot-custom-layer.js.


davidmurray:
We will close this PR in favor of https://github.com/chairemobilite/transition/pull/763, which adds a custom shader for the path animation. Thanks for the initial work! 🙂

PR #656, opened 2023-05-11
chaire-lib: Allow applications to specify menu items under users

tahini:
The current user menu, when clicking on the username on the header, contains a single element, that resets the session. It was done quickly for a requirement in Octavi, but it does not make much sense in Transition or Evolution surveys.

This commit adds a mechanism in the application configuration, where the application can set various menu items that will go in this user menu. In the Transition configuration, we have a single menu item that opens the Preferences edit form.


tahini:
@kaligrafy In Evolution, this will allow surveys to add menu items where the user can select an interview mode (eg. phone interviewer or participant)

kaligrafy (approved):

PR #659, opened 2023-05-12
UI: Display split lines passing by a same segment

tahini:
Offset transit paths on the map to better see where each goes. Offset
stop nodes to be in the middle of crossing paths to see which paths have
a stop there.
    
It is inspired by this article for by the transit app developers:
https://blog.transitapp.com/how-we-built-the-worlds-prettiest-auto-generated-transit-maps-12d0c6fa502f/
    
But instead of image processing, the geojson data is processed directly
and modified before display.
    
This feature is experimental and needs the `features.map.prettyDisplay`
preference to be set.
    
Also, the transit nodes by which paths pass are placed such that they
are in the middle of the lines.
    
The overlapping segments take time to calculate, it is done
asynchronously. After zooming or panning, it may take a few seconds
before the map shows the pretty path
    
Future work:
    
* Improve performance of the overlapping lines
* Cache the overlapping segments to avoid having to recalculate on every
  map refresh
* Find a better sorting algorithm to decide the order of he ofsetted
  lines
* Add a use configurable minimum zoom level for which the algorithm is
  applied (now hardcoded zoom limit of 14)
* There are still a few bugs in the display
* Add more tests
* Make clearer which lines stop at nodes or not (for example a segment with
  express and non-express routes)

Limits:
    
* The current offset is applied on a perfect match between the overlapped segment coordinates and the line coordinates. This can be adjusted in turf.lineOverlap function.

### DIFF
--- a/packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionFeatures.tsx
+++ b/packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionFeatures.tsx
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2023, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import React from 'react';
+import Collapsible from 'react-collapsible';
+import { withTranslation } from 'react-i18next';
+import _toString from 'lodash.tostring';
+import PreferencesResetToDefaultButton from '../PreferencesResetToDefaultButton';
+import InputWrapper from 'chaire-lib-frontend/lib/components/input/InputWrapper';
+import { InputCheckboxBoolean } from 'chaire-lib-frontend/lib/components/input/InputCheckbox';
+import PreferencesSectionProps from '../PreferencesSectionProps';
+
+const PreferencesSectionFeatures: React.FunctionComponent<PreferencesSectionProps> = (
+    props: PreferencesSectionProps
+) => {
+    const prefs = props.preferences.getAttributes();
+
+    return (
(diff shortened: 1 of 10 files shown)


packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionFeatures.tsx
```diff
@@ -0,0 +1,45 @@
...
+    props: PreferencesSectionProps
+) => {
+    const prefs = props.preferences.getAttributes();
+
+    return (
+        <Collapsible trigger={props.t('main:preferences:ExperimentalFeatures')} open={true} transitionTime={100}>
+            <div className="tr__form-section">
+                <InputWrapper twoColumns={true} label={props.t('main:preferences:MapPrettyDisplay')}>
+                    <InputCheckboxBoolean
+                        id={'formFieldPreferencesFeatureMapPrettyDisplay'}
```
greenscientist, line 26: Eventually, this list should be auto generated from a feature list defined elsewhere. (So that we don't have to change the UI for each experimental feature we add)

greenscientist:
I would have added a thanks and note that this project was done by a group of students from Polytechnique. (I know you put the name, but having more context could be nice)

Otherwise I don't have much comment. It's mostly UI stuff, so a bit outsite my expertise.

tahini:
I added a note in the main commit about the context of the project.

greenscientist:
@kaligrafy, can you have a look?

kaligrafy (approved):

tahini:
I just noticed this is still not merged! Let's push it

PR #670, opened 2023-05-18
preferences: Let the class have change listeners

tahini:
Some preferences methods, like load and update had an optional eventManager parameter which was used to notify of preferences changes. But interested classes need to know about this eventManager and all callers of preferences update/load needed to make sure to pass this specific eventManager. Now, they can register directly to the preferences class, which handles notifications.


tahini:
With this PR, Evolution, who doesn't have a central event manager, can now listen to preferences updates.

kaligrafy (approved):

PR #694, opened 2023-07-31
zones: Fix zone import when properties are not string

tahini:
The shortname substring was causing an exception if the property was, for example, an integer

### DIFF
--- a/packages/chaire-lib-backend/src/tasks/zones/__tests__/importZonesFromGeojson.test.ts
+++ b/packages/chaire-lib-backend/src/tasks/zones/__tests__/importZonesFromGeojson.test.ts
@@ -309,7 +309,7 @@ describe('Correct calls', () => {
                 properties: {
                     property1: 'a',
                     property2: `b${idx}`,
-                    property3: 3
+                    property3: '3'
                 }
             }))
         };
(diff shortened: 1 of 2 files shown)


kaligrafy:
What happens if we import a land role file, with id_provinc looking like this: "54025365143195500000000"? This is larger than most standard integer and must be saved as big int or string

greenscientist:
From what I understand in the code, it's now converted as a string, so should be good, right?

greenscientist:
The commit message could be more explicit about the change. Right now it only state the problem, not the applied solution.

kaligrafy:
Yeah, but sometimes, we should import data as integer, example:
zone A;
population: 2345
flats: 1342

Would it be possible to add as integer if it fits the standard javascript integer size and if not, just convert to string?

tahini:
I updated the commit message

kaligrafy:
Yes, true, if we import a geojson, it just keep the values as they were in the geojson, but we should at least check that the integer va.ue in the geojson is not overflowing the javascript integer size, beacue it will be truncated in any calculation.

tahini:
@kaligrafy you can open an issue for this, as it is not related to the shortname bug here

PR #698, opened 2023-08-07
utils: Add an async function which fails after a timeout

tahini:
The function receives a promise and a timeout as parameter and if the timeout is reached before the promise completes, the function throws an error.

Javascript taken from https://javascript.plainenglish.io/how-to-add-a-timeout-limit-to-asynchronous-javascript-functions-3676d89c186d


tahini:
@kaligrafy This should allow to add a timeout for https://github.com/chairemobilite/od_mtl_2023/issues/770

kaligrafy (approved):

PR #715, opened 2023-09-14
Feature/accessibility map

nik498:
Prototype of accessibility maps rasterizer. Rasterizer Python script is fully working, but integration into Typescript code and running it in browser is still to be implemented. The rest of the implementation should be completed.

Commits are also to be rebased and squashed.

Tested with RTL data.


greenscientist:
We will revisit this, linked this PR in the original issue.

PR #717, opened 2023-09-20
dependencies: Replace lodash per-method packages by latest

### LINKED ISSUE #716: Use latest lodash for all lodash packages
1- per-method packages are deprecated and often outdated, using the main lodash package will get latest of each method
2- Each per-method package bundles the necessary part of the lodash package that it needs, some parts are thus imported multiple times
3- We use quite a few lodash packages, it is worth consolidating

ftr, here's the size taken by the node_modules part of the development package of Transition

Capture d’écran du 2023-09-20 08-20-34

tahini:
fixes #716

Fix a few compilation errors with the lodash update and remove un-used lodash imports.


tahini:
I tested with a basic transition use case and it still works as expected

kaligrafy (approved):

PR #740, opened 2023-10-25
downloadOsmNetworkData: create output directory if it does not exist

### LINKED ISSUE #739: downloadOsmNetworkData: fail if output directory does not exist
With a brand new instance of Transition, the `runtime/imports` does not exist. The download task will fail:
```
davidmurray@Davids-MBP-2 transition % yarn node --max-old-space-size=4096 packages/chaire-lib-backend/lib/scripts/osrm/downloadOsmNetworkData.task.js --polygon-file polygon_mtl.geojson 
yarn node v1.22.19
Using .env file from /Users/davidmurray/code/poly/transition/.env
Read server configuration from /Users/davidmurray/code/poly/transition/config.js
Retrieving data from osm...
Done retrieving data from osm...
Writing osm data to /Users/davidmurray/code/poly/transition/runtime//imports/osm_network_data.osm
Error download from API [Error: ENOENT: no such file or directory, open '/Users/davidmurray/code/poly/transition/runtime//imports/osm_network_data.osm'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/davidmurray/code/poly/transition/runtime//imports/osm_network_data.osm'
}
task did not complete correctly [Error: ENOENT: no such file or directory, open '/Users/davidmurray/code/poly/transition/runtime//imports/osm_network_data.osm'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/davidmurray/code/poly/transition/runtime//imports/osm_network_data.osm'
}
✨  Done in 5.15s.
```

davidmurray:
Otherwise, the download will fail since fs.createWriteStream cannot write to the inexistant directory.
Fixes #739

### DIFF
--- a/packages/chaire-lib-common/src/tasks/osrm/downloadOsmNetworkData.ts
+++ b/packages/chaire-lib-common/src/tasks/osrm/downloadOsmNetworkData.ts
@@ -4,6 +4,7 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
+import fs from 'fs';
 import inquirer from 'inquirer';
 
 import osmDownloader from '../../utils/osm/OsmOverpassDownloader';
@@ -68,6 +69,7 @@ export class DownloadOsmNetworkData implements GenericTask {
         // Make sure the number of retries is between 0 and 6
         const retryAttempts = isNaN(retry) ? 0 : Math.min(Math.max(retry, 0), 6);
         const importDir = this._fileManager.directoryManager.projectDirectory + '/imports/';
+        fs.mkdirSync(importDir, { recursive: true });
         const polygonFile = filePath ? (filePath as string) : importDir + 'polygon.geojson';
         const osmRawDataFile = osmRawFile ? (osmRawFile as string) : importDir + 'osm_network_data.osm';
 


packages/chaire-lib-common/src/tasks/osrm/downloadOsmNetworkData.ts
```diff
@@ -68,6 +69,7 @@ export class DownloadOsmNetworkData implements GenericTask {
         // Make sure the number of retries is between 0 and 6
         const retryAttempts = isNaN(retry) ? 0 : Math.min(Math.max(retry, 0), 6);
         const importDir = this._fileManager.directoryManager.projectDirectory + '/imports/';
+        fs.mkdirSync(importDir, { recursive: true });
```
tahini, line 72: I think it's the directory manager's role to make sure the project directory is created. It does not make sense that every possible task that can be run before the server first starts needs to make sure the directory exists. Besides, this is in `common` and you should not import fs here.

For info, currently, all directory creations are done at server start in the serverApp.ts file

tahini (changes requested):

packages/chaire-lib-common/src/tasks/osrm/downloadOsmNetworkData.ts
davidmurray: good point!
So serverApp.ts is already creating that directory: `directoryManager.createDirectoryIfNotExists('imports')`
So this PR is not necessary, I just had to launch Transition once before running the task.

Should I update the README to mention that?

davidmurray: currently the steps in README.md are in the following order:
1. Create a config file
2. [Download and prepare the road network](https://github.com/chairemobilite/transition#download-and-prepare-the-road-network)
3. [Prepare the environment file](https://github.com/chairemobilite/transition#prepare-the-environment-file)
4. ...
5. [Start the Node.js server](https://github.com/chairemobilite/transition#start-the-nodejs-server)

so it seems like step 5 should be done before step 2?

greenscientist: I don't think we should have to run transition before the task (I run the task independently in some of our infra. 
I don't know if it should create the directly, but the task should at least fail cleanly, so it should at least check if the directory is present and print an error message saying that the directly is invalid and suggest appropriate steps.

greenscientist:
We will keep only https://github.com/chairemobilite/transition/pull/1150 for now

PR #751, opened 2023-11-02
Small UI fixes

davidmurray:


kaligrafy (commented):
seems nice, but can you add some screenshots?

davidmurray:
@kaligrafy :

 
Before:
image

After:
image

--
Before:
image

image

After:
image

image


--
Before:
image
After:
image


--
Before:
image
After:
image

kaligrafy:
OK, nice, but please keep the lat/lon as lat,lon because in at least 80% of the time, this the most useful format. Or else add a way to provide both lon/lat and lat/lon

davidmurray:
Actually, about that, I am wondering what is the use of this? Since as soon as your mouse exits the map, the coordinates disappear so you can’t copy anyway. But i can add the comma

tahini:
For select boxes, there is now no visible difference between a normal textbox and a select box. Was there ever an arrow for selects in Transition? I don't now, but you're bringing attention to it now

Capture d’écran du 2023-11-02 13-29-38

kaligrafy:
> Actually, about that, I am wondering what is the use of this? Since as soon as your mouse exits the map, the coordinates disappear so you can’t copy anyway. But i can add the comma

oh, yes, this should be copyable... but we need a way to freeze it for copy. So for now, keep as before with lat,lon and we will think about this later.

greenscientist:
Thoughts for future discussion: Maybe the display of lat/long or long/lat should be configurable in the user preference.

There should be a standard unit to display to identify which is lat and which is lon. (But I know, there's is none)

kaligrafy:
I think we should just have a drop-down menu (right-click) on map that makes it possible to copy the clicked location witht these formats:

lat,lon: 45.5,-73.6
lat/lon: 45.5/-73.6
lat lon: 45.5 -73.6
lon,lat: -73.6,45.5
lon lat: -73.6 45.5

which would work in almost all scenarios (geojson lon,lat,  usual google lat,lon, openstreetmap url lat/lon, and poly with spaces instead of commas)

davidmurray:
> For select boxes, there is now no visible difference between a normal textbox and a select box. Was there ever an arrow for selects in Transition? I don't now, but you're bringing attention to it now
> 
> Capture d’écran du 2023-11-02 13-29-38

Good catch!
the problem is that we set the background color to black, so we can't see the little arrow. But it is there!
image

Will try to fix this in this PR

davidmurray:
ready for merge @kaligrafy

packages/chaire-lib-frontend/src/styles/_form.scss
```diff
@@ -64,11 +64,24 @@ textarea, ._input-textarea {
...
...BpbmtzY2FwZTp6b29tPSIyMy42IgogICAgIGlua3NjYXBlOmN4PSIyLjQ3ODgxMzYiCiAgICAgaW5rc2NhcGU6Y3k9IjQuOTc4ODEzNiIKICAgICBpbmtzY2FwZTp3aW5kb3ctd2lkdGg9IjE0NDAiCiAgICAgaW5rc2NhcGU6d2luZG93LWhlaWdodD0iNzI4IgogICAgIGlua3NjYXBlOndpbmRvdy14PSIwIgogICAgIGlua3NjYXBlOndpbmRvdy15PSIxMzMiCiAgICAgaW5rc2NhcGU6d2luZG93LW1heGltaXplZD0iMCIKICAgICBpbmtzY2FwZTpjdXJyZW50LWxheWVyPSJMYXllcl8xIiAvPgogIDxkZWZzCiAgICAgaWQ9ImRlZnMxIj4KICAgIDxzdHlsZQogICAgICAgaWQ9InN0eWxlMSI+LmNscy0xe2ZpbGw6I2ZmZjt9LmNscy0ye2ZpbGw6IzQ0NDt9PC9zdHlsZT4KICA8L2RlZnM+CiAgPHRpdGxlCiAgICAgaWQ9InRpdGxlMSI+YXJyb3dzPC90aXRsZT4KICA8cG9seWdvbgogICAgIGNsYXNzPSJjbHMtMiIKICAgICBwb2ludHM9IjEuNDEgNC42NyAyLjQ4IDMuMTggMy41NCA0LjY3IDEuNDEgNC42NyIKICAgICBpZD0icG9seWdvbjEiCiAgICAgc3R5bGU9ImZpbGw6I2ZmZmZmZjtmaWxsLW9wYWNpdHk6MSIgLz4KICA8cG9seWdvbgogICAgIGNsYXNzPSJjbHMtMiIKICAgICBwb2ludHM9IjMuNTQgNS4zMyAyLjQ4IDYuODIgMS40MSA1LjMzIDMuNTQgNS4zMyIKICAgICBpZD0icG9seWdvbjIiCiAgICAgc3R5bGU9ImZpbGw6I2ZmZmZmZjtmaWxsLW9wYWNpdHk6MSIgLz4KICA8bWV0YWRhdGEKICAgICBpZD0ibWV0YWRhdGEyIj4KICAgIDxyZGY6UkRGPgogICAgICA8Y2M6V29yawogICAgICAgICByZGY6YWJvdXQ9IiI+CiAgICAgICAgPGRjOnRpdGxlPmFycm93czwvZGM6dGl0bGU+CiAgICAgIDwvY2M6V29yaz4KICAgIDwvcmRmOlJERj4KICA8L21ldGFkYXRhPgo8L3N2Zz4K);
```
kaligrafy, line 81: ce serait pas préférable de mettre l'image en format png  ou svg dans le repo directement?

config.js
kaligrafy, line 1: Why this file was added?

packages/chaire-lib-frontend/src/styles/_form.scss
davidmurray: Oui je peux faire ça!
Ça irait dans transition-frontent ou chaire-lib-frontent?
Pour l'instant je vois qu'il n'y aucun _asset_ dans chaire-lib-frontend. Est-ce un choix délibéré ou est-ce qu'il n'y avait simplement pas le besoin jusqu'à maintenant?
cc @tahini

config.js
davidmurray: because I messed up!
fixed it

packages/chaire-lib-frontend/src/styles/_form.scss
kaligrafy: ça irait clairement dans chaire-lib, ce n'est pas spécifique à transition

kaligrafy: et si possible prendre un svg vectoriel.

davidmurray: Bon... c'est plus compliqué que je pensais
Il y a un problème: mon icône ne semble pas être copié dans `public/dist/images/icons/interface`, même si j'ai modifié le fichier de config webpack et le package.json de chaire-lib-frontend pour copier les images au dossier lib.
:-(

Des suggestions? @kaligrafy  @tahini

packages/chaire-lib-frontend/src/styles/_form.scss
```diff
@@ -76,9 +76,9 @@ select {
...
+  background-image: url("/assets/images/icons/interface/select_chevron_white.svg");
```
davidmurray, line 81: not sure if this path would be correct

kaligrafy: Does it work? It it does, no problem there!

packages/chaire-lib-frontend/src/styles/_form.scss
kaligrafy: Déplacer dans Transition comme les autrs images, et on fixera ça plus tard. Séparer en issue distinct.

davidmurray: done! ready to merge

kaligrafy (approved):

PR #754, opened 2023-11-03
Right panel position

### LINKED ISSUE #753: Add a prefs switch to select the position of the info panel

davidmurray:
Demo: 

https://github.com/chairemobilite/transition/assets/545502/3e648025-7b40-47f5-b4bb-d6e8c67a3428

Fixes #753


kaligrafy:
Cool! Did you test with Firefox, Safari, Chrome and Edge?

kaligrafy:
There seems to be a double border on the right of the panel and none on the left when the panel is on the left.

davidmurray:
> There seems to be a double border on the right of the panel and none on the left when the panel is on the left.
It's fixed in the CSS, but when I recorded the video I was using an older version of the CSS.
There is no problem now:

image


The wider line between the panel and the map is to indicate that it's a movable divider.

kaligrafy:
> > There seems to be a double border on the right of the panel and none on the left when the panel is on the left.
> > It's fixed in the CSS, but when I recorded the video I was using an older version of the CSS.
> > There is no problem now:
> 
> image
> The wider line between the panel and the map is to indicate that it's a movable divider.

In the video, there is no border on the left, but in your image, there is one. I guess it is because you fixed it in-between?

davidmurray:
@kaligrafy Yes! Seems like my response was included in the quote :
image


Just tested with chrome, safari, firefox and edge and all works fine!

kaligrafy (approved):

PR #756, opened 2023-11-09
Fix 691: unzip files using a stream to reduce RAM load

davidmurray:
In Transition, we zip and unzip files for GTFS exporting and importing. 
I could not find a package that allowed unzipping using a stream but also allow zipping files. This PR uses https://github.com/antelle/node-stream-zip, which only allows unzipping. We must therefore keep our current dependency on JSZip.

The options are as follows:
1. Accept this PR which unzips using a stream, but adds an extra package dependency
2. Spend time writing our own zip/unzip code (or find an existing library to extend/improve)
3. Drop both packages (JSZip and node-stream-zip) and instead ask users to install the unix `zip` utility on their workstations (it is preinstalled on macOS and probably most linux distros anyway)

### DIFF
--- a/packages/transition-backend/package.json
+++ b/packages/transition-backend/package.json
@@ -49,6 +49,7 @@
         "lodash": "^4.17.21",
         "moment": "^2.29.4",
         "morgan": "^1.10.0",
+        "node-stream-zip": "^1.15.0",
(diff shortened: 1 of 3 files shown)


tahini (approved):
Please rebase this PR for the CI to pass, but looks good

PR #759, opened 2023-11-09
Fix 305

davidmurray:
Before:
image

After:
image


tahini:
Thanks, this fixes an alignment, but the issue #305 was related to the TransitServiceFilter component (that you can see when selecting services in the scenario edit form). I don't now if you want to fix that one also in this PR? Or remove the link to the issue in the commit message

davidmurray:
@tahini Good point! Hadn't noticed that this was in a different section. In any case, that bug seems to be gone because it looks like this now:

image

davidmurray:
i will update this commit's message to remove the link to the issue and we can close #305 independently

tahini (approved):

PR #763, opened 2023-11-13
Deck gl shader

davidmurray:
This PR adds a shader to Deck.gl to add a path animation like this:

https://github.com/chairemobilite/transition/assets/545502/55c5656f-e722-4574-b652-fd5684e5dcd4


kaligrafy:
obsolete, see #773 instead

PR #765, opened 2023-11-13
trRouting: Use v2 api for accessibility map

tahini:


packages/chaire-lib-common/src/services/trRouting/TrRoutingService.ts
```diff
@@ -173,13 +184,36 @@ const apiV2ToRouteResult = (apiResult: TrRoutingApi.TrRoutingV2.RouteSuccessResu
     };
 };
 
+const apiV2ToAccessMapResult = (
```
greenscientist, line 187: I think that should be in the backend

greenscientist (approved):
Code changes looks fine to me. 

But a general comment. The API exposed to the frontend from probably be a generic one agnostic of TrRouting. (If we ever want to expose multiple engine, we'd need that.) Any conversion should be handled directly in the backend. 
No need to change things now, but to keep in mind for future changes.

tahini:
We'll need an extra layer of abstraction when we support many tools. `TrRoutingService`, which calls the socket routes or fetch urls is the one called by consumers and has a trRouting agnostic return type (though that is still much inline with the trRouting v1 original API and should be revisited). But the socket route/fetch url receiver should probably be the one to dispatch to the right tool and return and handle properly the return value. I'll open 2 issues, after fixing this 1 issue to track all this.

PR #766, opened 2023-11-13
Remove references to trRouting v1

tahini:


tahini:
Based on #765, the last commit remove all calls to the trRouting v1 API

greenscientist:
That's a nice red diff

greenscientist (approved):

PR #777, opened 2023-11-22
Add DistanceUnitFormatter and DurationUnitFormatter, two react compon…

### LINKED ISSUE #775: Add unit conversion react components
Transition displays distances and time durations in many places.
If these values are small, it usually makes more sense to display in meters and seconds.
However, when the values get big, kilometres and minutes (or hours) are usually more appropriate. We could have some react components that automatically convert the values to the best unit for display. 

Here are a few examples where a better unit could improve legibility:

1. Layover should be in minutes if > 60s
image

2. Cumulative distance should switch to kilometres if > 1000 m:
image

3. Transit routing results (here it should be in hours and kilometres)
image

Thoughts and suggestions @kaligrafy  @tahini ?

davidmurray:
…ents to automatically choose the best unit for a value

They also allow clicking on the values to cycle between the different possible units. Example usage is shown in transit routing output for the walking mode. If code and UX seems satisfactory to you, I will use these components in many other parts of transition.

Demo:

https://github.com/chairemobilite/transition/assets/545502/b3cd1a0c-8627-4cf4-87f6-093245efe79e

Fixes #775


davidmurray:
I could also add one for speeds : m/s, km/h, mph, etc

kaligrafy (approved):
Verify code lint though

kaligrafy:
Very nice!

davidmurray:
> Verify code lint though

It was better before I ran `yarn format`
If we don't like this formatting we have to change the parameters to prettier eslint!

davidmurray:
added `SpeedUnitFormatter` too

packages/chaire-lib-frontend/src/components/pageParts/DistanceUnitFormatter.tsx
```diff
@@ -0,0 +1,61 @@
...
+export interface DistanceUnitFormatterProps extends WithTranslation {
+    value: number;
+    sourceUnit: 'kilometers' | 'meters';
+    destinationUnit?: destinationUnitOptionsType;
+}
+
+const DistanceUnitFormatter: React.FunctionComponent<DistanceUnitFormatterProps> = (
+    props: DistanceUnitFormatterProps
+) => {
+    const [destinationUnit, setDestinationUnit] = useState<string | undefined>(props.destinationUnit);
```
tahini, line 24: Instead of `string`, you can use your destinationUnitOptionType

packages/chaire-lib-frontend/src/components/pageParts/DurationUnitFormatter.tsx
```diff
@@ -0,0 +1,66 @@
...
+export interface DurationUnitFormatterProps extends WithTranslation {
+    value: number;
+    sourceUnit: 'seconds' | 'minutes' | 'hours';
+    destinationUnit?: destinationUnitOptionsType;
+}
+
+const DurationUnitFormatter: React.FunctionComponent<DurationUnitFormatterProps> = (
+    props: DurationUnitFormatterProps
+) => {
+    const [destinationUnit, setDestinationUnit] = useState<string | undefined>(props.destinationUnit);
```
tahini, line 24: Same here, use destiantionUnitOptionsType

tahini (changes requested):
Looks good! You can further type the states though, to limit to the possible values.

davidmurray:
@tahini  fixed :)

tahini (approved):
Thanks!

PR #781, opened 2023-11-27
Update Debian base to bookworm in docker build

greenscientist:


greenscientist:
It builds locally, with the latest update to trRouting to also use bookworm. This docker build will fail, since I have not pushed the trRouting update yet. I'll need to do them in sync.

PR #788, opened 2023-11-29
json2capnp: listen to ipv6 instead of ipv4

greenscientist:
Modern linux will try to connect to localhost on IPv6 first.

Further fix will be to listen on both


tahini (changes requested):
Wait for a solution that covers both ipv4 and v6

PR #798, opened 2023-12-06
Type the event manager and the `map.updateLayer` event

tahini:
* eventManager: Add methods to listen and emit typed events

This allows to use a type for the events, with a `name` and `args` fields for the name of the event and types of arguments in the callback.
    
The methods are generic, with the generic type being the type of the event. The listening method is `onEvent`, the emitting one is `emitEvent`. It can be called in the code with
    
`eventManager.emitEvent<MyEventType>('event.myevent', myArgs)`
`eventManager.onEvent<MyEventType>('event.myevent', (myArgs) => void)`

* map: type the `map.updateLayer` event

Use the new `onEvent` and `emitEvent` methods of the EventManager to type the calls to `map.updateLayer`. It receives an argument with 2 fields: `layerName` and `data`, which are respectively the layer to update and the geojson feature collection for the layer (or function to retrieve it).

### DIFF
--- a/packages/chaire-lib-common/src/services/events/EventManager.ts
+++ b/packages/chaire-lib-common/src/services/events/EventManager.ts
@@ -10,7 +10,43 @@ import Event from './Event';
 
 const prefix = '';
 
+export type EventType = {
+    name: string;
+    arguments: { [key: string]: unknown };
+};
+
+export type EventNameKey = 'name';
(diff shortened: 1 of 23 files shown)


tahini:
@kaligrafy ping on this review

kaligrafy (approved):

Issue #159, opened 2020-12-03
transition: trRouting: Allow different mode for access, egress and transfers => multimodal


kaligrafy:
Allow bimodal trips to be calculated (access/egress by car or bicycle, and transfers and egress/acess by foot)
trRouting needs to be updated too.

PR #806, opened 2023-12-13
Move transferable nodes to database

tahini:
* Add a database table for transferable node, with origin node id, destination node id, distance and travel time
* Add a database query to get transit node within a certain distance of a point
* Move the node save functionality to the backend instead of common. Transferable nodes are updated at that moment. A new socket route `transitNode.save` take the node's attribute and whether the geography has changed.
* Transferable node calculation functions in common's NodeGeographyUtils are moved in the TransferableNodesUtils in the backend.


packages/transition-backend/src/models/db/migrations/20231206083200_createTransferableNodeTbl.ts
```diff
@@ -0,0 +1,25 @@
...
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { Knex } from 'knex';
+
+const tableName = 'tr_transit_node_transferable';
+
+export async function up(knex: Knex): Promise<unknown> {
+    return knex.schema.createTable(tableName, (table: Knex.TableBuilder) => {
+        table.uuid('origin_node_id').notNullable().index(); // required
```
kaligrafy, line 13: Possible to add mode here, since later on, we could use bicycle or even car as transfer mode

kaligrafy (changes requested):

packages/transition-backend/src/models/db/migrations/20231206083200_createTransferableNodeTbl.ts
greenscientist: Do we need to add it now, or when we actually plan to use it ? Since we don't really know what it will look like, it's harder to do if it's the right design

tahini: Indeed, there's a lot to change to support various modes. We'll do that later

kaligrafy: PLease add an issue :-)

kaligrafy (approved):

packages/transition-backend/src/models/db/migrations/20231206083200_createTransferableNodeTbl.ts
tahini: #159 already covers this case

greenscientist (commented):
I might have missed it, but when we updateAllNodes, is there a need to clear the nodes already in the DB ?

tahini:
> I might have missed it, but when we updateAllNodes, is there a need to clear the nodes already in the DB ?

The database query for the transferable nodes takes ALL transferable nodes, so they are deleted before being inserted (see comment in query file as to why)

greenscientist (approved):

tahini:
ftr, for the whole region of Montreal, there are 13230 nodes, and 380779 transferable nodes, which take about 37MB of disk space (for 17MB for the nodes table)

Issue #813, opened 2023-12-18
tasks: Add end time and/or duration to batch jobs


tahini:
This information is available with the updated_at field in the executable job object.

tahini:
#815 is a requirement

PR #814, opened 2023-12-18
jobs: Add a column for end time/duration

tahini:


tahini:
It looks like this

Capture d’écran du 2023-12-18 14-07-00

I'm still waiting for suggestions to make this job list prettier, maybe the docker build list (but we don't see the duration)?

greenscientist:
The docker build list for exmple only has a start time and a duration. 
out of a start_time, end_time and duration, you only really need 2. The other can be infered with the other. 
Personnally, I think the duration is the field that people are probably most interested in

kaligrafy (approved):

tahini:
And that might be a problem too if the task has spent time in a pending state... this PR gives a duration from last updated_at to created_at, but we don't have a started_at field...

tahini:
But adding a started_at is a bit more involving than this quick one...

greenscientist:
So the current displayed duration might be a lie?  
I was under the impression that this PR looked too easy for what it was...

tahini:
#815 needs to be fixed first so we can calculate the actual duration of the job execution and not just it's life

tahini:
Actually, I can just put the end time in this PR for now, without fixing either #813 or #815. it gives an information to the user, without lying about anything. Would that be all right? @greenscientist ?

greenscientist:
Yes, that would work well for the moment and give a good enough insight to the user

tahini:
Only the end time is present now

greenscientist (approved):

PR #818, opened 2023-12-21
Set memory limit option in the right order of the node CLI

greenscientist:
We need to pass node arguments before the script filename, otherwise the arguments are passed to the script.

From the Node doc:
node [options] [V8 options] [<program-entry-point> | -e "script" | -] [--] [arguments]

Maybe this worked when we were using yarn babel-node, but it's not working with straight up yarn node


greenscientist:
The max-old-space parameter was ignored and we were running with only 2gig of max memory. 
I was hitting the limit when importing the STM GTFS. (I have lot of previously loaded nodes, which might explain why it takes more memory here)

tahini (approved):
I guess the good news is that it worked with 1GB all this time??

PR #820, opened 2024-01-08
Add send welcome email

samuel-duhaime:
Add send welcome email

I'm going to use this, to send welcome Email for survey that have this config.
I'm going to put this sendWelcomeEmail function inside Evolution: evolution\packages\evolution-backend\src\services\auth\participantAuthModel.ts createAndSave().


.vscode/settings.json
samuel-duhaime, line 1: This is a change because, boolean is not support anymore since 1.85.0. So we need to put "explicit". It does the same thing.

PR #822, opened 2024-01-09
osrm: Use the host even for local servers

### LINKED ISSUE #821: osrm: local osrm servers listen on ipv4, but transition may look for ipv6
In most recent debian, ipv6 is looked up by default, so localhost:5000 for example will try to connect to [::1]:5000 for the osrm server. The server listens on 0.0.0.0 so it does not appear to exist. We should be able to specify the host to 127.0.0.1 in those cases

tahini:
fixes #821

This uses the value set in the mode config's host field even for local servers. This allows to specify a specific URL on which to join the server, for example an ipv4 url where 'localhost' would have resolved to an ipv6 url, or simply a named URL for the local server.

### DIFF
--- a/packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
@@ -81,7 +81,8 @@ const configureAllOsrmServers = async function (startServers = true): Promise<vo
             // Only configure mode that are enabled
             if (modeConfig.enabled === true) {
                 const port = modeConfig.port;
-                let host = '';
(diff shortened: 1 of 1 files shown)


greenscientist (commented):
Shouldn't we add the host to the restart call a bit after this?

tahini:
> Shouldn't we add the host to the restart call a bit after this?

No, the host here is used locally for the mode registry, but is not used for the osrm-routed's host parameter (-i), so it is not required in the start/restart commands.

greenscientist:
ok, yeah I see. 
Please add a bit of comments to explain why we need the hostname even when we use the autostart.

greenscientist (approved):

tahini:
It's in the commit message, but I added it in the code too

PR #824, opened 2024-01-10
preferences: Allow to set the color preferences for routing and accessibility map

tahini:


kaligrafy:
Seems good! However, I am not sure it is wise in the long term to allow different colors for each user/mode for routing. Using the same color for distinct mode in alll instances of Transition would bring some standards and make comparisons easier. Needs dicussion I think.

kaligrafy (approved):

PR #842, opened 2024-02-05
path: add tests and documentation for node/waypoint operations

### LINKED ISSUE #835: Path: Add tests and document functions for path's building
* Test adding/updating/removing waypoints and document the expected parameters
* Test adding/updating/removing nodes and document expected parameters

While refactoring the map for deck.gl, a lot of the insertion logic that was in the map events now does not need to exist. But this highlighted problems in the Path's functions for managing waypoints and nodes.

Proper tests can help identify those problems and solve them.

tahini:
fixes #835

This adds exhaustive unit tests for the waypoints insert/update/delete and nodes insert/delete operations. It also adds some check of indexes before doing the operation.

After these addition, the coverage of the Path.ts file increased as follows:

% Stmts: 27.16    --> 58.6
% Branch: 34.75  --> 49.7
% Funcs: 33.77    --> 49.33
% Lines: 27.08     --> 58.94

### DIFF
--- a/packages/transition-common/src/services/path/Path.ts
+++ b/packages/transition-common/src/services/path/Path.ts
@@ -368,23 +368,50 @@ export class Path extends MapObject<GeoJSON.LineString, PathAttributes> implemen
         return this.updateGeography();
     }
 
-    insertWaypoint(waypointCoordinates: [number, number], waypointType = 'engine', afterNodeIndex?, insertIndex?) {
+    /**
+     * Insert a waypoint at a given location
+     *
+     * @param waypointCoordinates The coordinates of the waypoint to insert
+     * @param waypointType The type of the waypoint
+     * @param afterNodeIndex The index of the node after which to insert this
+     * waypoint. If not set, it is added at the end of the path
+     * @param insertIndex The index at which to insert this waypoint. If not
+     * set, it is added at the end of the waypoints for this path
+     * @returns The updated path
+     */
+    async insertWaypoint(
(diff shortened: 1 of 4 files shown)


packages/transition-common/src/services/path/Path.ts
```diff
@@ -444,31 +533,66 @@ export class Path extends MapObject<GeoJSON.LineString, PathAttributes> implemen
...
+     * @param waypointType The type of this updated waypoint
+     * @param afterNodeIndex The index of the node after which the waypoint is
+     * located
+     * @param waypointIndex The index of the waypoint to update
+     * @returns The updated path
+     */
     updateWaypoint(
         waypointCoordinates: [number, number],
-        waypointType = null,
+        waypointType: string | undefined = undefined,
```
kaligrafy, line 555: Should we add a type here?

kaligrafy (approved):

packages/transition-common/src/services/path/Path.ts
tahini: We need to add type for the return value ideally 2 lines below, do you mean here in this line? there is a type...

kaligrafy: I mean should we replace string by possible values?

tahini: Indeed, we eventually should, but we can wait after the refactoring and maybe profile congestion to see if 'manual' | 'engine' is the complete and right type.

Issue #854, opened 2024-02-21
schedules: More frequent return trips generated at the beginning of the period


tahini:
Units are initialized at the beginning of the period, each being at an evenly spaced location in the cycle, which is not an exact fraction of the frequency.

This means all units that have been assigned somewhere in the first trip will reach the time of the return trip, start a trip and wait patiently for its correct frequency at the end of the whole cycle.

Not clear? Here's a screenshot

![Capture d’écran du 2024-02-21 15-10-51](https://github.com/chairemobilite/transition/assets/4398897/8ff6709b-33a4-4c64-9a6f-38fb2bc19f9f)

First return trip is at 4:06, 15 minutes earlier than the next at 4:21.

Issue #681, opened 2023-07-12
schedules: generating timetables do not consider previous periods


tahini:
See screenshot, the second period schedules, every 5 minutes, starts at 6:00 exactly, while the last trip of the previous period started at 5:57.

It should adapt the trips. But what if we change the previous period interval? Do we ask the user to re-generate trips for periods _after_ the current one?

![schedulesProblem](https://github.com/chairemobilite/transition/assets/4398897/4beb51c1-22eb-42ac-823d-40755291e991)

PR #858, opened 2024-02-21
gtfs: Always use custom times when inferring frequencies

tahini:
This seems to be currently more intuitive for end users, instead of using the period start/end if the diff between the time and closest trip time is lower than frequency.

Given various issues with schedules (#681, #854, the fact that all trips start at the exact period start, problems inferring frequency with asymetrical trips, to name a few), this sounds like an acceptable compromise.


tahini:
cc @tibobliss

kaligrafy (approved):

PR #862, opened 2024-02-22
Use threaded TrRouting

greenscientist:
Splitted into 3 commits: 
* Use threaded TrRouting for the normal case
* Adapt batch routing and batch accessibility map
* Fix port configuration for batch accessibility

### DIFF
--- a/packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
@@ -4,11 +4,15 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
+import os from 'os';
 import { directoryManager } from '../filesystem/directoryManager';
 import Preferences from 'chaire-lib-common/lib/config/Preferences';
 import ProcessManager from './ProcessManager';
 import osrmService from '../osrm/OSRMService';
 import config from '../../config/server.config';
+
+// Set 2 threads as default, just in case we have more than one request being handled
+const DEFAULT_THREAD_COUNT = 2;
 
 const availablePortsByStartingPort: { [startingPort: number]: { [port: number]: boolean } } = {};
 
@@ -19,6 +23,7 @@ const getServiceName = function (port) {
 const startTrRoutingProcess = async (
     port: number,
     attemptRestart = false,
+    threadCount = 1,
     parameters: { debug?: boolean; cacheDirectoryPath?: string } = { debug: false, cacheDirectoryPath: undefined }
 ) => {
     const osrmWalkingServerInfo = osrmService.getMode('walking').getHostPort();
@@ -41,6 +46,12 @@ const startTrRoutingProcess = async (
     } else {
         commandArgs.push(`--cachePath=${directoryManager.projectDirectory}/cache/${config.projectShortname}`);
     }
+
+    // Set the threads argment if we have a threadCount param and it's higher than one
+    if (threadCount && threadCount > 1) {
+        commandArgs.push(`--threads=${threadCount}`);
+    }
+
     const waitString = 'ready.';
 
     const processStatus = await ProcessManager.startProcess(
@@ -67,7 +78,7 @@ const start = async (parameters: { port?: number; debug?: boolean; cacheDirector
     const port = parameters.port || Preferences.get('trRouting.port');
 
     // TODO Check why we need this await, should not be useful before returning
-    return await startTrRoutingProcess(port, undefined, parameters);
+    return await startTrRoutingProcess(port, false, DEFAULT_THREAD_COUNT, parameters);
 };
 
 const stop = async (parameters) => {
@@ -92,7 +103,7 @@ const restart = async (parameters) => {
         };
     } else {
         // TODO Check why we need this await, should be be useful before returning
-        return await startTrRoutingProcess(port, true, parameters);
+        return await startTrRoutingProcess(port, true, DEFAULT_THREAD_COUNT, parameters);
     }
 };
 
@@ -113,6 +124,40 @@ const status = async (parameters) => {
             name: serviceName
         };
     }
+};
+
+const startBatch = async function (
+    numberOfCpus: number,
+    port: number = Preferences.get('trRouting.batchPortStart', 14000),
+    cacheDirectoryPath?: string
+) {
+    // Ensure we don't use more CPU than configured
+    // TODO The os.cpus().length should move to a "default config management class"
+    const maxThreadCount = config.maxParallelCalculators || os.cpus().length;
+    if (numberOfCpus > maxThreadCount) {
+        console.warn('Asking for too many trRouting threads (%d), reducing to %d', numberOfCpus, maxThreadCount);
+        numberOfCpus = maxThreadCount;
+    }
+
+    const params = { cacheDirectoryPath: cacheDirectoryPath };
+
+    await startTrRoutingProcess(port, false, numberOfCpus, params);
+
+    return {
+        status: 'started',
+        service: 'trRoutingBatch',
+        port: port
+    };
+};
+
+const stopBatch = async function (port = Preferences.get('trRouting.batchPortStart', 14000)) {
+    await stop({ port: port });
+
+    return {
+        status: 'stopped',
+        service: 'trRoutingBatch',
+        port: port
+    };
 };
 
 const startMultiple = async function (
@@ -183,6 +228,8 @@ export default {
     stop,
     restart,
     status,
+    startBatch,
+    stopBatch,
     startMultiple,
     stopMultiple,
     getAvailablePortsByStartingPort,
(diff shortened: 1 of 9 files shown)


packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -67,7 +75,8 @@ const start = async (parameters: { port?: number; debug?: boolean; cacheDirector
     const port = parameters.port || Preferences.get('trRouting.port');
 
     // TODO Check why we need this await, should not be useful before returning
-    return await startTrRoutingProcess(port, undefined, parameters);
+    // Set 2 threads as default, just in case we have more than one request being handled
```
tahini, line 78: This comment will be wrong if we change the default to 3... The comment should be where you actually set the DEFAULT_THREAD_COUNT above

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
```diff
@@ -30,10 +31,12 @@ startProcessMock.mockImplementation(async (serviceName: string,
     service: tagName,
     name: serviceName
 }));
-
 beforeEach(function () {
     startProcessMock.mockClear();
     stopProcessMock.mockClear();
+    // Override max parallel setting
+    // TODO might be a better way to do this
```
tahini, line 38: You could mock the config, which is sort of equivalent, though I guess mocking would be better since you are sure you are not touching the config for others. but I think jest runs each test separately so you shouldn't. But since you're asking...

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
```diff
@@ -56,7 +59,7 @@ describe('TrRouting Process Manager: start', () => {
             'trRouting4000',
             'trRouting',
             'trRouting',
-            ['--port=4000', `--osrmPort=${walkingOsrmMode.getHostPort().port}`, `--osrmHost=${walkingOsrmMode.getHostPort().host}`, '--debug=0', `--cachePath=${directoryManager.projectDirectory}/cache/test`],
+            ['--port=4000', `--osrmPort=${walkingOsrmMode.getHostPort().port}`, `--osrmHost=${walkingOsrmMode.getHostPort().host}`, '--debug=0', `--cachePath=${directoryManager.projectDirectory}/cache/test`, '--threads=2'],
```
tahini, line 62: So this is not configurable for the main trRouting process?

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
```diff
@@ -78,11 +81,72 @@ describe('TrRouting Process Manager: start', () => {
...
+            undefined,
+            false
+        );
+    });
+    test('start batch process with 4 cpus with 2 limit', async () => {
+        // Override the configuration
+        // TODO might have a better way to do this
+        config.maxParallelCalculators = 2;
+
+        const status = await TrRoutingProcessManager.startBatch(4);
```
tahini, line 134: I know you are really testing the number of threads, but might be worth adding a test with a different port and cache directory while you're here.

packages/transition-backend/src/services/simulation/SimulationRun.ts
```diff
@@ -58,20 +58,15 @@ export default class SimulationRunBackend extends SimulationRun {
     }
 
     async restartTrRoutingInstances() {
-        await TrRoutingProcessManager.stopMultiple(
-            this.attributes.options?.numberOfThreads || 1,
+        await TrRoutingProcessManager.stopBatch(
```
tahini, line 61: `getStartingPort`'s name should be changed too

packages/transition-backend/src/services/simulation/SimulationRun.ts
```diff
@@ -130,7 +124,6 @@ export default class SimulationRunBackend extends SimulationRun {
                     } catch (error) {
                         reject(error);
                     } finally {
-                        this.poolOfTrRoutingPorts.push(trRoutingPort);
```
tahini, line 133: You can remove the finally block, it's empty now

packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
```diff
@@ -183,9 +178,6 @@ export const batchAccessibilityMap = async (
                     console.error(`Error getting od trip result ${error}`);
                 }
             } finally {
-                if (trRoutingPort !== undefined) {
```
tahini, line 186: remove finally block

packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
```diff
@@ -205,7 +197,7 @@ export const batchAccessibilityMap = async (
         await promiseQueue.onIdle();
 
         progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 0.0 });
-        const stopStatus = await TrRoutingProcessManager.stopMultiple(trRoutingInstancesCount);
+        const stopStatus = await TrRoutingProcessManager.stopBatch();
```
tahini, line 200: Shouldn't you give the port number here? In case we have concurrency of jobs at some point and have multiple instances running.

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -196,7 +188,7 @@ class TrRoutingBatch {
             this.options.progressEmitter.emit('progress', { name: 'BatchRouting', progress: 1.0 });
             this.options.progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 0.0 });
 
-            const stopStatus = await TrRoutingProcessManager.stopMultiple(trRoutingInstancesCount);
+            const stopStatus = await TrRoutingProcessManager.stopBatch();
```
tahini, line 191: Pass port number

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -317,26 +309,26 @@ class TrRoutingBatch {
...
-
+            await TrRoutingProcessManager.stopBatch();
+            const startStatus = await TrRoutingProcessManager.startBatch(trRoutingInstancesCount);
+            trRoutingPort = startStatus.port;
             console.log('trRouting multiple startStatus', startStatus);
         } finally {
             this.options.progressEmitter.emit('progress', { name: 'StartingRoutingParallelServers', progress: 1.0 });
         }
-        return trRoutingInstancesCount;
+        return [trRoutingInstancesCount, trRoutingPort];
```
tahini, line 331: You can put the return in the try, to avoid adding the initializing the trRoutingPort above and giving the impression that -1 might be a possible value. There is no catch for a potential error anyway.

packages/transition-common/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -219,6 +219,9 @@ export class TransitAccessibilityMapCalculator {
                 accessibleNodes
             });
 
+            //accessibleMap expect the port number to be in an HostPort struct
+            //TODO This should be refactored, but this is a quick fix for #861
+            queryOptions.hostPort = {port: queryOptions.port};
```
tahini, line 224: Please run `yarn format` on the code, there should be spaces after the `{`

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
greenscientist: Yeah, forgot to move it after adding the constant

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
greenscientist: So I tried the mocking, but it did not work. Could not find an easy way to change the mock of the config in the test. If you have code example

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
greenscientist: No!   We can add an issue if we want to add a parameter for this. But I would wait until we refactor configuration management

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
greenscientist: We started with the default, so I stop with the default.

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
greenscientist: oh I see your comment about the no catch. Having just the finally was a bit confusing. Will need to have a comment about the no catch, just in case somebody add one someday.

greenscientist:
Addressing most comment. (ALso moved the batch accessibility port to the first commit of the stack)

tahini (approved):

PR #863, opened 2024-02-28
update lua profiles for walking to top at 5 km/h

kaligrafy:
add trunk with greater penalty for walking

### DIFF
--- a/packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/walking_no_penalty.lua
+++ b/packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/walking_no_penalty.lua
@@ -115,10 +115,11 @@ function setup()
 
     speeds = Sequence {
       highway = {
-        trunk           = walking_speed*0.98,
-        trunk_link      = walking_speed*0.98,
-        primary         = walking_speed*0.98,
-        primary_link    = walking_speed*0.98,
+        /* These speed are adjusted to penalize less pedestrian-friendly roads like trunk, primary and secondary */
+        trunk           = walking_speed*0.96,
+        trunk_link      = walking_speed*0.96,
+        primary         = walking_speed*0.97,
+        primary_link    = walking_speed*0.97,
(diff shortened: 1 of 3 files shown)


greenscientist (changes requested):
Please add comments explaining the chnange

kaligrafy:
> Please add comments explaining the chnange

Done!

greenscientist (changes requested):
I should have been more specific. Please add some comments in the lua file. It's really hard to understand what is happening when we compare them with upstream ones.

greenscientist:
Does not need to be a comment on each line, but at least a comment on the top to explain why it's different than upstream.

packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/walking_no_penalty.lua
```diff
@@ -115,28 +115,28 @@ function setup()
 
     speeds = Sequence {
       highway = {
-        trunk           = walking_speed*0.98,
-        trunk_link      = walking_speed*0.98,
-        primary         = walking_speed*0.98,
-        primary_link    = walking_speed*0.98,
+        trunk           = walking_speed*0.96,
```
tahini, line 118: Would it be possible to generalize the walking profiles, you have 3 files to change here, they may end up not being identical! You can open an issue to track this.

tahini (approved):

kaligrafy:
> Does not need to be a comment on each line, but at least a comment on the top to explain why it's different than upstream.

You mean add comments in the code?

packages/chaire-lib-backend/src/utils/processManagers/osrmProfiles/walking_no_penalty.lua
kaligrafy: https://github.com/chairemobilite/transition/issues/864

tahini (approved):

PR #877, opened 2024-03-13
users: add a last login time to users

### LINKED ISSUE #876: Implement lastLogin in user db

tahini:
fixes #876

* Add a field: `last_login_at` to the users table.
* Add the `recordLogin` method to the UserModel to let the model implement recording login. The default implementation does nothing
* Set the last login time to the user model based on the `users` table
* Call the recordLogin method in passport strategies whenever there are successful logins.


kaligrafy (approved):

greenscientist:
I'm curious, do we want to count API login separatly or not?

tahini:
> I'm curious, do we want to count API login separatly or not?

1- To get the API token, one needs to login with username/password, so this login will be recorded
2- API is just one request, so it won't be considered a log in. But we could record the last token usage date

I'll wait for the API PR to be in before merging this one. There will be conflicts...

PR #878, opened 2024-03-15
alow higher running speeds for transferable mode

### LINKED ISSUE #875: Remove speed limitation on transfer paths
The default cruising speed is limited to 5 km/h for transfer paths, which is problematic for two reasons:

1. They are tagged as "walking/cycling", but 5 km/h is way too slow for cycling
2. Transfer paths are not necessarily physical paths, but can also be used to represent conceptual itineraries between two nodes and, as such, could be used to speed up transfers between nodes that are located too far from each other. 

A warning could be displayed when excessive speeds are input, but without actually blocking their use.

kaligrafy:
closes #875

### DIFF
--- a/packages/chaire-lib-common/src/config/lineModesDefaultValues.ts
+++ b/packages/chaire-lib-common/src/config/lineModesDefaultValues.ts
@@ -172,6 +172,6 @@ export default {
         defaultDeceleration: 1.5,
         defaultDwellTimeSeconds: 0,
         customLayoverMinutes: 0,
-        maxRunningSpeedKmH: 5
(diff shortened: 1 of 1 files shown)


greenscientist (commented):
Shoudn't we change the name of this functionnality/variable if it's can be confusing ?

kaligrafy:
We should use a virtual osm editor for this instead, because it is a kind of a hack (virtual path/route), but it works for now, so I would wait before changing anything.

tahini (approved):

kaligrafy:
@tahini 3 days waiting for docker status... We force merge?

tahini:
> @tahini 3 days waiting for docker status... We force merge?

No! we rebase, the docker status is fixed!

PR #885, opened 2024-03-21
update bus lua profiles to accept highway=busway

kaligrafy:


greenscientist:
Is this something that is present upstream?

kaligrafy:
No, OSRM default does not include any bus specific routing profile. But in Transition, we must allow buses to route on busways

greenscientist (approved):
ok, indeed, OSRM does not have a default bus profile.

PR #904, opened 2024-04-12
Add bearer strategy with token

tahini:
This cleanly adds the authentication method with token. It adds a user in the auth model with the token as username (this avoids adding a new field to the user table). Relogging with the same token in the URL will cause the user to log back in with the same user data.

### DIFF
--- a/packages/chaire-lib-backend/src/config/auth/__tests__/directToken.config.test.ts
+++ b/packages/chaire-lib-backend/src/config/auth/__tests__/directToken.config.test.ts
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import passport from 'passport';
+import each from 'jest-each';
+
+import directTokenLogin from '../directToken.config';
+import usersDbQueries from '../../../models/db/users.db.queries';
+import { userAuthModel } from '../../../services/auth/userAuthModel';
+import config from '../../server.config';
+
+// req.logIn needs to be set and is called by passport when successful
+const logInFct = jest.fn().mockImplementation((_a, _b, callback) => {
+    callback();
+});
+
+jest.mock('../../../models/db/users.db.queries', () => ({
+    find: jest.fn().mockResolvedValue(undefined),
+    create: jest.fn(),
+    setLastLogin: jest.fn()
+}));
+const mockFind = usersDbQueries.find as jest.MockedFunction<typeof usersDbQueries.find>;
+const mockCreate = usersDbQueries.create as jest.MockedFunction<typeof usersDbQueries.create>;
+const mockSetLastLogin = usersDbQueries.setLastLogin as jest.MockedFunction<typeof usersDbQueries.setLastLogin>;
+
+const newUserId = 7;
+
+beforeEach(() => {
+    logInFct.mockClear();
+    mockFind.mockClear();
+    mockCreate.mockClear();
+    mockCreate.mockImplementation(async (attribs) => {
+        return {
+            id: newUserId,
+            uuid: 'arbitrary',
+            ...attribs
+        }
+    });
+    mockSetLastLogin.mockClear();
+});
+
+test('regex', () => {
+    expect('abcdef'.match(/[a-z]+/gi)).not.toBeNull();
+})
+
+each([
+    ['undefined format', '1234-1234', undefined, true],
+    ['undefined format, no token', '', undefined, false],
+    ['simple regex, matching token', 'abdefg', /^[a-z]+$/gi, true],
+    ['simple regex, partly matching token at start, invalid', 'abcdefg87', /^[a-z]+$/gi, false],
+    ['simple regex, partly matching token at end, invalid', '87abcdefg', /^[a-z]+$/gi, false],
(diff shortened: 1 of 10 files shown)


kaligrafy:
I will let @greenscientist review this one

greenscientist (changes requested):
We need to document a bit more the configuraiton format for the "regex"
If your example, you use: 
            tokenFormat: /^[\d]{10}$/gi
But the regex part is really just "^[\d]{10}$", the rest is specific to the regex engine used and could have quite different meanings. 
(I'm not fully sure we want people to be able to use the g to be able to match multiple lines of content.

packages/chaire-lib-backend/src/config/auth/directToken.config.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const isTokenValid = (token) => {
+    const tokenFormat = config.auth?.directToken?.tokenFormat;
+    return tokenFormat === undefined ? true : token.match(tokenFormat);
+};
+
+export default <U extends IUserModel>(passport: PassportStatic, authModel: IAuthModel<U>) => {
+    passport.use(
+        'direct-token',
+        new BearerStrategy.Strategy(async (token, done) => {
```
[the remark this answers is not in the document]
tahini: done

packages/chaire-lib-backend/src/config/auth/directToken.config.ts
```diff
@@ -0,0 +1,47 @@
...
+import { PassportStatic } from 'passport';
+import BearerStrategy from 'passport-http-bearer';
+import { IAuthModel, IUserModel } from '../../services/auth/authModel';
+import config from '../server.config';
+
+// Direct token login strategy: the token will be used as the username. If it
+// exists in the database, the user will be logged in, otherwise, it will be
+// automatically created.
+
+const isTokenValid = (token) => {
```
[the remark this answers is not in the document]
tahini: done

tahini:
> We need to document a bit more the configuraiton format for the "regex" If your example, you use: tokenFormat: /^[\d]{10}$/gi But the regex part is really just "^[\d]{10}$", the rest is specific to the regex engine used and could have quite different meanings. (I'm not fully sure we want people to be able to use the g to be able to match multiple lines of content.

I just realized the configs are documented in chaire-lib-common, so it's been added there.

packages/chaire-lib-backend/src/config/auth/directToken.config.ts
```diff
@@ -0,0 +1,52 @@
...
+import { PassportStatic } from 'passport';
+import BearerStrategy from 'passport-http-bearer';
+import { IAuthModel, IUserModel } from '../../services/auth/authModel';
+import config from '../server.config';
+
+// Direct token login strategy: the token will be used as the username. If it
+// exists in the database, the user will be logged in, otherwise, it will be
+// automatically created.
+
+const isTokenValid = (token: string) => {
```
greenscientist, line 16: I would expect the function to return a boolean ???

packages/chaire-lib-backend/src/config/auth/directToken.config.ts
```diff
@@ -0,0 +1,52 @@
...
+    const tokenFormat = config.auth?.directToken?.tokenFormat;
+    return tokenFormat === undefined ? true : token.match(tokenFormat);
```
greenscientist, line 18: token match return an array or null, we should probably check explicitely for those, instead of returning the matching array. 

Also should we check that the regex is valid ??

packages/chaire-lib-backend/src/config/auth/__tests__/directToken.config.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+    mockSetLastLogin.mockClear();
+});
+
+test('regex', () => {
+    expect('abcdef'.match(/[a-z]+/gi)).not.toBeNull();
+})
+
+each([
+    ['undefined format', '1234-1234', undefined, true],
```
greenscientist, line 50: Probably worth it to have a test with an invalid regex expression

packages/chaire-lib-backend/src/config/auth/directToken.config.ts
greenscientist: (I would love to have a log message saying that the configured regex is invalid. It could be elsewhere, in the app startup maybe. Or I guess eventually in a config validator, but would be good to have at least a TODO)

greenscientist (changes requested):
Requesting change to have a return type specificied. The rest is only advisory.

packages/chaire-lib-frontend/src/components/forms/auth/anonymous/AnonymousLogin.tsx
```diff
@@ -1,9 +1,14 @@
+/*
+ * Copyright 2022, Polytechnique Montreal and contributors
```
greenscientist, line 2: Is 2022 the right date?

greenscientist (approved):
There is just one question...

packages/chaire-lib-frontend/src/components/forms/auth/anonymous/AnonymousLogin.tsx
tahini: yes, the header was missing, but the file is from 2022.

PR #920, opened 2024-04-29
frontend: Extract the `verifyAuthentication` method

tahini:
This allows other code path to re-use the app loading code to verify authentication. One example is when the user is logged out, the requests will typically return a 401 response to the application. The code handling this response code can request the server to verifyAuthentication and thus differentiate between a logged out user and one trying to access a forbidden resource.


tahini:
This will allow to fix evolution's issue https://github.com/chairemobilite/evolution/issues/53

kaligrafy (approved):

PR #925, opened 2024-05-01
pyTransition: Added pyTransition source code

tahini:
pyTransition is a Python package designed to interact with the public API of the transit planning application Transition. It allows users to retrieve and request geographic and routing data from the app.


greenscientist (approved):
There would be minor things to change, (like the author field), but we can do that later

Issue #926, opened 2024-05-01
scorecard: builds not working


tahini:
@greenscientist  see https://github.com/chairemobilite/transition/actions/runs/8913696591/job/24479716372

tahini:
Fixed with #927

PR #927, opened 2024-05-01
Use full hash in Openssf scorecard-action

greenscientist:


greenscientist:
Test by adding (and then removing) this action on PR)

greenscientist:
@tahini sorry about that
Closes #926

tahini (approved):

PR #928, opened 2024-05-02
userEmail: Add try/catch around email sending functions

tahini:
Sending email is often done asynchronously by the server and there is nothing to catch the error in the sendEmail promise. All email sending functions now catch and log the exception before throwing it back.


greenscientist (approved):
Wonder if we want more logging into the sendEmail function directly (if possible, it might be an external lib)

tahini:
> Wonder if we want more logging into the sendEmail function directly (if possible, it might be an external lib)

more logging as in "log more stuff"? like mail to, mail from, etc? Callers already log the fact that an email was sent, now it should log errors as well. We'll see if we get the information we might need for the unsent emails. Otherwise, we can look into more logging.

Issue #898, opened 2024-04-05
User should not have to wait that cache is recreated before logging in at instance startup


greenscientist:
The following code is in the startup path: 
```
    // Batch calculation jobs require the cache to exist. It should not be
    // possible to run those without the cache, so we wait for its creation
    // before allowing users to connect
    if (_booleish(process.env.STARTUP_RECREATE_CACHE)) {
        console.log('Recreating cache files');
        // TODO get cachePathDIrectory from params
        await recreateCache({ refreshTransferrableNodes: true, saveLines: true });
    }

    // Enqueue/resume running and pending tasks
    // FIXME This implies a single server process for a given database. We don't
    // know if the job is enqueued in another process somewhere and may be
    // executed twice. For pending jobs, we could have a first run, first serve,
    // but for jobs in progress, we don't know if they are actively being run or
    // not
    if (process.env.STARTUP_RESTART_JOBS === undefined || _booleish(process.env.STARTUP_RESTART_JOBS)) {
        await ExecutableJob.enqueueRunningAndPendingJobs();
    }
```
It should be done in the background, while still letting the user come in. We should not start any calculation before this is done, but users should be able to use the interface.

tahini:
This was done to make sure that the cache was always present when the application is available, so we don't have to manage if it's there or not, because the next block makes enqueues previously running tasks and when they run, the cache has better be there.

I don't like the idea of needing an extra mechanism to  make sure calculations are ready and having to wait for it. But if it can be done cleanly and be transparent to calculation tasks, then I guess I can be convinced.

greenscientist:
The biggest issue in the code path is recalculating all the transferable nodes. Maybe that could be skipped now, isn't everything in the DB ?

tahini:
@greenscientist  is this fixed by any chance? Or is there anything remaining?

PR #930, opened 2024-05-02
Don't recalculate transferable nodes on startup

greenscientist:
We now save the transferable nodes in the database, so we don't really need to recalculate them at each startup. But we still need to save them to the capnp cache file.

We add a saveAllNodesToCache function to NodeCollectionUtils, based on saveAndUpdateAllNodes, that only store the node data without recalculating it. We add a else in the recreateCache function where we test for the refreshTransferrableNodes flag to call our new function in that case. (That way we always save the transferable nodes to cache)

The dbToCache were missing a mockClear, we switch it to jest.clearAllMocks() and fixes the erroneous tests

The call in socketServerApp was changed to ask for not refreshing the transferable node when recreating the cache on startup

### DIFF
--- a/packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
+++ b/packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
@@ -79,3 +79,41 @@ export const saveAndUpdateAllNodes = async (
     await Promise.allSettled(addPromises);
     progressEmitter?.emit('progress', { name: 'UpdatingTransferableNodes', progress: 1.0 });
 };
+
+/**
+ * Save the nodes to cache including the transferable node
+ * We do the saving in batch, to not overload json2capnp
+ *
+ * @param nodeCollection The collection containing the nodes to update and save
+ * @param collectionManager The collection manager, used to create the nodes
+ */
+export const saveAllNodesToCache = async (
+    nodeCollection: NodeCollection,
+    collectionManager?,
+    cachePathDirectory?: string
+): Promise<void> => {
+    const features = nodeCollection.getFeatures();
+    // The number of current save was chosen arbitrarily. If we send them
+    // all "at once", json2capnp will create a thread of each and ultimately
+    // ran out of memory. This number seem a good tradeoff between speed of execution
+    // and presentation of resources
+    const promiseQueue = new PQueue({ concurrency: 250 });
+
+    const promiseProducer = async (nodeGeojson, index) => {
+        const node = new Node(nodeGeojson.properties, false, collectionManager);
+        // TODO This is ugly, we should not have to manually fetch the transferable nodes
+        // This need to be refactored, part of a Node refactor
+        // Copied from TransitNode.saveNode
+        const transferableNodes = await transferableNodesDbQueries.getFromNode(node.id);
+        node.setData('transferableNodes', transferableNodes);
+        await objectToCache(node, cachePathDirectory);
+    };
+
+    const addPromises = features.map(async (feature, index) =>
+        promiseQueue.add(async () => promiseProducer(feature, index))
+    );
+
+    // Run all the promises, no matter their result.
+    // TODO: What about failures? Should we track them, reject upon first failure (using Promise.all), just console.error them?
+    await Promise.allSettled(addPromises);
+};
--- a/packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
+++ b/packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
@@ -9,10 +9,13 @@ import events from 'events';
 import { v4 as uuidV4 } from 'uuid';
 import { NodeAttributes } from 'transition-common/lib/services/nodes/Node';
 import NodeCollection from 'transition-common/lib/services/nodes/NodeCollection';
-import { saveAndUpdateAllNodes } from '../NodeCollectionUtils';
+import { saveAndUpdateAllNodes, saveAllNodesToCache } from '../NodeCollectionUtils';
 import { objectToCache } from '../../../models/capnpCache/transitNodes.cache.queries';
 import { getTransferableNodes } from '../TransferableNodeUtils';
+import transferableNodesDbQueries from '../../../models/db/transitNodeTransferable.db.queries';
 
+// Mock all DB queries about transferable nodes
+jest.mock('../../../models/db/transitNodeTransferable.db.queries');
 jest.mock('../TransferableNodeUtils', () => ({
     getTransferableNodes: jest.fn()
 }));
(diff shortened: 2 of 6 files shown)


greenscientist:
This save 5 minutes on startup on my machine. (on a total that was about 7 minutes)

tahini:
Please run `yarn format` to get the linter to pass

packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
```diff
@@ -79,3 +79,45 @@ export const saveAndUpdateAllNodes = async (
...
+ * @param collectionManager The collection manager, used to create the nodes
+ */
+export const saveAllNodesToCache = async (
+    nodeCollection: NodeCollection,
+    placeCollection: PlaceCollection | undefined,
+    collectionManager?,
+    cachePathDirectory?: string
+): Promise<void> => {
+    // Make sure the spatial index is up to date
+    // TODO is this necessary in this case ?
```
tahini, line 97: No it's not

packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
```diff
@@ -79,3 +79,45 @@ export const saveAndUpdateAllNodes = async (
...
+/**
+ * Save the nodes to cache including the transferable node
+ * We do the saving in batch, to not overload json2capnp
+ *
+ * @param nodeCollection The collection containing the nodes to update and save
+ * @param collectionManager The collection manager, used to create the nodes
+ */
+export const saveAllNodesToCache = async (
+    nodeCollection: NodeCollection,
+    placeCollection: PlaceCollection | undefined,
```
tahini, line 92: I don't think the collection is necessary in this case either

packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
```diff
@@ -79,3 +79,45 @@ export const saveAndUpdateAllNodes = async (
...
+
+        await objectToCache(node, cachePathDirectory);
+    };
+
+    const addPromises = features.map(async (feature, index) =>
+        promiseQueue.add(async () => promiseProducer(feature, index))
+    );
+
+    // Run all the promises, no matter their result.
+    // TODO: What about failures? Should we track them, reject upon first failure (using Promise.all), just console.error them?
```
tahini, line 121: console.error them is a good start. I suppose this is handled by the objectToCache function already?

tahini (changes requested):
There are some simplifications to the function that can be done

packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
greenscientist: No idea, just a copy/paste from the other function

packages/transition-backend/src/services/nodes/NodeCollectionUtils.ts
greenscientist: You mean the placeCollection?

tahini: yes, in fact, I am sure it is not necessary

tahini:
Also, isn't there an issue related to this PR? If so, please add to the commit message.

greenscientist:
> Also, isn't there an issue related to this PR? If so, please add to the commit message.

It did not add it because it did not want for github to close the issue. It's not fully solving it (#898)

packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
```diff
@@ -113,7 +113,7 @@ test('saveAndUpdateAllNodes without collection manager', async() => {
         walkingDistancesMeters: [0]
     };
     mockGetTransferableNodes.mockImplementation(async (node, _) => node.getId() === nodeAttributesClose1.id ? node1TransferableNodes : node.getId() === nodeAttributesClose2.id ? node2TransferableNodes : node.getId() === nodeAttributesClose3.id ? node3TransferableNodes : nodeFarTransferableNodes);
-    await saveAndUpdateAllNodes(nodeCollection, undefined, eventEmitter);
+    await saveAndUpdateAllNodes(nodeCollection, eventEmitter);
```
tahini, line 116: ne pas modifier cette ligne

greenscientist:
Added the ugly quick fix

packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
tahini: Il reste à remettre le undefined ici

tahini (changes requested):
Le test fail, mais c'est. Ça fonctionne avec trRouting?

greenscientist:
> Le test fail, mais c'est. Ça fonctionne avec trRouting?

I know...  I don't know why the test for the previous function have functionned at all :(

Mais oui, ca fonctionne avec trRouting.

tahini (approved):

packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
```diff
@@ -9,10 +9,13 @@ import events from 'events';
...
 import { NodeAttributes } from 'transition-common/lib/services/nodes/Node';
 import NodeCollection from 'transition-common/lib/services/nodes/NodeCollection';
-import { saveAndUpdateAllNodes } from '../NodeCollectionUtils';
+import { saveAndUpdateAllNodes, saveAllNodesToCache } from '../NodeCollectionUtils';
 import { objectToCache } from '../../../models/capnpCache/transitNodes.cache.queries';
 import { getTransferableNodes } from '../TransferableNodeUtils';
+import transferableNodesDbQueries from '../../../models/db/transitNodeTransferable.db.queries';
 
+// Mock all DB queries about transferable nodes
+jest.mock('../../../models/db/transitNodeTransferable.db.queries');
```
tahini, line 18: You might want to make some check on these functions, with some returning some transferrable nodes, and you can then make sure the transferable nodes are also sent to the save method.

packages/transition-backend/src/services/nodes/__tests__/NodeCollectionUtils.test.ts
```diff
@@ -132,3 +135,21 @@ test('saveAndUpdateAllNodes without collection manager', async() => {
...
+    // Make sure nodes have been updated
+    const savedObject1 = mockedObjectToCache.mock.calls[0][0];
+    const savedObject2 = mockedObjectToCache.mock.calls[1][0];
+    const savedObject3 = mockedObjectToCache.mock.calls[2][0];
+    const savedObject4 = mockedObjectToCache.mock.calls[3][0];
+    // Validate node data was successfully saved
+    expect(savedObject1.getId()).toEqual(nodeAttributesClose1.id);
+    expect(savedObject2.getId()).toEqual(nodeAttributesClose2.id);
+    expect(savedObject3.getId()).toEqual(nodeAttributesClose3.id);
+    expect(savedObject4.getId()).toEqual(nodeAttributesFar.id);
```
tahini, line 154: You can also check here that data.transferrableNode is correctly set with the mock mentioned above.

tahini (approved):
I still approve, but you might want to add some more test statements to your test. For when we refactor all this, to make sure all is still right.

greenscientist:
Yeah, no I really don't want to add more test for that. I don't see any way to do a clean test for that. It should all go away anyway with a good refactor. 
This data structure really upset me with the way it is at the moment!

tahini:
What upsets you so much? Transferable nodes are a concept outside the node itself, they should not be inside the node. Is it because it is an optional property of the node's data field that bugs you most?

greenscientist:
> What upsets you so much? Transferable nodes are a concept outside the node itself, they should not be inside the node. Is it because it is an optional property of the node's data field that bugs you most?

You say it's separate, by everywhere in our applications, it's contained inside the Node (in TrRouting, Transition, in the capnp cache). If it's a separate thing, it should be it's own object, not part of a Node one.

PR #936, opened 2024-05-06
db: Allow all queries to receive a transaction parameter

tahini:
Also make sure individual operations that span many tables (like scenarios and schedules) are all inside transactions to avoid errors while saving those object to leave the object is an intermediate, unknown, state.

While doing so, to make transactions as fast and short as possible, the queries that involved reading/processing/writing now try to do it in a single database query, or favor batch inserts/updates/deletes.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
+++ b/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
@@ -12,7 +12,18 @@ import TrError from 'chaire-lib-common/lib/utils/TrError';
 
 // TODO Move these to a generic class, so we can use identical types for U and T in generic methods.
 
-const stringifyDataSourceIds = function (dataSourceIds: string[]): string[] {
+/**
+ * Helper function to wrap the data source IDs in '' to be used in database
+ * queries.
+ *
+ * FIXME Mostly called in the context of raw queries with a 'IN' where
+ * statement. See if we can use a regular `whereIn` instead and remove the need
+ * for this function.
+ *
+ * @param dataSourceIds An array of data source IDs to stringify
+ * @returns A DB-ready array of data source id strings
+ */
+export const stringifyDataSourceIds = function (dataSourceIds: string[]): string[] {
     if (dataSourceIds && !Array.isArray(dataSourceIds)) {
         dataSourceIds = [dataSourceIds];
     }
@@ -23,7 +34,15 @@ const stringifyDataSourceIds = function (dataSourceIds: string[]): string[] {
         : [];
 };
 
-const exists = async (knex: Knex, tableName: string, id: string): Promise<boolean> => {
+/**
+ * Check whether a record with the given id exists in the table
+ *
+ * @param knex The database configuration object
+ * @param tableName The name of the table on which to execute the operation
+ * @param id The ID of the object
+ * @returns Whether an object with the given ID exists in the table
+ */
+export const exists = async (knex: Knex, tableName: string, id: string): Promise<boolean> => {
     if (!uuidValidate(id)) {
         throw new TrError(
             `Cannot verify if the object exists in ${tableName} because the required parameter id is missing, blank or not a valid uuid`,
@@ -53,17 +72,38 @@ const exists = async (knex: Knex, tableName: string, id: string): Promise<boolea
     }
 };
 
-const create = async <T extends GenericAttributes, U>(
+/**
+ * Insert a new record in a table
+ *
+ * @param knex The database configuration object
+ * @param tableName The name of the table on which to execute the operation
+ * @param parser A parser function which converts an object's attributes to db
+ * fields
+ * @param newObject The new object's attributes
+ * @param options Additional options parameter. `returning` specifies which
+ * field's or fields' values to return after insert. `transaction` is an
+ * optional transaction of which this insert is part of.
+ * @returns The requested `returning` field values.
+ */
+export const create = async <T extends GenericAttributes, U>(
     knex: Knex,
     tableName: string,
     parser: ((arg: T) => U) | undefined,
     newObject: T,
-    returning: string | string[] = 'id'
+    options: {
+        returning?: string | string[];
+        transaction?: Knex.Transaction;
+    } = { returning: 'id' }
 ): Promise<string | { [key: string]: unknown }> => {
     try {
+        const returning = options.returning || 'id';
         const _newObject = parser ? parser(newObject) : newObject;
 
-        const returningArray = await knex(tableName).insert(_newObject).returning(returning);
+        const query = knex(tableName).insert(_newObject).returning(returning);
+        if (options.transaction) {
+            query.transacting(options.transaction);
+        }
+        const returningArray = await query;
         return typeof returning === 'string' ? returningArray[0][returning] : returningArray[0];
     } catch (error) {
         throw new TrError(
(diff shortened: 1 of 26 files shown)


kaligrafy (approved):

greenscientist:
@kaligrafy, did you really review this ? Seem quick to take 4 minutes to review 24 files that affect how we interract with the database...

tahini:
Tested with sequential tests and transition (gtfs import and calculations)

kaligrafy:
I did review, but i did not test

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -58,12 +58,20 @@ const create = async <T extends GenericAttributes, U>(
     tableName: string,
     parser: ((arg: T) => U) | undefined,
     newObject: T,
-    returning: string | string[] = 'id'
+    options: {
+        returning?: string | string[];
+        transaction?: Knex.Transaction;
```
greenscientist, line 63: Would it be a good opportunity to add a function header that document the parameters??

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -177,14 +202,18 @@ const updateMultiple = async <T extends GenericAttributes, U>(
...
+    } = {}
 ): Promise<string[][]> => {
     const _attributeCollection: any[] = parser
         ? attributeCollection.map((attribute) => parser(attribute))
         : attributeCollection;
 
     try {
-        return (await knex.transaction(async (trx) => {
+        const returning = options.returning || 'id';
+        const updateMultipleTransaction = async (trx: Knex.Transaction) => {
```
greenscientist, line 216: Maybe a comment about this inner function...

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -194,8 +223,11 @@ const updateMultiple = async <T extends GenericAttributes, U>(
                     .transacting(trx);
             });
             const returningArray = await Promise.all(queries);
-            return returningArray;
-        })) as string[][];
+            return returningArray as string[][];
+        };
+        return await (options.transaction !== undefined
```
greenscientist, line 228: Not required, but it could  be interesting to so if this function could be refactored to avoid an inner
function and an inline if like that. 
At least, please add some comments.

packages/transition-backend/src/models/db/__tests__/transitAgencies.db.test.ts
```diff
@@ -5,6 +5,7 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 import { v4 as uuidV4 } from 'uuid';
+import knex from 'chaire-lib-backend/lib/config/shared/db.config';
```
greenscientist, line 8: There's no chances that we messup a prod DB which this ?

packages/transition-backend/src/models/db/migrations/20240503162600_addUniqueScenarioServiceIdx.ts
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { Knex } from 'knex';
```
greenscientist, line 7: We should really have a standard header comment with a description for each migration! (This is just a general comment)

packages/transition-backend/src/models/db/transitScenarios.db.queries.ts
```diff
@@ -112,192 +114,256 @@ const read = async (id: string) => {
     }
 };
 
-const create = async (newObject: ScenarioAttributes, returning?: string) => {
+const _create = async (
```
greenscientist, line 117: Please add a comment to explain why we have a private create

greenscientist: (It probably need a more elaborate name than juste _create

packages/transition-backend/src/models/db/transitScenarios.db.queries.ts
```diff
@@ -112,192 +114,256 @@ const read = async (id: string) => {
...
-    // Create the main object in main table
-    const updated = await defaultUpdate(knex, tableName, attributesCleaner, id, updatedObject, { returning });
-    try {
-        // Update services if they are not undefined (partial update)
-        const services = updatedObject.services;
-        if (services !== undefined) {
-            // Prepare current services for scenario
-            const currentServices = await knex(serviceTableName).select().where('scenario_id', id);
-            const { newServices, deletedServices } = getNewAndDeletedServicesForScenario(id, services, currentServices);
+const _update = async (
```
greenscientist, line 212: same about this name

packages/transition-backend/src/models/db/transitSchedules.db.queries.ts
```diff
@@ -101,128 +106,291 @@ const scheduleTripsAttributesParser = function (attributes: any): SchedulePeriod
...
+            });
+            return _createTrips(period.trips, options);
+        });
+    await Promise.all(tripPromises);
+    return ids;
 };
 
-// create all needed inserts from scheduleData (from scheduleByServiceId[serviceId])
-const createFromScheduleData = async function (scheduleData: ScheduleAttributes) {
+const _createFromScheduleData = async function (
```
greenscientist, line 152: As usual, we have a private and public function with the same name. Please explain

packages/transition-backend/src/models/db/transitSchedules.db.queries.ts
```diff
@@ -101,128 +106,291 @@ const scheduleTripsAttributesParser = function (attributes: any): SchedulePeriod
...
+    });
+
+    const periodsWithTrips = schedulePeriods.filter((period) => period.trips !== undefined);
+    if (periodsWithTrips.length > 0) {
+        await _updateTripsForPeriods(periodsWithTrips, scheduleId, options);
+    }
+    return ids;
+};
+
+const _updateFromScheduleData = async function (
```
greenscientist, line 279: same

greenscientist (changes requested):
My request change is for adding comment and for the "private" function name that duplicate the public one. (either comment or use a more explicit name, ideally both)

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -92,7 +104,11 @@ const createMultiple = async <T extends GenericAttributes, U>(
 
         const chunkSize = 250;
 
-        return await knex.batchInsert(tableName, _newObjects as any, chunkSize).returning(_returning);
+        const query = knex.batchInsert(tableName, _newObjects as any, chunkSize).returning(returning);
+        if (options.transaction) {
+            query.transacting(options.transaction);
+        }
+        return await query;
```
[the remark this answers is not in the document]
tahini: If you await and something bad happens, the try catch in this function will catch it. If you return without the await, then it's the caller's try/catch that will be responsible to catch the error, wherever that is. Since we want to sort of control the exceptions that are thrown by database queries (ie have our own exceptions thrown and not knex's), then the queries need to await.

That said, with the current rewrite, the try/catch are in the public user-facing functions, so those need to await, but maybe not the inner private functions, where there is no try/catch.

packages/transition-backend/src/models/db/transitScenarios.db.queries.ts
tahini: Unless the private functions are reused by multiple public ones or are really huge (like in the schedules), I'll nest them in their public counterpart to limit their scope. Like what is done in the default.db.queries `updateMultiple` function.

tahini:
@greenscientist  I added tons of comments and used nested function whenever it made more sense scope-wise than private functions.

greenscientist (approved):
I still don't like inner functions, but that's just my general hatred of javascript. At least they have good names now. 

THanks for the comments.

tahini:
Nested functions, when used judiciously, is a quite powerful feature of javascript! But apparently people coming from C/C++ don't like them... :p

All this and I'm not even closing an issue!

Issue #297, opened 2021-09-15
transition: move agency/line/etc duplication operations to the backend


tahini:
The *Duplicator code is executed in the client, which sends a lot of queries to the server. It could be done directly by the server instead of individually sending all the data there, one object at a time.

The steps for this issue are

- [x] #939
- [ ] #940
- [ ] #941
- [ ] #942
- [x] #943
- [ ] #944
- [ ] #945

The first 5 are the critical ones as duplicating an agency requires all of them and that operation should be "atomic".

greenscientist:
Just bumped into this again, it would be a great optimisation, it's quite long and error prone at the moment

PR #946, opened 2024-05-13
Bring the service duplication operation server side 

tahini:
First and easiest step towards #297, but it prepares the approach:

* Add argument to database `collection` function to fetch partial collection of services
* Add database queries to fetch all services names starting with a certain prefix, to generate unique names
* Add operations to duplicate single or multiple services, with separate function for saving, which allow longer running duplicate operations (like agency or line duplication), to postpone saving when all objects are ready to reduce the transaction time.
* Add socket route to duplicate services
* Let the service duplication button use this new socket route to duplicate the service. Other callers, like line duplicator are not ready yet to call the new socket route. They will themselves be moved to backend later.

### DIFF
--- a/packages/transition-backend/src/services/gtfsImport/ServiceImporter.ts
+++ b/packages/transition-backend/src/services/gtfsImport/ServiceImporter.ts
@@ -16,7 +16,7 @@ import ServiceCollection from 'transition-common/lib/services/service/ServiceCol
 import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import { ServiceImportData, GtfsImportData } from 'transition-common/lib/services/gtfs/GtfsImportTypes';
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
-import { getUniqueServiceName } from 'transition-common/lib/services/service/ServiceUtils';
+import { getUniqueServiceName } from '../transitObjects/transitServices/ServiceUtils';
 import { GtfsObjectImporter } from './GtfsObjectImporter';
 import { formatColor, GtfsInternalData } from './GtfsImportTypes';
 import LineCollection from 'transition-common/lib/services/line/LineCollection';
@@ -346,7 +346,7 @@ export class ServiceImporter implements GtfsObjectImporter<ServiceImportData, Se
         if (existingService) {
             return existingService;
         }
-        serviceAttributes.name = getUniqueServiceName(this._existingServices, serviceAttributes.name || '');
+        serviceAttributes.name = await getUniqueServiceName(serviceAttributes.name || '');
         // Create a new service
         const newService = new Service(serviceAttributes, true);
         // TODO Save only at the end of the whole import, with a batch save
--- a/packages/transition-backend/src/services/transitObjects/transitServices/ServiceDuplicator.ts
+++ b/packages/transition-backend/src/services/transitObjects/transitServices/ServiceDuplicator.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+// This file contains functions related to the duplication and copy of the
+// services. Except for the high level database transactions, this file should
+// not contain any functions calling directly database queries. These should be
+// in the ServiceUtils.ts file.
+import { Knex } from 'knex';
+import knex from 'chaire-lib-backend/lib/config/shared/db.config';
+import Service from 'transition-common/lib/services/service/Service';
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
+import { getServicesById, getUniqueServiceName, saveServices } from './ServiceUtils';
+import * as Status from 'chaire-lib-common/lib/utils/Status';
+import { WithTransaction } from 'chaire-lib-backend/lib/models/db/types.db';
+
+// TODO Add more options, like the complete name of the new service
+export type DuplicateServiceOptions = {
+    newServiceSuffix?: string;
+};
+
+/**
+ * Duplicate services and save them in the database.
+ *
+ * @param serviceIds The IDs of the services to duplicate
+ * @param options Duplication options
+ * @returns A status object a mapping of the previous service IDs to the new
+ * ones
+ */
+export const duplicateServices = async (
+    serviceIds: string[],
+    { transaction, ...options }: DuplicateServiceOptions & WithTransaction
+): Promise<Status.Status<{ [previousId: string]: string }>> => {
+    try {
+        // Nested function to require a transaction around the duplication
+        const duplicateWithTransaction = async (trx: Knex.Transaction) => {
+            const duplicatedServices: Service[] = [];
+            const oldNewIdMapping = {} as { [previousId: string]: string };
+            const services = await getServicesById(serviceIds, { transaction: trx });
+
+            for (const service of services) {
+                const duplicatedService = await duplicateService(service, { ...options, transaction: trx });
+                duplicatedServices.push(duplicatedService);
+                oldNewIdMapping[service.getId()] = duplicatedService.getId();
+            }
+
+            await saveServices(duplicatedServices, { transaction: trx });
+            return Status.createOk(oldNewIdMapping);
+        };
+        // Make sure the update is done in a transaction, use the one in the options if available
+        return transaction
+            ? await duplicateWithTransaction(transaction)
+            : await knex.transaction(duplicateWithTransaction);
+    } catch (error) {
+        console.log('An error occurred while duplicating services: ', error);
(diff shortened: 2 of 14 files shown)


kaligrafy (approved):

greenscientist:
> * Add operations to duplicate single or multiple services, with separate function for saving, which allow longer running duplicate operations (like agency or line duplication), to postpone saving when all objects are ready to reduce the transaction time.

I haven't read the code yet, but with a duplication operation, the reading and the writing should probably be part of the same transaction to avoid inconsistency.

tahini:
> > * Add operations to duplicate single or multiple services, with separate function for saving, which allow longer running duplicate operations (like agency or line duplication), to postpone saving when all objects are ready to reduce the transaction time.
> 
> I haven't read the code yet, but with a duplication operation, the reading and the writing should probably be part of the same transaction to avoid inconsistency.

What bugs me is that duplication operations can be quite long (like duplicating a complete agency like the STM), so if reading is part of the transaction, it will be long and may block other operations. I was thinking to read current data during duplication and only use a transaction once all the data is ready at the end. Though maybe once it is in the backend, it will turn out to be not so bad after all... we can benchmark then.

(Note that as other operations are migrated to the backend the ServiceUtils's function will receive an additional transaction argument).

packages/transition-backend/src/services/transitObjects/transitServices/ServiceDuplicator.ts
```diff
@@ -0,0 +1,92 @@
...
+ * @param options Duplication options
+ * @returns A status object with the IDs of the duplicated services, or an error
+ * status
+ */
+export const duplicateAndSaveServices = async (
+    serviceIds: string[],
+    options: DuplicateServiceOptions
+): Promise<Status.Status<string[]>> => {
+    try {
+        const duplicatedServices: Service[] = await duplicateServices(serviceIds, options);
```
greenscientist, line 34: We need a transaction to wrap the read and save

packages/transition-backend/src/services/gtfsImport/ServiceImporter.ts
```diff
@@ -16,7 +16,7 @@ import ServiceCollection from 'transition-common/lib/services/service/ServiceCol
 import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import { ServiceImportData, GtfsImportData } from 'transition-common/lib/services/gtfs/GtfsImportTypes';
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
-import { getUniqueServiceName } from 'transition-common/lib/services/service/ServiceUtils';
+import { getUniqueServiceName } from '../transitObjects/transitServices/ServiceUtils';
```
greenscientist, line 19: Do we delete the old getUniqueServiceName ?

packages/transition-common/src/services/service/ServiceDuplicator.ts
```diff
@@ -17,6 +17,7 @@ export interface DuplicateServiceOptions {
     serviceCollection?: ServiceCollection;
 }
 
+// @deprecated Service duplication should be done in the backend, but this function is still used in the frontend by other object's duplication
```
greenscientist, line 20: We should probably add a Logline somewhere saying we are using a deprecated function

packages/transition-frontend/src/components/forms/service/TransitServiceButton.tsx
```diff
@@ -48,23 +48,21 @@ const TransitServiceButton: React.FunctionComponent<ScheduleButtonProps> = (prop
...
+            [props.service.getId()],
+            { newServiceSuffix: props.t('main:copy') },
+            async (response: Status.Status<string[]>) => {
+                if (Status.isStatusOk(response)) {
+                    await serviceLocator.collectionManager
+                        .get('services')
+                        .loadFromServer(serviceLocator.socketEventManager, serviceLocator.collectionManager);
+                    serviceLocator.collectionManager.refresh('services');
+                } else {
+                    console.error(response.error); // todo: better error handling
```
greenscientist, line 62: I kind of wonder if somewhere along the call chain we should have an automated retry

greenscientist (changes requested):
I think the duplication is still error prone since we lack a synchronisation around the read and write parts.

greenscientist:
> > > * Add operations to duplicate single or multiple services, with separate function for saving, which allow longer running duplicate operations (like agency or line duplication), to postpone saving when all objects are ready to reduce the transaction time.
> > 
> > 
> > I haven't read the code yet, but with a duplication operation, the reading and the writing should probably be part of the same transaction to avoid inconsistency.
> 
> What bugs me is that duplication operations can be quite long (like duplicating a complete agency like the STM), so if reading is part of the transaction, it will be long and may block other operations. I was thinking to read current data during duplication and only use a transaction once all the data is ready at the end. Though maybe once it is in the backend, it will turn out to be not so bad after all... we can benchmark then.
> 
> (Note that as other operations are migrated to the backend the ServiceUtils's function will receive an additional transaction argument).

So technically, the operation could be quite fast and done in a single SQL command. I don't think that having some "longer" running transaction should be that bad. Our application is not write heavy, so should not block that much. (I would need to review the transaction documentation to be sure)

tahini:
I'll read more about locking.

My first idea was to read all the data, duplicate it (whence the duplicate without save functions), then take a transaction that (I expected) would lock the tables while the records are being written, as fast as possible.

But maybe it's possible to fine-tune locking, to be able to do it all in a transaction. Duplication does not update any records, it just inserts new ones so it should not lock technically any other operation by other users. If we can ensure that we don't lock other operations, then we could put the whole duplication into a single transaction and do the duplicate and save operation at once.

packages/transition-backend/src/services/gtfsImport/ServiceImporter.ts
tahini: > Do we delete the old getUniqueServiceName ?

Not yet, it is still used by lines/agency duplication while it is still in the frontend. Would be more trouble than it's worth to remove this now, knowing that all duplications will move to the backend shortly anyway.

packages/transition-common/src/services/service/ServiceDuplicator.ts
tahini: It will be removed shortly anyway

tahini:
Everything is in a transaction now. Only the inserted rows will be locked, so even if the duplication operation takes time, it won't block any other operation.

packages/transition-backend/src/services/gtfsImport/ServiceImporter.ts
greenscientist: It could be interesting to do add a TODO there to not forget to delete it.

packages/transition-common/src/services/service/ServiceDuplicator.ts
greenscientist: "shortly" ;)

greenscientist (commented):
Not hyper important, but if you update the commits, you have a few typos in the message: 
"services: Add argument to fetch partial colletion from DB
This adds an optional argument to the `collection` funtion, which"

packages/transition-backend/src/services/transitObjects/transitServices/ServiceDuplicator.ts
```diff
@@ -0,0 +1,82 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
```
greenscientist, line 6: We probably need a header to explain what does into that file. 
We have many ServiceSomething.ts which could make it confusing to decide where to put stuff.

greenscientist (approved):
I have some comments, but I'll accept this.

I still believe it would be more efficient to do the full copy in one SQL query instead of reading the data, duplicating it in JS and then doing another query. But I guess I can create an issue for that.

tahini:
The one-query SQL copy will not give you a unique name for the service, you'll still have to do selects to get and retrieve that name. That done, we'll have to do SQL insert/select queries one record at a time to set the unique name. Unless there's a functionality in postgres to do that that I do not know about.

tahini:
I know we can insert/select while updating only a few fields, but we still have to know (locally in typescript) the values of those fields, so you still have to fetch the original names/ids, find their unique equivalent, then insert/select each service individually. There will be no fewer queries this way, though a bit less data transferred between DB and client. But services aren't that big anyway. 

But for paths, this might be non-negligible. I'll try to keep the data exchanged minimal for the other objects, but we still have to keep the mapping between original IDs and new ones... anyway, just to say that it's not just a dump/simple copy of the original object.

greenscientist:
What was one of my question: Why do you need the mapping anyway?? That does not sounds really useful.

tahini:
> That was one of my question: Why do you need the mapping anyway?? That does not sounds really useful.

foreign key references. You need the new service ID to copy the schedules, you need the new path IDs to copy the trips, etc.

PR #948, opened 2024-05-16
pyTransition: Clean extra "/" at the end of an URL

greenscientist:
We already add the "/" at the begin of the path in the URLs, so we should remove the extra one if the use has added one when configuring his access


greenscientist:
Fixes: https://github.com/chairemobilite/transition_qgis/issues/13

greenscientist:
Thinking more about it, I might have something better. I'll try to instead using urllib to handle the URL. (And by I, I mostly mean chatgpt...)

greenscientist:
Doing #949 instead

PR #949, opened 2024-05-19
Use urllib functions to normalize URL

greenscientist:
We add a normalize_url function that remove extra characters from an URL and also add build_url function so that we don't manipulate URL as string everywhere

Add a test with an extra "/" at the end of the base URL

Created with the help of chatgpt

### DIFF
--- a/pyTransition/pyTransition/transition.py
+++ b/pyTransition/pyTransition/transition.py
@@ -23,13 +23,16 @@
 import requests
 from datetime import time
 import json
+from urllib.parse import urlparse, urlunparse, urlencode, parse_qs
 
 
 class Transition:
     def __init__(self, url, username, password, token=None):
         if url is None or url == "":
             raise ValueError("URL cannot be empty.")
-        self.base_url = url
+
+        # Parse and normalize the URL
+        self.base_url = self.__normalize_url(url)
 
         # To instantiate Transition instance from token only
         if username is None and password is None and token is not None:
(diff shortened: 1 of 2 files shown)


greenscientist:
@tahini , this could be better than #948

pyTransition/pyTransition/transition.py
```diff
@@ -39,6 +42,46 @@ def __init__(self, url, username, password, token=None):
...
+    def __normalize_url(url):
+        """Normalize the URL.
+
+        Remove extra characters from the URL
+        """
+        parsed_url = urlparse(url)
+        netloc = parsed_url.netloc.lower()
+
+        if netloc.endswith(':80') and parsed_url.scheme == 'http':
+            netloc = netloc[:-3]
```
tahini, line 55: What does this do? What's `:-3`?

pyTransition/pyTransition/transition_urlparse.py
```diff
@@ -0,0 +1,118 @@
+from urllib.parse import urlparse, urlunparse, urlencode, parse_qs
+
+class Transition:
```
tahini, line 3: Remove this file?

tahini (changes requested):
You seem to have committed a copy of the file, unless I'm mistaken

pyTransition/pyTransition/transition.py
greenscientist: You'll have to ask chatgpt... ;)

Just kidding, it remove the extra :80 if we already specified http as protocol.

tahini (approved):
c'est bon! j'imagine que tu as testé (je vois que le CI ne roule pas encore les tests python)

tahini:
@greenscientist Est-ce que pip a été mis à jour avec cette nouvelle version ou il faut faire qqc manuellement?

greenscientist:
Pas encore, je dois faire un release. 
Aujourd'hui, ou demain.

Issue #286, opened 2021-09-03
transition: Revisit the types of the routing results


tahini:
With the multiple modes routing calculation, a lot of types were added to describe the results, there are a few typeguard functions in the `transition-common/src/services/transitRouting` classes to check correct types. And a bit of hard-coded code was added in the result display components to handle the special case of transit and walk only routes (the walk only route is a different type that other routing path).

We should have a common interface that all routing results can implement and not have, or at least minimize, any custom code per mode.

tahini:
fixed by #963

Issue #954, opened 2024-05-23
routing: Move most of routing calculation functionalities to chaire-lib-backend


tahini:
Currently, the routing calculation (both transit and other modes) is concentrated in the `TransitRoutingCalculator` file, in transition-common. That means from the frontend, each mode is a request to the backend and backend callers (batch routing, public API) pass through a fake socket event to call the modes.

Part of those calculations are not transition specific (only getting the geojson data for the transit routes require an actual PathCollection from Transition, but they could be gotten in another static way for other consumers). If they were in chaire-lib, Evolution could also make use of those functions.

Also, keeping the implementation details in the backend will make the frontend lighter.

The files that will need to be moved/refactored are `TransitRoutingCalculator.ts`, `TransitRoutingResult.ts` and `RouteCalculatorResult.ts`.

tahini:
completed with a678fe89261744bb416535793b05e668d147a642

PR #963, opened 2024-05-28
Move RoutingResult classes to chaire-lib, add tests and use common interface

tahini:
This commit is part of work to move the routing calculations to chaire-lib instead of Transition, so that other tools can benefit from it (#954).

The main result interface is called `RoutingResult`. 2 classes implement it `UnimodalRoutingResult` and `TransitRoutingResult`. There is no need to call the `getWalkOnlyRoute` method anymore (it has been removed), as it is not part of the interface and required custom code. Rather, the `getPath` method will return either a Route or TrRoutingRoute in the `TransitRoutingResult` class.

Tests have been added for the results class. `TransitRoutingResult` required to be split between the common chaire-lib part (the class itself) and the transition part (requiring a PathCollection to get the geojson from transition paths). A parameter allows to specify how to retrieve the geojson from the route data.

The result classes may still be improved later, as the route calculations are also moved to chaire-lib, to find more common grounds or add better types to route properties. This is mostly moving what currently exists, with minimal changes to existing code.

### DIFF
--- a/packages/chaire-lib-common/src/services/routing/RoutingResult.ts
+++ b/packages/chaire-lib-common/src/services/routing/RoutingResult.ts
@@ -12,6 +12,10 @@ import { RoutingOrTransitMode, RoutingMode } from '../../config/routingModes';
 import TrError, { ErrorMessage } from '../../utils/TrError';
 import { TrRoutingRoute } from '../trRouting/TrRoutingService';
 
+export const pathIsRoute = (path: Route | TrRoutingRoute | undefined): path is Route => {
+    return typeof (path as any).distance === 'number';
+};
+
 // TODO Add a common type to getPath(index)
 // TODO Have a common type for all results, not requiring the TResultData generic type
 /**
@@ -82,10 +86,6 @@ export class UnimodalRoutingResult implements RoutingResult<UnimodalRoutingResul
             type: 'FeatureCollection',
             features: [this._params.origin, this._params.destination]
         };
-    }
-
-    getWalkOnlyRoute(): Route | undefined {
-        return undefined;
     }
 
     async getPathGeojson(index: number, _options: { [key: string]: any } = {}): Promise<GeoJSON.FeatureCollection> {
--- a/packages/chaire-lib-common/src/services/routing/RoutingResult.ts
+++ b/packages/chaire-lib-common/src/services/routing/RoutingResult.ts
@@ -21,6 +21,17 @@ export interface RoutingResult<TResultData> {
     hasAlternatives: () => boolean;
     getAlternativesCount: () => number;
     originDestinationToGeojson: () => GeoJSON.FeatureCollection<GeoJSON.Point>;
+    /**
+     * Get the geojson geometry of the path at the given index
+     *
+     * TODO Type the options so they are common for all routing results
+     *
+     * TODO Type the FeatureCollection properties
+     *
+     * @param index The index of the alternative path to fetch
+     * @param options Additional options to pass to generate geometries
+     * @returns A feature collection with the steps of the path
+     */
     getPathGeojson: (index: number, options: { [key: string]: any }) => Promise<GeoJSON.FeatureCollection>;
     getPath: (index: number) => TrRoutingRoute | Route | undefined;
     getRoutingMode(): RoutingOrTransitMode;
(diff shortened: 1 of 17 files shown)


greenscientist:
Without having looked at the code, I would name the TransitRoutingResult MultimodalRoutingResult. 
(Would a bike->carshare route be called Transit or something else ??)

tahini:
> Without having looked at the code, I would name the TransitRoutingResult MultimodalRoutingResult. (Would a bike->carshare route be called Transit or something else ??)

Currently, the class is really for transit result. When we really have Multi modal, we can revisit it and make it really generic for this kind of data.

packages/chaire-lib-common/src/services/routing/RoutingResult.ts
```diff
@@ -6,36 +6,63 @@
...
-import { TrRoutingRoute } from 'chaire-lib-common/lib/services/trRouting/TrRoutingService';
+import Preferences from '../../config/Preferences';
+import { Route } from './RoutingService';
+import { RoutingOrTransitMode, RoutingMode } from '../../config/routingModes';
+import TrError, { ErrorMessage } from '../../utils/TrError';
+import { TrRoutingRoute } from '../trRouting/TrRoutingService';
+
+export const pathIsRoute = (path: Route | TrRoutingRoute | undefined): path is Route => {
+    return typeof (path as any).distance === 'number';
+};
```
kaligrafy, line 17: Some transit route have no distance calculated if the shape of the path is not provided in the GTFS. Should we chack for travel time instead?

kaligrafy (changes requested):

packages/chaire-lib-common/src/services/routing/RoutingResult.ts
kaligrafy: OK, but could we have a route result with only a travel time/duration and no distance? A ferry route by instance?

kaligrafy: I would say that routing engine wil alsways return a duration, but not necessarly a distance

tahini: I definitely cannot know about "always", either for the duration or the distance. But osrm's doc says it returns a distance and we, when we handle the routing results, always set it and the field is mandatory. Maybe it's 0, or -1, but it's there and a number!

kaligrafy: OK then !

kaligrafy (approved):

tahini:
Incidentally also fixes #286

packages/chaire-lib-common/src/services/routing/RoutingResult.ts
```diff
@@ -6,36 +6,63 @@
...
 
 // TODO Add a common type to getPath(index)
-export interface RouteCalculatorResult<InputParams> {
+// TODO Have a common type for all results, not requiring the TResultData generic type
```
tahini, line 20: This TODO will be done in the next PR. The `getParams()` is only used in the backend to convert back from the object to the data, but the backend calculation will return the data directly and we won't need to pass by the object.

greenscientist (approved):
I think it's still a bit more complex, but I will wait and see the next steps in the refactoring, hopefully, complexity will reduce then.

packages/chaire-lib-common/src/services/routing/RoutingResult.ts
```diff
@@ -6,36 +6,63 @@
...
+     * @param options Additional options to pass to generate geometries
+     * @returns A feature collection with the steps of the path
+     */
     getPathGeojson: (index: number, options: { [key: string]: any }) => Promise<GeoJSON.FeatureCollection>;
     getPath: (index: number) => TrRoutingRoute | Route | undefined;
     getRoutingMode(): RoutingOrTransitMode;
     hasError: () => boolean;
     getError: () => TrError | undefined;
-    getParams: () => InputParams;
+    getParams: () => TResultData;
```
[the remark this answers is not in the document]
tahini: > Params sounds like Input data, not result. Is the getParams still a good name for the function.

Cette fonction disparaît dans la PR suivante...

PR #967, opened 2024-05-31
db: Add the updated_at field to the tokens table

tahini:
The table was built with the `onUpdateTrigger` constraint, but the updated_at field is not present, so it is not possible to update the table. Also rename `creation_date` to `created_at` to match other tables.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/tokens.db.queries.ts
+++ b/packages/chaire-lib-backend/src/models/db/tokens.db.queries.ts
@@ -18,30 +18,6 @@ const defaultTokenLifespanDays: number = isNaN(Number(config.tokenLifespanDays))
     ? 10
     : Number(config.tokenLifespanDays);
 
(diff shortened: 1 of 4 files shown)


packages/chaire-lib-backend/src/models/db/tokens.db.queries.ts
```diff
@@ -34,12 +34,12 @@ const attributesParser = (dbAttributes: {
     user_id: number;
     apiToken: string;
     expiryDate: string;
-    creationDate: string;
+    created_at: string;
```
kaligrafy, line 37: Why not camelcase like expiryDate and apiToken here? Likewise for user_id? userId?

tahini: Honnêtement, parser et cleaner ne sont pas utilisés, aussi ben les enlever.

greenscientist (approved):
Looks good to me

PR #976, opened 2024-06-07
deck.gl: Support the max/minRadiusPixels properties

### LINKED ISSUE #974: deck.gl: After a certain zoom level, nodes get too big
The nodes should have a maximum size

cc @tibobliss

tahini:
fixes #974
    
The `maxRadiusPixels` and `minRadiusPixels` can be set in the layer
description to prevent the points from getting too big/small when
zooming in/out. If the property is a number, the lineWidth min/max
pixels are also set to 1/3 of the corresponding radius.

Here's how it looks now (@kaligrafy 10 review points if you find where this image shows :p ):

Capture d’écran du 2024-06-07 14-32-26


kaligrafy:
Boulevard Roberval, à Longueuil

kaligrafy (approved):

PR #980, opened 2024-06-10
routing: Allow service manager to receive an array of engines

tahini:
This allows clients to specify various routing engines, in order of preference, some of which may not apply to all modes. Only the first available one will be returned by the RoutingServiceManager.

### DIFF
--- a/packages/chaire-lib-common/src/services/routing/RoutingServiceManager.ts
+++ b/packages/chaire-lib-common/src/services/routing/RoutingServiceManager.ts
@@ -32,9 +32,10 @@ class RoutingServiceManagerImpl implements RoutingServiceManager {
         };
     }
 
-    public getRoutingServiceForEngine = (engine: string): RoutingService => {
-        const service = this._routingServices[engine];
-        return service ? service : this._defaultService;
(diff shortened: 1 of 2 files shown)


greenscientist:
What is the intent of this? I don't see why it's useful

tahini:
This is part of the calculation refactoring. It will allow to more transparently support transit or routing engines together. The RoutingServiceManager will pick whichever applies first

tahini:
It's just so the callers don't have to worry about whether the request is for transit or routing, but just ask its modes and engines and the calculation will know what to do with it.

greenscientist:
Do we really want to expose this to the client? I was thinking the client only ask for a mode, then the backend pick the engine for this specific mode.

tahini:
Transition supporte déjà le mode manuel et engine selon si on veut forcer le passage sur une route non-existente, donc non, on ne peut pas vraiment juste laisser le backend décider. En tout cas, pas à moins de faire du code custom du côté de Transition pour l'édition de Path

greenscientist:
And what would be an exemple for this use case?

kaligrafy (approved):

tahini:
I'll put this commit as part of the routing calculation refacotirng PR later

PR #982, opened 2024-06-11
Move routing calculation to chaire-lib-backend

tahini:
Add an review all data types regarding transit routing queries to chaire-lib-common.

Add a `Routing.ts` file which contains the main calculation function. This function takes a json object with the query parameters as argument and return results also as json objects, for easier serialization if necessary.

The `TransitRoutingCaculator` from transition-common is replaced by the `Routing`'s chaire-lib-backend calculator and a `RouitngUtils` file in the transition-frontend to make the call to the routing socket route on the server.

In the frontend, the `TransitRouting` object is kept as it is easier to edit from the form by the user, but in the backend, it is all replaced with the parameter objects from chaire-lib-common.

A utility function can transform the result json objects to RoutingResult instances for easier manipulation by result handling functionalities.


packages/chaire-lib-common/src/services/routing/RoutingServiceManager.ts
```diff
@@ -32,8 +32,8 @@ class RoutingServiceManagerImpl implements RoutingServiceManager {
         };
     }
 
-    public getRoutingServiceForEngine = (engine: string): RoutingService => {
-        const service = this._routingServices[engine];
+    public getRoutingServiceForEngine = (engine?: string): RoutingService => {
```
greenscientist, line 35: Would be great if this would be moved into either the backend or the frontend.

greenscientist (approved):
Things are surprisingly not simpler. 

I imagine it's all necessary, but it's a lot of change for a small split of frontend/backend.

tahini:
There's already an issue to make it simpler: #983 

At least this limits the number of callers to the inners of the calculation

PR #987, opened 2024-06-20
Add support to enable DKIM signature in email sending

greenscientist:
This add the necessary config to let nodemailer add a DKIM signature to sent email

We need 3 news environment variables:
MAIL_TRANSPORT_DKIM_SELECTOR (Select which DKIM key to use) MAIL_TRANSPORT_DKIM_DOMAIN (Domain to select the key from) MAIL_TRANSPORT_DKIM_PRIVATE_PATH (File path to the DKIM private key)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -55,5 +55,18 @@ MAIL_TRANSPORT_SMTP_SECURE=false
 MAIL_TRANSPORT_SMTP_AUTH_USER=myUser
 MAIL_TRANSPORT_SMTP_AUTH_PWD=password
 
+# DKIM (DomainKeys Identified Mail) signature related configuration
+# You need all three if you want to use DKIM, but you don't have to set them if you
+# do not want to add the DKIM signature to the emails
+#
+## Prefix to use to select the public key. For example, this will match the record
+## dkim._domainkey.example.com
+# MAIL_TRANSPORT_DKIM_SELECTOR=dkim
+## DNS Domain to match the public key lookup
+# MAIL_TRANSPORT_DKIM_DOMAIN=example.com
+## File path to the private key matching the one in the Selector key
+# MAIL_TRANSPORT_DKIM_PRIVATE_PATH=/path/to/privkey.pem
+
+
 # From email
 MAIL_FROM_ADDRESS=example@example.com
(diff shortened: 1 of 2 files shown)


greenscientist:
Tested sending email with and without the new environment variable
Also tested with invalid options

packages/chaire-lib-backend/src/services/mailer/transport.ts
```diff
@@ -5,14 +5,48 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 import nodemailer, { Transporter } from 'nodemailer';
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
+import fs from 'fs';
+import DKIM from 'nodemailer/lib/dkim';
+
+// Generate the config block for DKIM if the right ENV variables are defined
+const getDkimConfig = (): DKIM.Options | undefined => {
```
tahini, line 13: You probably know exactly what dkim is, but for readers neophyte in email configuration, an additional sentence explaining why they should google dkim to have more information about it would be helpful. So just why we need a dkim config, what problem it fixes.

tahini (changes requested):
Some more comments required.

Also, please add those 3 variables in the .env.example file, in the mail configuration block, with documentation and maybe an example of the expected data.

.env.example
```diff
@@ -55,5 +55,18 @@ MAIL_TRANSPORT_SMTP_SECURE=false
 MAIL_TRANSPORT_SMTP_AUTH_USER=myUser
 MAIL_TRANSPORT_SMTP_AUTH_PWD=password
 
+# DKIM (DomainKeys Identified Mail) signature related configuration
+# You need all three if you want to use DKIM, but you don't have to set them if you
+# do not want to add the DKIM signature to the emails
```
tahini, line 60: It applies to all email strategies? If so, you have nothing to add here, but if it applies to only one, please add a comment

tahini (approved):
Just one comment, you can merge if it's ok, otherwise, add the missing information.

.env.example
greenscientist: Yeah, all

greenscientist:
Exceptionnally I will merge the PR, to get it in prod before the weekend

PR #992, opened 2024-06-26
socket: Set the `maxHttpBufferSize` when initializing socket

### LINKED ISSUE #991: Can't upload large files
Since the socket.io upgrade, we can't upload files larger than 1MB.

There was this warning in the release notes:

⚠️ WARNING ⚠️

The default value of the maxHttpBufferSize option has been decreased from 100 MB to 1 MB, in order to prevent attacks by denial of service.

We have not heeded it, let's fix it now.

tahini:
fixes #991

The latest socket.io upgrade set the upload size default to 1MB. We set it back to 100MB since Transition users often have to upload large files for calculations.

### DIFF
--- a/packages/transition-backend/src/socketServerApp.ts
+++ b/packages/transition-backend/src/socketServerApp.ts
@@ -65,7 +65,8 @@ const setupSocketServerApp = async function (server, session) {
 
     const io = socketIO(server, {
         pingTimeout: 60000,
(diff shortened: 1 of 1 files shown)


greenscientist (approved):
Wondering if it would be worth it to also summarize the commit message as a comment in the code. 
I'll let you decide, I'm not sure myself.

PR #994, opened 2024-06-27
api: fix `/api/v1/nodes` when stops are undefined

### LINKED ISSUE #990: API : Can't access api/v1/nodes
Scenarios and paths are accessible, but not nodes. Also goes for the QGIS plugin: we can fetch paths but nothing happens when fetching nodes.

### THIS WORKS
image
image

### BUT THIS DOESN'T

image
image

tahini:
fixes #990

Nodes that are not imported from gtfs may not have the `stops` data property defined. It is not mandatory anyway, so we should not expect it to exist.

Also add unit tests for the NodesAPIResponse class.

### DIFF
--- a/packages/transition-backend/src/api/public/NodesAPIResponse.ts
+++ b/packages/transition-backend/src/api/public/NodesAPIResponse.ts
@@ -7,25 +7,19 @@
 import { Feature, FeatureCollection, Point } from 'geojson';
 import APIResponseBase from './APIResponseBase';
 
-export type NodesAPIResponseFormat = {
-    type: 'FeatureCollection';
-    features: Array<{
-        type: 'Feature';
-        id: number;
-        geometry: Point;
-        properties: {
-            id: string;
-            code: string;
-            name: string;
-            stops: Array<{
-                id: string;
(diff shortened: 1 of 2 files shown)


packages/transition-backend/src/api/public/NodesAPIResponse.ts
```diff
@@ -39,12 +33,13 @@ export default class NodesAPIResponse extends APIResponseBase<NodesAPIResponseFo
...
                     code: feature.properties!.code,
                     name: feature.properties!.name,
-                    stops: feature.properties!.data.stops.map((stop) => ({
-                        id: stop.id,
-                        code: stop.code,
-                        name: stop.name,
-                        geography: stop.geography
-                    }))
+                    stops:
+                        feature.properties!.data.stops?.map((stop) => ({
```
greenscientist, line 37: I feel that some day we should encapsulate those conversion in better object oriented structure. (But I don't know what would be best. Anyway, just thoughts for the future.

greenscientist (approved):

kaligrafy (approved):

PR #998, opened 2024-06-28
dep: Upgrade `passport` and `passport-magic-login`

tahini:
fixes https://github.com/chairemobilite/transition/security/dependabot/44

* passport: 0.6.0 => 0.7.0
* passport-magic-login: 1.0.9 => 1.2.2


greenscientist (approved):
ok. 

Changelog from 0.6 to 0.7 is minimal

PR #1000, opened 2024-06-28
Enable build on Node 20.x in github workflow

greenscientist:
Node 20 is the current LTS, we should migrate our builds to it


greenscientist:
replaced by #1035

PR #1001, opened 2024-06-28
routing: Support alternatives for single-mode routing calculations

tahini:
This requires to pass down the trip routing query attributes all the way down to the OSRM call. It is added as an optional parameter at the end of the `route` and other calculation functions, as well as in the socket route for routing calculation.

If the `withAlternatives` attribute is `true` , the osrm backend call sets the alternatives attribute to `true`.

### DIFF
--- a/packages/chaire-lib-backend/src/utils/osrm/OSRMMode.ts
+++ b/packages/chaire-lib-backend/src/utils/osrm/OSRMMode.ts
@@ -14,6 +14,7 @@ import osrm from 'osrm'; // Types from the osrm API definition see https://githu
 import * as RoutingService from 'chaire-lib-common/lib/services/routing/RoutingService';
 import * as Status from 'chaire-lib-common/lib/utils/Status';
 import { transitionRouteOptions, transitionMatchOptions } from 'chaire-lib-common/lib/api/OSRMRouting';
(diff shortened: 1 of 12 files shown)


packages/chaire-lib-backend/src/utils/osrm/OSRMMode.ts
```diff
@@ -66,14 +67,17 @@ class OSRMMode {
         }
     }
 
-    public async route(params: transitionRouteOptions): Promise<Status.Status<osrm.RouteResults>> {
+    public async route(
+        params: transitionRouteOptions,
+        routingAttributes?: TripRoutingQueryAttributes
+    ): Promise<Status.Status<osrm.RouteResults>> {
```
greenscientist, line 73: Instead of adding a new parameter, we already have the params object which include the boolean for alternatives. We should probably fill it in there before calling this function.

greenscientist:
@tahini , see my (old) comment

packages/chaire-lib-backend/src/utils/osrm/OSRMMode.ts
tahini: I will add it to the transitionRouteOptions as a new property, but we won't use the one from  osrm as it should be eventually rewritten (see #1115)

greenscientist (approved):

PR #1002, opened 2024-07-02
dep: upgrade version of `request-ip`

tahini:
* request-ip: 2.1.3 => 3.3.0

This removes dependency to is_js, which has a vulnerability issue.

### DIFF
--- a/packages/transition-backend/package.json
+++ b/packages/transition-backend/package.json
@@ -57,7 +57,6 @@
         "pbf": "^3.2.1",
         "proj4": "^2.7.5",
         "random": "^3.0.6",
-        "request-ip": "^2.1.3",
         "serve-favicon": "^2.5.0",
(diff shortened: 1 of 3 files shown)


greenscientist (approved):
Yes, finally one that remove some deps!

greenscientist:
Checking the code quickly, I'm wondering if we really need this package...

greenscientist:
It's used in evolution (look for clientIp), but not in Transition.

tahini:
I was wondering if it's being used passively by something, like just using it gives some data somewhere, but maybe not.

greenscientist:
I've looked at the package and it juste add the clientIp field.

greenscientist (changes requested):
Need to update the commit message.

tahini:
C'est fait

greenscientist (approved):

PR #1003, opened 2024-07-02
Send routingPort all the way throught the stack when doing batch calc…

greenscientist:
…ulation

Recent refactoring have dropped the routingPort, which was used for handling batch calculation to a separate trRouting instance.
This add the field to the type and create an hostport parameter before calling the trRouting interface


greenscientist:
Locally tested the "single" routing calculation and the batch calculation. Confirmed in the TrRouting logs that the request were sent to the right one.

PR #1018, opened 2024-07-25
Enable code coverage generation in test execution

greenscientist:


tahini (approved):

tahini:
Will this be used for anything else than just adding more verbosity to the test output (we can run the coverage locally to get this information too). Like show a timeline with coverage improvement? Or add a badge to our packages?

greenscientist:
Thinking of adding a github action that will comment on PRs and maybe block them if coverage gets too low. 

I'd love to have a timeline, but not sure how feasable this is in the github environment.

Issue #214, opened 2021-03-12
transition: killall trRouting command will stop all instances of trRouting running on the same machine


kaligrafy:
Make sure we only quit trRouting instances associated with the current project

See socketServerApp line 25

greenscientist:
We don't use killall anymore

PR #1020, opened 2024-07-25
Autostart TrRouting on server startup

greenscientist:
With the use case of the API, we might have user that will not login in the web interface to manually start TrRouting.

Since we now recreate the cache automatically on startup, we can also start TrRouting.

We use a delayed await, so that we can do other stuff while trRouting start

We removed the line that would kill other TrRouting, since we now use the restart function

### DIFF
--- a/packages/transition-backend/src/socketServerApp.ts
+++ b/packages/transition-backend/src/socketServerApp.ts
@@ -12,6 +12,7 @@ import events from 'events';
 
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import OSRMProcessManager from 'chaire-lib-backend/lib/utils/processManagers/OSRMProcessManager';
+import trRoutingProcessManager from 'chaire-lib-backend/lib/utils/processManagers/TrRoutingProcessManager';
 import { _booleish } from 'chaire-lib-common/lib/utils/LodashExtensions';
 
 import { recreateCache } from './services/capnpCache/dbToCache';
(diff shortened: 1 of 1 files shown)


greenscientist:
Handle #214

kaligrafy (approved):

greenscientist:
I've tested in a docker environnement. You might want to test in a "native" env.

tahini (approved):
I tested locally, the process automatically starts. It does not kill any previously running trRouting process (started outside Transition).

But if the outside trRouting process is running on port 4000, Transition's trRouting startup does not work correctly. The queries from the interface are executed and results come back fine. But the icon remains yellow and clicking on it gives an error restarting trRouting.

It's not blocking as the user gets results anyway and it won't happen in the public instances. But if you know how to solve it quickly, you can fix it now. Otherwise, you can open an issue to remember this (or we can just get rid of the icon! Ni vu, ni connu...)

greenscientist:
You are right, it will only kill a process that was saved in the PID file. 

I don't think I can do a quick fix. A good solution would be to detect that the port is in use and select a different one. But that's a bit more complicated. 

But I still prefer the new situation than killing a random process that might be not related to running transition. And it allows you to start a different trRouting ;)

tahini:
Don't get me wrong, I totally agree you should never kill a random process! The [soft] problem is just the signals that the user get: error starting the process, yet plausible results (supposing the trRouting process that answers was started with the right cache... otherwise, simply wrong results!)

Issue #1028, opened 2024-08-07
routing: The `Routing`.`calculate` function should query all modes in parallel


tahini:
Currently, for each mode, we create a result object, where we await the result. Which means all modes are done sequentially instead of in parallel. We should make them parallel, especially that each mode makes a call to a different routing server.

The function is in `chaire-lib-backend/src/services/routing/Routing.ts`

tahini:
@greenscientist si tu as du temps à optimiser...

greenscientist:
Did the code and ran some benchmarks. Currently, what takes the most time is the transit calculation. All the other ones are OSRM call which are quite fast. Would still be worth it to do, but the gain is minimal. Maybe about a second in saving.

PR #1033, opened 2024-08-13
Move the await outside of the for loop in the Routing::calculate func…

greenscientist:
…tion

The await in the for loop made the call to each RoutingMode sequential. The the promise generation and the promise.all after the loop, we can make all the API call in parallel. Those are mostly fast OSRM calls which are quite fast, but we can still have a small gain.

We moved the core of the calculate function into a private calculateRoutingMode function. The later handle a single mode, and the former handle the loop, the awaiting and the result handling.

Closes: #1028

### DIFF
--- a/packages/chaire-lib-backend/src/services/routing/Routing.ts
+++ b/packages/chaire-lib-backend/src/services/routing/Routing.ts
@@ -121,51 +121,30 @@ export class Routing {
         const { isCancelled } = options;
 
         // ** backend
-        const routingResult: TransitOrRouteCalculatorResult[] = [];
+        const routingPromises: Promise<TransitOrRouteCalculatorResult>[] = [];
 
-        for (let i = 0, count = routingAttributes.routingModes.length; i < count; i++) {
+        console.log(`Routing beginning with ${routingAttributes.routingModes.length} modes`);
+        for (const routingMode of routingAttributes.routingModes) {
             if (isCancelled && isCancelled()) {
                 throw 'Cancelled';
             }
-
-            const routingMode = routingAttributes.routingModes[i];
-            try {
-                if (routingMode === 'transit') {
-                    routingResult.push({
-                        routingMode,
-                        result: await calculateTransit(
-                            [routingAttributes.originGeojson, routingAttributes.destinationGeojson],
-                            routingAttributes
-                        )
-                    });
-                } else {
-                    routingResult.push({
-                        routingMode,
-                        result: await calculateRoute(
-                            [routingAttributes.originGeojson, routingAttributes.destinationGeojson],
-                            routingAttributes,
-                            routingMode
-                        )
-                    });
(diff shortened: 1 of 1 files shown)


greenscientist:
Further benchmark notes: For a route from the north shore of MTL to about St-Jean, it took previously about 0.3s on my computer and with the change about 0.05s.

greenscientist:
Done with the help of Meta.ai

kaligrafy (approved):

packages/chaire-lib-backend/src/services/routing/Routing.ts
```diff
@@ -121,60 +121,77 @@ export class Routing {
         const { isCancelled } = options;
 
         // ** backend
-        const routingResult: TransitOrRouteCalculatorResult[] = [];
+        const routingPromises: Promise<TransitOrRouteCalculatorResult>[] = [];
```
GabrielBruno24, line 124: Should be routingResultPromises

packages/chaire-lib-backend/src/services/routing/Routing.ts
```diff
@@ -121,60 +121,77 @@ export class Routing {
...
-                            `cannot calculate routing for mode ${routingMode}:  ${error}`,
-                            'TRCalculatorError',
-                            { text: 'transit:transitRouting:errors:ErrorForMode', params: { mode: routingMode } }
-                        )
-                        : error
-                });
-            }
+        // Cancel further processing if the request was cancelled
+        if (isCancelled && isCancelled()) {
+            throw 'Cancelled';
```
GabrielBruno24, line 136: Do we need this both in the loop and after?

packages/chaire-lib-backend/src/services/routing/Routing.ts
GabrielBruno24, line 124: Should be routingResultPromises

packages/chaire-lib-backend/src/services/routing/Routing.ts
GabrielBruno24, line 136: Do we need this both in the loop and after?

greenscientist: The one in the loop check cancellation at the beginning, the one outside check after we added all the calls. The idea with cancellation is that you need to check often to interrupt the processing. 
It is really needed, I don't know, we should check in which context it's really used.

packages/chaire-lib-backend/src/services/routing/Routing.ts
greenscientist: Could be. Probably not worth it to change it. The function is small enough that it's easy to see what it is.

samuel-duhaime (approved):

GabrielBruno24 (approved):

Issue #1011, opened 2024-07-16
FetchError on TrRouting on long running batchrouting


greenscientist:
During a specific big batch calculation, we get the following error message. Calculation continue, but seems something that might be bad from the TrRouting side. 

2024-07-16T14:12:05.956561133Z FetchError: request to http://localhost:14000/v2/route?origin=-73.572,45.5061&destination=-73.1775,45.5308&scenario_id=27ef6b8a-af95-491b-8015-e59d67f81cc1&time_of_trip=57600&time_type=0&alternatives=true&min_waiting_time=180&max_access_travel_time=1200&max_egress_travel_time=1200&max_transfer_travel_time=1200&max_travel_time=14400&max_first_waiting_time=3600 failed, reason: socket hang up
2024-07-16T14:12:05.956604916Z     at ClientRequest.<anonymous> (/app/node_modules/node-fetch/lib/index.js:1491:11)
2024-07-16T14:12:05.956611874Z     at ClientRequest.emit (node:events:517:28)
2024-07-16T14:12:05.956616000Z     at Socket.socketOnEnd (node:_http_client:525:9)
2024-07-16T14:12:05.956619494Z     at Socket.emit (node:events:529:35)
2024-07-16T14:12:05.956622941Z     at endReadableNT (node:internal/streams/readable:1400:12)
2024-07-16T14:12:05.956626388Z     at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
2024-07-16T14:12:05.956630402Z   type: 'system',
2024-07-16T14:12:05.956633899Z   errno: 'ECONNRESET',
2024-07-16T14:12:05.956637313Z   code: 'ECONNRESET'
2024-07-16T14:12:05.956640735Z }
2024-07-16T14:12:05.956644358Z FetchError: request to http://localhost:14000/v2/route?origin=-73.572,45.5061&destination=-73.1775,45.5308&scenario_id=27ef6b8a-af95-491b-8015-e59d67f81cc1&time_of_trip=57600&time_type=0&alternatives=true&min_waiting_time=180&max_access_travel_time=1200&max_egress_travel_time=1200&max_transfer_travel_time=1200&max_travel_time=14400&max_first_waiting_time=3600 failed, reason: socket hang up
2024-07-16T14:12:05.956647814Z     at ClientRequest.<anonymous> (/app/node_modules/node-fetch/lib/index.js:1491:11)
[...]

PR #1045, opened 2024-09-12
debug: Add logging before/after handling batch routing trip

tahini:
Tentatively trying to investigate issue #1011 on instances. This will allow to see if the checkpoint handling caused issues, as the problem seems to come from "old" trip queries, ie that should have long been handled.

This commit can be reverted once the issue is solved


kaligrafy (approved):

greenscientist:
This seems useful to have, I don't think we should aim to revert!

PR #1048, opened 2024-09-13
debug: Add loggin before/after various routing functions

tahini:
To investigate issue #1011 on instances, we add logs before and after
the calculation functions, each individual mode calculation and the
trRouting requests. The log messages include the origin/destination
coordinates to allow to track individual queries. These can be matched
with the odTripIndex from a log message in the batch routing task.

### DIFF
--- a/packages/chaire-lib-backend/src/services/routing/Routing.ts
+++ b/packages/chaire-lib-backend/src/services/routing/Routing.ts
@@ -123,7 +123,10 @@ export class Routing {
         // ** backend
         const routingPromises: Promise<TransitOrRouteCalculatorResult>[] = [];
 
-        console.log(`Routing beginning with ${routingAttributes.routingModes.length} modes`);
+        const origDestStr = `${routingAttributes.originGeojson.geometry.coordinates.join(',')} to ${routingAttributes.destinationGeojson.geometry.coordinates.join(',')}`;
+        console.log(
+            `tripRouting: Routing beginning with ${routingAttributes.routingModes.length} modes for ${origDestStr}`
+        );
         for (const routingMode of routingAttributes.routingModes) {
             if (isCancelled && isCancelled()) {
                 throw 'Cancelled';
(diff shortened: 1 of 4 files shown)


packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -145,7 +145,8 @@ class TrRoutingBatch {
                     console.log(
                         'calc/sec',
                         Math.round(
-                            (100 * completedRoutingsCount) / ((1 / 1000) * (performance.now() - benchmarkStart))
+                            (100 * (completedRoutingsCount - startIndex)) /
```
greenscientist, line 148: This is starting to get complex, would benefit having a comment

packages/chaire-lib-backend/src/services/routing/Routing.ts
```diff
@@ -123,7 +123,10 @@ export class Routing {
         // ** backend
         const routingPromises: Promise<TransitOrRouteCalculatorResult>[] = [];
 
-        console.log(`Routing beginning with ${routingAttributes.routingModes.length} modes`);
+        const origDestStr = `${routingAttributes.originGeojson.geometry.coordinates.join(',')} to ${routingAttributes.destinationGeojson.geometry.coordinates.join(',')}`;
```
greenscientist, line 126: could be nice to be in a "toString" function instead

greenscientist (approved):
Accepting, but the             const origDestStr = `${odTrip.attributes.origin_geography.coordinates.join(',')} to ${odTrip.attributes.destination_geography.coordinates.join(',')}`;
that repeats in many place would benefit being in a function that simply both the origin and destination

tahini:
Added a comment for calculation, I'll open an issue for the utility function, as we may not necessarily want to keep the logs as is for the future. This is for very short term debugging

greenscientist:
> Added a comment for calculation, I'll open an issue for the utility function, as we may not necessarily want to keep the logs as is for the future. This is for very short term debugging

Ok. But printing an OD pair seems to be a very useful things for us in general!

PR #1056, opened 2024-09-18
API: Change `summary` endpoint from GET to POST

tahini:
This API should mirror the `route` API, which is also a POST. While they could be GET API, as the same parameters expect to return the same data and it does not trigger any data update in the system, the fact that they pass geojson features makes it easier to use a POST request. For a GET, ideally, only the coordinates should be returned.

### DIFF
--- a/docs/APIv1/API.yml
+++ b/docs/APIv1/API.yml
@@ -85,7 +85,7 @@ paths:
                 $ref: 'routingModes.yml#/routingModes'
   
   /api/v1/summary:
-    get:
+    post:
       summary: Get summary of route calculation result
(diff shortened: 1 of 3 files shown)


greenscientist (commented):
Is it possible to add a unit test that would have catched the issue of the fact that we were using a GET instead lf a POST??

tahini:
> Is it possible to add a unit test that would have catched the issue of the fact that we were using a GET instead lf a POST??

I wish. We are using `supertest`. Ideally, it would not accept a `send` call on the `get` request, but it does.

GabrielBruno24 (approved):

PR #1061, opened 2024-09-20
Move the TransitRoutingService to the backend

tahini:
* Move the types used in the current `TrRoutingService` to their own files
* Rename the current `TrRoutingService` to `TransitRoutingService`, which seems less tied to the `trRouting` calculator

* Move the `TransitRoutingService` class to the backend and delete the `TrRoutingServiceManager` who is not necessary anymore. This also allows to remove the socket routes used for the trRouting queries, as all frontend component already use the `routing.calculate` socket route to direct all their calculations to the backend.


kaligrafy (approved):

tahini:
Je merge car j'iamerais tester si ça fixe les batchs sur les instances

PR #1067, opened 2024-09-25
Allow the process manager to configure log files count and sizes

tahini:
* Use named objects instead of ordinal parameters for the numerous function parameters. This makes a call more readable, make it easier to fine-tune specific parameters while keeping the default values for others, and will make adding yet more parameters simpler.

* Let `ProcessManager` support log file count and size as extra parameter

* Let `TrRoutingProcessManager` support log file count and size as extra parameter. `startBatch` now also supports the `debug` option.

* Add tests for the new functionalities

### DIFF
--- a/examples/config.js
+++ b/examples/config.js
@@ -25,12 +25,16 @@ module.exports = {
     //         single: {
     //           port: 4000,
     //           // Enable caching of connections for all scenarios in trRouting. Will use more memory
-    //           cacheAllScenarios: false
+    //           cacheAllScenarios: false,
+    //           debug: false,
+    //           logs: { maxFileSizeKB: 5120, nbFiles: 3 }
     //         },
     //         batch: {
     //           port: 14000,
     //           // Enable caching of connections for all scenarios in trRouting. Will use more memory and may not be necessary for batch calculations as currently there's only one scenario per task
-    //           cacheAllScenarios: false
+    //           cacheAllScenarios: false,
+    //           debug: false,
+    //           logs: { maxFileSizeKB: 5120, nbFiles: 3 }
     //         }
     //       }
     //     }
--- a/packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
@@ -139,7 +139,9 @@ const start = function (parameters = {} as any): Promise<any> {
         const commandArgs = getOsrmRoutedStartArgs(osrmDirectoryPath, mode, port);
         const waitString = 'running and waiting for requests';
 
-        resolve(ProcessManager.startProcess(serviceName, tagName, command, commandArgs, waitString, true));
+        resolve(
+            ProcessManager.startProcess({ serviceName, tagName, command, commandArgs, waitString, useShell: true })
+        );
     });
 };
 
@@ -167,7 +169,15 @@ const restart = function (parameters): Promise<any> {
         const waitString = 'running and waiting for requests';
 
         resolve(
-            ProcessManager.startProcess(serviceName, tagName, command, commandArgs, waitString, true, undefined, true)
+            ProcessManager.startProcess({
+                serviceName,
+                tagName,
+                command,
+                commandArgs,
+                waitString,
+                useShell: true,
+                attemptRestart: true
+            })
         );
     });
 };
--- a/packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
@@ -60,7 +60,7 @@ const getLoggerName = function (tagName: string, serviceName: string) {
     return `${tagName}Logger_${serviceName}_`;
 };
 
-const setupLogger = function (tagName: string, serviceName: string) {
+const setupLogger = function ({ tagName, serviceName }: { tagName: string; serviceName: string }) {
     const loggerName = getLoggerName(tagName, serviceName);
     serviceLocator.addService(
         loggerName,
@@ -80,16 +80,25 @@ const setupLogger = function (tagName: string, serviceName: string) {
     return loggerName;
 };
 
-const startProcess = async function (
-    serviceName: string,
-    tagName: string,
-    command: string,
-    commandArgs: any,
-    waitString: string,
-    useShell: boolean,
-    cwd?: string,
+const startProcess = async function ({
+    serviceName,
+    tagName,
+    command,
+    commandArgs,
+    waitString,
+    useShell,
+    cwd,
     attemptRestart = false
-): Promise<any> {
+}: {
+    serviceName: string;
+    tagName: string;
+    command: string;
+    commandArgs: any;
+    waitString: string;
+    useShell: boolean;
+    cwd?: string;
+    attemptRestart?: boolean;
+}): Promise<any> {
     return new Promise((resolve) => {
         // Check if we already a process running for that serviceName
         // TODO Should be an await instead of then but need to rework the promise to make it work
@@ -99,7 +108,7 @@ const startProcess = async function (
             const doStart = function () {
                 // If we reached this point, we don't have a process running or it was stopped
 
-                const loggerName = setupLogger(tagName, serviceName);
+                const loggerName = setupLogger({ tagName, serviceName });
 
                 const spawnParams = {
                     shell: useShell,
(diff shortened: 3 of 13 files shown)


samuel-duhaime (approved):
Good.

tahini:
I added the possibility to configure the log files and whether to run in debug in the config file by default, with always the possibility to override in individual calls, but we don't support it yet.

@greenscientist fwiw we can now allow more log files and of larger sizes in our instances and we could support running in debug in some instances if required, batch and not batch separately.

greenscientist:
Kinda wondering if we should just strip all this out and let the standard OS tools handle log management.

greenscientist:
@tahini, as-tu regardé si il existait des modules standard pour gérer tout ça ?

tahini:
> @tahini, as-tu regardé si il existait des modules standard pour gérer tout ça ?

C'est ce que winston fait, on peut envoyer les logs où on veut. Ceci a l'avantage que qqn qui fait rouler Transition peut contrôler le logging. Si on laisse l'OS faire, il faudrait documenter la procédure ou configurer un autre outil séparément.

packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
```diff
@@ -86,18 +97,23 @@ const startProcess = async function ({
...
     serviceName: string;
     tagName: string;
     command: string;
     commandArgs: any;
     waitString: string;
-    useShell: boolean;
+    useShell?: boolean;
     cwd?: string;
     attemptRestart?: boolean;
+    logFiles?: {
```
greenscientist, line 113: Do we realely need a struct in a struct? This make calling the function a bit more complex with useful reason.

packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
```diff
@@ -60,7 +60,17 @@ const getLoggerName = function (tagName: string, serviceName: string) {
     return `${tagName}Logger_${serviceName}_`;
 };
 
-const setupLogger = function ({ tagName, serviceName }: { tagName: string; serviceName: string }) {
+const setupLogger = function ({
+    tagName,
+    serviceName,
+    nbFiles = 1,
+    maxFileSizeKb = 1000
```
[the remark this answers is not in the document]
greenscientist: Also, a bit unreleated, but 1 meg is probably a bit small to be the default value. Thoughts?

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -14,6 +14,13 @@ import config from '../../config/server.config';
 // Set 2 threads as default, just in case we have more than one request being handled
 const DEFAULT_THREAD_COUNT = 2;
 
+type StartParameters = {
```
greenscientist, line 17: Please add a command to mention what is that type and where it will be used.

greenscientist: If it's of general use it should be in the ProcessManager file. Otherwise it whould be named TrRouting something...

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -22,10 +29,17 @@ const startTrRoutingProcess = async (
...
     attemptRestart = false,
     threadCount = 1,
-    parameters: { debug?: boolean; cacheDirectoryPath?: string; cacheAllScenarios?: boolean } = {
-        debug: false,
-        cacheDirectoryPath: undefined,
-        cacheAllScenarios: false // Flag to enable the trRouting connection cache for all scenario
+    {
+        debug = false,
+        cacheDirectoryPath,
+        // Flag to enable the trRouting connection cache for all scenario
```
greenscientist, line 35: Is it possible to keep the comment after the param to make it clear to which entry it's related.

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -81,15 +96,16 @@ const startTrRoutingProcess = async (
...
     const port = parameters.port || Preferences.get('trRouting.port');
     const cacheAllScenarios =
         config.trRoutingCacheAllScenarios === undefined ? false : config.trRoutingCacheAllScenarios;
 
     const params: Parameters<typeof startTrRoutingProcess>[3] = {
         debug: parameters.debug,
         cacheDirectoryPath: parameters.cacheDirectoryPath,
-        cacheAllScenarios: cacheAllScenarios
+        cacheAllScenarios: cacheAllScenarios,
+        logFiles: parameters.logFiles
```
greenscientist: Not sure it's a good idea of expose the logFiles setting externally. We should know what we expect here.

examples/config.js
```diff
@@ -17,6 +17,19 @@ module.exports = {
 
     // Enable caching of connections for all scenarios in trRouting. Will use more memory
     trRoutingCacheAllScenarios: false,
+    // Extra options for running trRouting
```
greenscientist, line 20: We if add a config block for trRouting, then we need to move the cacheAllFlag inside of it.

greenscientist: We should also move the port number there.

greenscientist (changes requested):

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -139,7 +139,9 @@ const start = function (parameters = {} as any): Promise<any> {
         const commandArgs = getOsrmRoutedStartArgs(osrmDirectoryPath, mode, port);
         const waitString = 'running and waiting for requests';
 
-        resolve(ProcessManager.startProcess(serviceName, tagName, command, commandArgs, waitString, true));
+        resolve(
+            ProcessManager.startProcess({ serviceName, tagName, command, commandArgs, waitString, useShell: true })
```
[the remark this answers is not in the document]
tahini: Yes it's magic, we never do serviceName: serviceName in javascript

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
tahini: Another option for this PR would be to simply put the winston config in the config.js file (maybe not exactly the config, but something tha twill allow us to easily build that config). That would avoid all the parameter passing, each service can get its log configuration and we won't have to add too much when we want to support something else than log files.

packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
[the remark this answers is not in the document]
tahini: Defaults to 3 files of 5MB each now, for the processManager and specifically for trRouting as well

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
tahini: Done, this is for the TrRotuingProcessManager only

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
tahini: I updated so log files are only configurable through the config, there's no use case for individually setting it.

examples/config.js
tahini: Done in previous PR

tahini:
@greenscientist I rebase and updated to address all your comments. With this, next time we need to investigate some trRouting related errors, we'll be ready to configure debug and logging in our instances.

packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
```diff
@@ -12,6 +12,8 @@ import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import ProcessUtils from './ProcessUtils';
 
 const PID_DIRECTORY = 'pids';
+const DEFAULT_LOG_FILE_COUNT = 3; // 3 files
```
greenscientist, line 15: Shouldn't those be removed in the last commit, now that we have default in the config file???

greenscientist (commented):
One last question.

packages/chaire-lib-backend/src/utils/processManagers/ProcessManager.ts
greenscientist: Can we combine those constant somehow? Can the trRouting default config information use those constant ?

tahini: Moved to the config file

greenscientist (approved):

PR #1072, opened 2024-10-03
trRouting: Move port and cacheAllScenarios to instance specific config

tahini:
This adds the `trRouting` property to the config type. This property contains data for the `single` instance for individual calculations (UI and public API) and `batch` instance (for tasks). For each instance, the `port` and `cacheAllScenarios` properties can be set.

The configuration objects replace the `trRouting` preference that contained the `port` and `batchPortStart` and the current `trRoutingCacheAllScenarios` configuration option.

Port management when starting or querying trRouting is updated to match the new configuration options.

### DIFF
--- a/packages/chaire-lib-backend/src/config/ServerConfig.ts
+++ b/packages/chaire-lib-backend/src/config/ServerConfig.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import config, { TrRoutingConfig } from './server.config';
+import { RoutingOrTransitMode } from 'chaire-lib-common/lib/config/routingModes';
+
+/**
+ * Helper class to manage accesses to server configuration
+ */
+class ServerConfig {
+    constructor() {
+        // Nothing to do
+        // TODO Currently, the `server.config` file sets the default value,
+        // because there is not that many we want to set, but as we add more,
+        // this class should be responsible to set them
+    }
+
+    /**
+     * Get the routing mode configuration
+     * @param mode The mode to get the configuration for
+     * @returns the project configuration, or undefined if the mode is not configured
+     */
+    getRoutingModeConfig = (mode: RoutingOrTransitMode) => {
+        if (config.routing[mode] === undefined) {
+            console.error('Configuration for routing mode %s not found', mode);
+        }
+        return config.routing[mode];
+    };
+
+    /**
+     * Get the routing engine configuration for a given mode
+     *
+     * TODO See if we can type the return value better as each engine should
+     * have their own config type
+     *
+     * @param mode The mode to get the configuration for
+     * @param engine The engine to use
+     * @returns
+     */
+    getRoutingEngineConfigForMode = (mode: RoutingOrTransitMode, engine: string) => {
+        if (config.routing[mode] === undefined) {
+            console.error('Configuration for routing mode %s not found', mode);
+        }
+        if (config.routing[mode] !== undefined && config.routing[mode]!.engines[engine] === undefined) {
+            console.error('Configuration for engine %s for routing mode %s not found', engine, mode);
+        }
+        return config.routing[mode]?.engines[engine];
+    };
+
+    getTrRoutingConfig = (instance: 'single' | 'batch'): TrRoutingConfig => {
+        const engineConfiguration = this.getRoutingEngineConfigForMode('transit', 'trRouting');
+        // The configuration should be defined, throw an error if it is not, we have a problem
+        if (engineConfiguration === undefined) {
+            throw new Error('trRouting configuration not found');
+        }
+        return engineConfiguration[instance] || { port: 4000, cacheAllScenarios: false };
+    };
+}
+
+const instance = new ServerConfig();
+export default instance;
--- a/packages/chaire-lib-backend/src/config/server.config.ts
+++ b/packages/chaire-lib-backend/src/config/server.config.ts
@@ -6,9 +6,78 @@
  */
 import path from 'path';
 import os from 'os';
+import _merge from 'lodash/merge';
 
-import config, { setProjectConfiguration } from 'chaire-lib-common/lib/config/shared/project.config';
+import config, {
+    ProjectConfiguration,
+    setProjectConfiguration as setProjectConfigurationCommon
+} from 'chaire-lib-common/lib/config/shared/project.config';
 import fs from 'fs';
+import { RoutingMode } from 'chaire-lib-common/lib/config/routingModes';
+
+export type TrRoutingConfig = {
+    /**
+     * Port to use for this trRouting instance
+     */
+    port: number;
+    /**
+     * If set to `true`, enable caching of connections for all scenarios in
+     * trRouting. Will use more memory
+     */
+    cacheAllScenarios: boolean;
+    // FIXME Do we need to configure a host here? If so, we need to properly
+    // support it in both batch and single calculation
+};
+
+// TODO Some project config option depend on the application, so should not be
+// typed in chaire-lib. Each app (transition , evolution, etc) should add types
+// to the config and should have a project.config file which imports this one
+// for common options.
+export type ServerSideProjectConfiguration = {
+    /** @deprecated Use `routing.transit.engines.trRouting.*.cacheAllScenarios instead */
+    trRoutingCacheAllScenarios?: boolean;
+
+    /**
+     * Configuration for the various engines used for the routing modes
+     *
+     * TODO Actually use this configuration in the various calculators, once we
+     * support more than a single engine per mode
+     */
+    routing: {
+        transit: {
+            defaultEngine: string;
+            engines: {
+                /**
+                 * Configuration for the trRouting engine
+                 */
+                trRouting?: {
+                    /**
+                     * Configuration for the single calculation trRouting service
+                     */
+                    single: TrRoutingConfig;
+                    /**
+                     * Configuration for the batch calculation trRouting service
+                     */
+                    batch: TrRoutingConfig;
+                };
+            };
+        };
+    } & {
+        [key in RoutingMode]?: {
+            defaultEngine: string;
+            engines: {
+                // TODO Type and intialize these configuration for routing modes
+                [key: string]: any;
+            };
+        };
+    };
+
+    /**
+     * The number of days that a token is valid for. After this time, the token
+     * will expire
+     */
+    tokenLifespanDays: number;
+};
 
 const etcConfigFile = '/etc/transition/config.js';
 const homeDirConfigFileRelative = '.config/transition/config.js';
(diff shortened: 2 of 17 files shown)


tahini:
This PR is pre #1067 

@GabrielBruno24 Voici un exemple d'ajout d'option de config, avec valeurs par défaut et son utilisation ailleurs, pour quand tu auras à ajouter le routage manuel.

greenscientist:
Est-ce qu'on peut avoir un support pour avoir   trRoutingCacheAllScenarios: false, en mode deprecated (avec un warning message), le temps qu'on puisse changer l'infra ?

tahini:
> Est-ce qu'on peut avoir un support pour avoir trRoutingCacheAllScenarios: false, en mode deprecated (avec un warning message), le temps qu'on puisse changer l'infra ?

J'ai déjà un PR prête pour l'infra... Je mets le deprecated quand même?

tahini:
> > Est-ce qu'on peut avoir un support pour avoir trRoutingCacheAllScenarios: false, en mode deprecated (avec un warning message), le temps qu'on puisse changer l'infra ?
> 
> J'ai déjà un PR prête pour l'infra... Je mets le deprecated quand même?

Bon ok, m'a l'faire, tout d'in coup que qqn d'autre utilise ça en qq part...

greenscientist:
> > Est-ce qu'on peut avoir un support pour avoir trRoutingCacheAllScenarios: false, en mode deprecated (avec un warning message), le temps qu'on puisse changer l'infra ?
> 
> J'ai déjà un PR prête pour l'infra... Je mets le deprecated quand même?

Ca reste quand meme un peu difficile synchroniser tous les changements.

GabrielBruno24 (approved):

tahini:
c'est fait, c'est quand même moins long que s'obstiner pour pas le faire :D

greenscientist:
bikeshedding a little bit here: 
Should we put those entries into a "RoutingEngine" section ? We could then select which routing engine we'd want and then specfiy it's parameters directly?
Kinda ties in with Gabriel task

tahini:
> We could then select which routing engine we'd want and then specfiy it's parameters directly?

What do you mean by that? How/where would you specify its parameters? The RoutingEngine section would be `{[RoutingEngineName: string]: ParameterType}`? 

Besides, each routing engine (trRouting, osrm) has different parameter types, so I'm not 100% sure this would simplify. In this case, the config type for the trRouting would just move to RoutingEngine instead of the root. If it's just a matter of organization, it could go in RoutingEngine, if you insist

greenscientist:
The idea is to organise things coherently. We will need to add a switch to specify manual routing. We probably need a TransitRouting section and a more "generic routing" section, I think we have a different behavior for both of these. 

Maybe we have a routing section: 

Routing: 
    Transit: 
          Single: 
                TrRouting: 
                       TrRoutingParam1
                       TrRoutingParam2
                       TrRoutingParam3
          Batch
                TrRouting:
                       ....

    "simple":      //Don't know if we put a section there or just put Walking directly
         Walking: 
              OSRM: 
                   OSRMParameter1 
                   OSRMParameter2


Then with something manuel:
Routing: 
    ....
   Walking:
         Manual:
               ManualRoutingPARAM1
               ....

Thoughts?

tahini:
If we want to organize things coherently, it would make sense that they match the calculation requests, where the key is the `RoutingOrTransitMode` type, so first level would be 'transit', 'walking', etc. Next level would be an identifier of the engine and its properties as you propose. There could be a `defaultEngine` property also if there are many, but the type, the engines could in an `engines` property. As for single vs batch, I don't know where in that hierarchy they should go! We have this concept with trRouting, but not other engines, but maybe we will eventually...

Anyway, we may be overengineering and this approach will make it harder to set/propose default values, but anyway, currently we have all this in the defaultPreferences and administrators need to overwrite those default preferences in the config, so not much difference there, the defaultPreferences values would go in the default config.

So, here's how I see this, if we decide to go this way, the driving would use manual routing, while the walking would use osrm: 
```
Routing: 
    Transit:
        defaultEngine: 'trRouting',
        engines:
             trRouting:
                Single: // If this applies only to trRouting, put it at this level
                       TrRoutingParam1
                       TrRoutingParam2
                       TrRoutingParam3
                 Batch:
                       ....

     Walking: 
        defaultEngine: 'osrm',
        engines: 
              osrm: 
                   OSRMParameter1 
                   OSRMParameter2
    Driving:
        defaultEngine: 'manual',
        engines:
            manual:
                   ManualRoutingParameter1
```

tahini:
(à noter que ma responsabilité dans tout ça se limite à la portion trRouting, les autres, si on va là, seront pour @GabrielBruno24 )

greenscientist:
Ok, I like your counter proposal, I think it's quite flexible, we can eventually put "global" setting at the mode level, common to all engines. 
can you rework your PR to match this structure? Obviously only the TrRouting part. Will open an issue to migrate the OSRM config to that structure.

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -81,10 +80,28 @@ const startTrRoutingProcess = async (
     return processStatus;
 };
 
+const getCacheAllScenarios = () => {
```
greenscientist, line 83: Could we make a class that manage access to the config (a TrRoutingConfig or something like that.) Eventually, it could be merge to a more global config thing, but to group all config access in one places. 
Could all be static function. 

Optionaly, for the lazy path, Can we name the function something with Config in the name.

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -81,10 +80,28 @@ const startTrRoutingProcess = async (
...
+    // FIXME Should be simply `return
+    // config.trRouting?.single?.cacheAllScenarios === undefined ? false :
+    // config.trRouting.single.cacheAllScenarios;`, but we need to handle
+    // deprecation of `trRoutingCacheAllScenarios` value. Deprecated config
+    // option has precedence as it means it was already set by the user. The
```
greenscientist, line 88: It's an edge case,but if the user have set both the old config setting and the new one I think that the new one should have precedence.

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
```diff
@@ -31,9 +39,12 @@ startProcessMock.mockImplementation(async (serviceName: string,
     service: tagName,
     name: serviceName
 }));
+let defaultConfig = _cloneDeep(config);
```
greenscientist, line 42: Maybe defaultTestConfig ?    
It kind of confuse me. 
Maybe a comment about why we need this cloneDeep. 
Should this be done in the beforeEach instead?

packages/chaire-lib-backend/src/utils/trRouting/TrRoutingServiceBackend.ts
```diff
@@ -8,9 +8,8 @@ import _get from 'lodash/get';
 //TODO replace this fetch-retry library with one compatible with TS
 const fetch = require('@zeit/fetch-retry')(require('node-fetch'));
 
-import Preferences from 'chaire-lib-common/lib/config/Preferences';
+import config from '../../config/server.config';
```
greenscientist, line 11: That's nice :)

packages/chaire-lib-backend/src/utils/trRouting/TrRoutingServiceBackend.ts
```diff
@@ -191,9 +190,11 @@ class TrRoutingServiceBackend {
         port: string | number | undefined,
         customRequestPath: string
     ): string {
-        const trRoutingConfig = Preferences.get('trRouting');
+        // FIXME We should not get the port from here, as we assume here it is for the single instance and not the batch.
+        const trRoutingConfig = config.trRouting?.single || { port: 4000 };
```
greenscientist, line 194: We really need a variable DEFAULT_TRROUTING_PORT here instead of the 4000

greenscientist (commented):
A few comment on the code.

packages/chaire-lib-backend/src/config/ServerConfig.ts
```diff
@@ -0,0 +1,60 @@
...
+        return config.routing[mode];
+    };
+
+    /**
+     * Get the routing engine configuration for a given mode
+     * @param mode The mode to get the configuration for
+     * @param engine The engine to use
+     * @returns
+     */
+    getRoutingEngineConfigForMode = (mode: RoutingOrTransitMode, engine: string) => {
```
greenscientist, line 39: I feel there's something more typed (and later more generic for the engine management), but we can think about this later

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -6,9 +6,78 @@
...
+    /**
+     * If set to `true`, enable caching of connections for all scenarios in
+     * trRouting. Will use more memory
+     */
+    cacheAllScenarios: boolean;
+    // FIXME Do we need to configure a host here? If so, we need to properly
+    // support it in both batch and single calculation
+};
+
+// TODO Some project config option depend on the application, so should not be
```
greenscientist, line 32: So the TODO means the routing section should be moved to the transition project and not have it here right ?  
Can we make sure we have an issue ? (Maybe add exemple of section in the comment)

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -47,11 +116,53 @@ const getConfigFilePath = (): string => {
 const configFileNormalized = getConfigFilePath();
 
 // Initialize default server side configuration
-setProjectConfiguration({
+setProjectConfigurationCommon<ServerSideProjectConfiguration>({
     // Initialize runtime directory at the root of the repo
-    projectDirectory: path.normalize(`${__dirname}/../../../../runtime/`)
+    projectDirectory: path.normalize(`${__dirname}/../../../../runtime/`),
```
greenscientist, line 121: Can we have a comment that says that these define the default values.

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -81,10 +81,11 @@ const startTrRoutingProcess = async (
     return processStatus;
 };
 
+const getCacheAllScenariosFromConfig = () => serverConfig.getTrRoutingConfig('single').cacheAllScenarios;
```
greenscientist, line 84: Do we still need this function?

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
greenscientist: Comment still valid.

greenscientist (commented):
Couple more small things

packages/chaire-lib-backend/src/config/server.config.ts
tahini: This TODO is copy-pasted from the chaire-lib-common's config file.

In the case we're considering, all routing engines are in chaire-lib already, so their configs need to be here too, otherwise TrRoutingProcessManager cannot access them if they are in Transition.

packages/chaire-lib-backend/src/config/server.config.ts
tahini: line 118 is the comment

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
tahini: yeah, the function contains the 'single' instance definition, so it gets less error-prone and we are sure that all function that use it will use the exact same config. But it's debatable. It's less necessary now.

packages/chaire-lib-backend/src/config/ServerConfig.ts
tahini: Added a TODO

packages/chaire-lib-backend/src/utils/processManagers/__tests__/TrRoutingProcessManager.test.ts
tahini: Added

greenscientist (approved):
I just have a small issue with the name of getCacheAllScenariosFromConfig
I'll approve and let you decide what to this with it.

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
greenscientist: Worst case, a comment with a WARNING to not use with the batch path

PR #1086, opened 2024-10-21
schedules: Use numeric ids for schedules tables primary keys

tahini:
Change the schedule, schedule_periods and schedule_period_trips's
primary id columns from uuid to auto-increment numeric IDs. The old
uuid is kept is a column called `uuid`. The foreign keys are modified
to use the numeric key instead. The `schedule_id` column of the trips'
table is removed, as there is already a schedule_period_id foreign which
has a foreign key to the schedule_id. Having both in trips open the door
to data integrity issues where the period could not belong to the same
schedule as the trip.

auto-increment numeric IDs are more performant than uuids as keys. It
will also allow to more easily duplicate multiple elements in a table
with predictable sort order, that uuids do not allow.

This commit only changes the database table queries. The Schedule class
still use the uuid as `id` as this would require more refactoring of the
base classes, history tracker, forms and client/server messaging, that
will be done in later commits.

The numeric IDs are kept in the `integer_id` fields of the Schedule. The
foreign key are changed to map to the numeric ID.

The impact is "minimal" as the schedules are used mostly in calculations
by trRouting and, in the frontend, are [re-]loaded on demand when
changes occur on a specific line. There is no collection of this object
type where the change of ID could cause issues.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
+++ b/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
@@ -40,9 +40,19 @@ export const stringifyDataSourceIds = function (dataSourceIds: string[]): string
  * @param knex The database configuration object
  * @param tableName The name of the table on which to execute the operation
  * @param id The ID of the object
+ * @param {Object} options Additional options parameter.
+ * @param {Knex.Transaction} [options.transaction] - transaction that this query
+ * is part of
  * @returns Whether an object with the given ID exists in the table
  */
-export const exists = async (knex: Knex, tableName: string, id: string): Promise<boolean> => {
+export const exists = async (
+    knex: Knex,
+    tableName: string,
+    id: string,
+    options: {
+        transaction?: Knex.Transaction;
+    } = {}
+): Promise<boolean> => {
     if (!uuidValidate(id)) {
         throw new TrError(
             `Cannot verify if the object exists in ${tableName} because the required parameter id is missing, blank or not a valid uuid`,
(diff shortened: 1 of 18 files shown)


tahini:
Work in ongoing to update the base classes/forms/communication, etc for numeric IDs instead of UUID. While doing this, some cleanup and simplifications will take place, so it is not a trivial work.

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -42,16 +42,27 @@ export const stringifyDataSourceIds = function (dataSourceIds: string[]): string
...
-    if (!uuidValidate(id)) {
+export const exists = async (
+    knex: Knex,
+    tableName: string,
+    id: string | number,
+    options: {
+        transaction?: Knex.Transaction;
+    } = {}
+): Promise<boolean> => {
+    if (typeof id === 'string' && !uuidValidate(id)) {
```
kaligrafy, line 53: Can you add a comment saying that we now accept both uuids (string) or id (integer)?

kaligrafy (changes requested):

packages/chaire-lib-backend/src/models/db/default.db.queries.ts
tahini: jsdoc was updated, also with the new options parameters

kaligrafy (approved):

greenscientist:
@tahini What's the status of this?

tahini:
It's ready and works, but it is just a step of a larger rewrite where most or all objects will also have numeric IDs and will require to re-factor the forms save and object interactions, as with numeric IDs, a new object will not have an ID until saved, unlike now with the UUIDs.

tldr; I was waiting to be further in the refactoring to make sure everything is ok here. But then again, we could merge and fix whatever will need to be fixed later.

Issue #1084, opened 2024-10-11
Migrate OSRM settings from Preferences to Config


greenscientist:
As issues in the PR #1072 , we can use the proposed structure for OSRM.

PR #1087, opened 2024-10-23
Migrate OSRM settings

GabrielBruno24:
Migrate OSRM settings from the defaultPreferences.config.ts file in chaire-lib-common to server.config,ts file in chaire-lib-backend
Also adds some todos as a reminder to rework some settings that can't be moved at the moment.
Fix: #1084

### DIFF
--- a/examples/config.js
+++ b/examples/config.js
@@ -38,7 +38,79 @@ module.exports = {
     //         }
     //       }
     //     }
-    //   }
+    //   },
+    //   // Configuration for simple routing modes, using OSRM-like engines 
+    //   driving: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { 
+    //         port: 7000, // Port used to access OSRM, either locally or remotely
+    //         host: null, // If set to null, localhost will be used. Ignored if autoStart set to true
+    //         autoStart: true, // If true, a local instance of OSRM will be started
+    //         enabled: true // If true, this mode will be configured, otherwise will be left out
+    //       }
+    //     }
+    //   },
+    //   cycling: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 8000, host: null, autoStart: true, enabled: true }
+    //     }
+    //   },
+    //   walking: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 5001, host: null, autoStart: true, enabled: true }
+    //     }
+    //   },
+    //   bus_suburb: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 7110, host: null, autoStart: true, enabled: true }
+    //     }
+    //   },
+    //   bus_urban: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 7120, host: null, autoStart: true, enabled: true }
+    //     }
+    //   },
+    //   rail: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9000, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
+    //   tram: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9100, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
+    //   tram_train: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9200, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
+    //   metro: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9300, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
+    //   monorail: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9400, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
+    //   cable_car: {
+    //     defaultEngine: 'osrmRouting',
+    //     engines: {
+    //       osrmRouting: { port: 9500, host: null, autoStart: false, enabled: false }
+    //     }
+    //   },
     // },
   
     mapDefaultCenter: {
@@ -78,77 +150,6 @@ module.exports = {
     },
   
     defaultPreferences: {
-        osrmRouting: {
-            modes: {
-                driving: {
-                    // !!! Be careful: use your own server, since you may be blocked on the osrm demo server after too many queries
-                    port     : 7000, // Port used to access OSRM, either locally or remotely
-                    host     : null, // If set to null, localhost will be used. Ignored if autoStart set to true
-                    autoStart: true, // If true, a local instance of OSRM will be started
-                    enabled  : true  // If true, this mode will be configured, otherwise will be left out 
-                },
-                cycling: {
-                    port     : 8000,
-                    host : null,
-                    autoStart: true,
-                    enabled  : true
-                },
-                walking: {
-                    port     : 5001,
-                    host : null,
-                    autoStart: true,
-                    enabled  : true
-                },
-                bus_suburb: {
-                    port     : 7110,
-                    host : null,
-                    autoStart: true,
-                    enabled  : true
-                },
-                bus_urban: {
-                    port     : 7120,
-                    host : null,
-                    autoStart: true,
-                    enabled  : true
-                },
-                rail: {
-                    port     : 9000,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                },
-                tram: {
-                    port     : 9100,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                },
-                tram_train: {
-                    port     : 9200,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                },
-                metro: {
-                    port     : 9300,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                },
-                monorail: {
-                    port     : 9400,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                },
-                cable_car: {
-                    port     : 9500,
-                    host : null,
-                    autoStart: false,
-                    enabled  : false
-                }
-            }
-      },
       transit: {
         routing: {
           batch: {
--- a/packages/chaire-lib-backend/src/config/__tests__/ServerConfig.test.ts
+++ b/packages/chaire-lib-backend/src/config/__tests__/ServerConfig.test.ts
@@ -37,7 +37,7 @@ describe('get routing mode and engine configs', () => {
     });
 
     test('getRoutingModeConfig, mode does not exist', () => {
-        expect(ServerConfig.getRoutingModeConfig('walking')).toBeUndefined();
+        expect(ServerConfig.getRoutingModeConfig('walking_way_data_as_name')).toBeUndefined();
     });
 
     test('getRoutingEngineConfigForMode, mode and engine exist', () => {
@@ -52,7 +52,29 @@ describe('get routing mode and engine configs', () => {
     });
 
     test('getRoutingEngineConfigForMode, mode does not exist', () => {
-        expect(ServerConfig.getRoutingEngineConfigForMode('walking', 'osrm')).toBeUndefined();
+        expect(ServerConfig.getRoutingEngineConfigForMode('walking_way_data_as_name', 'osrmRouting')).toBeUndefined();
     });
 
+    test('getAllModesForEngine, at least one mode uses engine', () => {
+        expect(ServerConfig.getAllModesForEngine('trRouting')).toEqual(['transit']);
+        expect(ServerConfig.getAllModesForEngine('osrmRouting')).toEqual([
+            'driving',
+            'driving_congestion',
+            'cycling',
+            'walking',
+            'bus_suburb',
+            'bus_urban',
+            'bus_congestion',
+            'rail',
+            'tram',
+            'tram_train',
+            'metro',
+            'monorail',
+            'cable_car'
+        ]);
+    });
+
+    test('getAllModesForEngine, no mode uses engine', () => {
+        expect(ServerConfig.getAllModesForEngine('valhalla')).toEqual([]);
+    });
 });
--- a/packages/chaire-lib-backend/src/config/server.config.ts
+++ b/packages/chaire-lib-backend/src/config/server.config.ts
@@ -50,6 +50,26 @@ export type TrRoutingConfig = {
     // support it in both batch and single calculation
 };
 
+export type OsrmRoutingConfig = {
+    /**
+     * Port used to access OSRM, either locally or remotely.
+     */
+    port: number | null;
+    /**
+     * If set to null, localhost will be used. Ignored if autoStart set to true.
+     * Do not prefix with the protocol (ex: 'some.osrm.site.com')
+     */
+    host: string | null;
+    /**
+     * If true, a local instance of OSRM will be started when launching transition (client).
+     */
+    autoStart: boolean;
+    /**
+     * If true, this mode will be configured, otherwise it will be left out. Usable when routing paths.
+     */
+    enabled: boolean;
+};
+
 // TODO Some project config option depend on the application, so should not be
 // typed in chaire-lib. Each app (transition , evolution, etc) should add types
 // to the config and should have a project.config file which imports this one
@@ -87,8 +107,7 @@ export type ServerSideProjectConfiguration = {
         [key in RoutingMode]?: {
             defaultEngine: string;
             engines: {
-                // TODO Type and intialize these configuration for routing modes
-                [key: string]: any;
+                osrmRouting?: OsrmRoutingConfig;
             };
         };
     };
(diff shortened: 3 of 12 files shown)


packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -4,29 +4,31 @@
...
+import { setProjectConfiguration } from '../../config/server.config';
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 
 const getServiceName = function (mode: RoutingMode = 'walking', port: number | null = 5000) {
     return `osrmMode__${mode}__port${port}`;
 };
 
 const routingModeIsAvailable = async function (routingMode: RoutingMode): Promise<boolean> {
-    const osrmServerPrefs = Preferences.get('osrmRouting', {});
+    const osrmModePref = ServerConfig.getRoutingEngineConfigForMode(routingMode, 'osrmRouting');
```
tahini, line 21: Rename to `osrmModeConfig`

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -61,28 +63,34 @@ const availableRoutingModes = async function (): Promise<RoutingMode[]> {
...
-    const updatedProperties = {};
-    updatedProperties['modes'] = {};
-    updatedProperties['modes'][mode] = { enabled: false };
-    Preferences.updateServerPrefs({ osrmRouting: updatedProperties });
+    setProjectConfiguration({
+        routing: {
+            [mode]: {
+                defaultEngine: 'osrmRouting',
+                engines: {
+                    osrmRouting: {enabled: false}
```
tahini, line 71: Also, please run `yarn format` before committing as it will change the formatting

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -129,7 +137,7 @@ function getOsrmRoutedStartArgs(osrmDirectory: string, mode: string, port: strin
 //TODO set type for Promise return (in all the file)
 const start = function (parameters = {} as any): Promise<any> {
     const mode = parameters.mode || 'walking';
-    const port = parameters.port || Preferences.get(`osrmRouting.${mode}.port`);
+    const port = parameters.port || ServerConfig.getRoutingEngineConfigForMode(mode, 'osrmRouting').port;
```
tahini, line 140: Just wondering if we need this, maybe add a TODO to force receiving the mode and port (because the caller will have checked it already) and not initialize to config here

packages/chaire-lib-common/src/services/routing/OSRMRoutingService.ts
```diff
@@ -88,6 +88,7 @@ export default class OSRMRoutingService extends RoutingService.default {
             const origin = waypoints[0];
             const destination = waypoints[waypoints.length - 1];
 
+            // TODO: Don't use osrmRouting params in chaire-lib-common
```
tahini, line 91: Here the TODO should be more Determine if this value should be a config or a preference

tahini (changes requested):
Nothing major

packages/chaire-lib-backend/src/utils/processManagers/__tests__/OSRMProcessManager.test.ts
```diff
@@ -188,13 +202,16 @@ describe('OSRM Service Manager', function() {
     const prepareResult = await OSRMServicePreparation.prepare(existingOsmFilePath, ['walking']);
     expect(prepareResult.status).toBe('prepared');
 
+    const osrmModes = ServerConfig.getAllModesForEngine("osrmRouting");
```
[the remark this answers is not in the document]
GabrielBruno24: This function is already tested in another test. It also just returns all modes in the config that have the osrmROuting engine, regardless of whether they are disabled or not, so nothing done in this test should be capable of changing the excepted result

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
GabrielBruno24: I ran the command but it didn't change this part specifically

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -4,29 +4,27 @@
...
 };
 
 const routingModeIsAvailable = async function (routingMode: RoutingMode): Promise<boolean> {
-    const osrmServerPrefs = Preferences.get('osrmRouting', {});
-    if (
-        osrmServerPrefs.modes[routingMode] &&
-        osrmServerPrefs.modes[routingMode].enabled === true &&
-        osrmServerPrefs.modes[routingMode].autoStart !== true
-    ) {
+    const osrmModeConfig = ServerConfig.getRoutingEngineConfigForMode(routingMode, 'osrmRouting');
```
tahini, line 21: J'ai oublié de le mentionner au premier review, mais il faudrait que tout reste compatible avec le fichier de config actuel (par exemple, la configuration de nos instances ne sera probablement pas mise à jour exactement en même temps que le prochain update de transition une fois cette PR mergée). Regarde ce qui a été fait pour la config `trRoutingCacheAllScenarios` dans `server.config.ts`, in the `setProjectConfiguration` function, if the old way of configuring is set, but now the new way, it is merged to make sure old configs still work. You can do something similar for the osrm routing's default preferences

GabrielBruno24: I added it to the setProjectConfiguration. I didn't push anything related to that to the tests since I didn't want to add deprecated stuff to any config file but I did test it locally and it works.

tahini (changes requested):
je ne suis pas certaine que l'ancienne config fonctionnerait encore puisqu'il manque le defaultPreferences. Pour tester, voir si tout fonctionne bien avec le fichier de config comme il était avant.

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -191,6 +353,35 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
             }
         });
     }
+
+    // If osrmRouting is configured using the old Preferences format and not the new one in server.config, set it to the new format
+    // FIXME Remove once the old osrmRouting format is fully deprecated
+    const osrmModesInDeprecatedFormat = Preferences.get('osrmRouting').modes;
```
tahini, line 359: I think there might be a circular dependency as the preferences need the config to be initialized to get the default preferences from the config. I used my old config file and I don't get the warning and the process does not start on the right port.

I think here you need to get the defaultPreferences from the newConfig directly: `newConfig.defaultPreferences!.osrmRouting`.

tahini (changes requested):

tahini (approved):

tahini:
@kaligrafy @greenscientist voulez-vous aussi reviewer avant de merger?

kaligrafy (approved):

greenscientist:
> @kaligrafy @greenscientist voulez-vous aussi reviewer avant de merger?

Je jette un coup d'oeil tantot.

examples/config.js
```diff
@@ -18,27 +18,94 @@ module.exports = {
...
+    //                     debug: false,
+    //                     logs: { maxFileSizeKB: 5120, nbFiles: 3 }
+    //                 }
+    //             }
     //         }
-    //       }
-    //     }
-    //   }
+    //     },
+    //     // Configuration for the osrm modes
```
greenscientist, line 42: Maybe just change the comment here to say something like config for simple routing modes (using engines like OSRM)

examples/config.js
```diff
@@ -18,27 +18,94 @@ module.exports = {
...
+    //                     logs: { maxFileSizeKB: 5120, nbFiles: 3 }
+    //                 }
+    //             }
     //         }
-    //       }
-    //     }
-    //   }
+    //     },
+    //     // Configuration for the osrm modes 
+    //     driving: {
```
greenscientist, line 43: Can we add comments to the first entry to describe each fields, like we did for the old config.

packages/chaire-lib-backend/src/config/__tests__/ServerConfig.test.ts
```diff
@@ -52,7 +52,29 @@ describe('get routing mode and engine configs', () => {
     });
 
     test('getRoutingEngineConfigForMode, mode does not exist', () => {
-        expect(ServerConfig.getRoutingEngineConfigForMode('walking', 'osrm')).toBeUndefined();
+        expect(ServerConfig.getRoutingEngineConfigForMode('walking_way_data_as_name', 'osrmRouting')).toBeUndefined();
     });
 
+    test('getAllModesForEngine, engine exists', () => {
```
greenscientist, line 58: Just nit picking on the test name here. 

The test validate that the modes have a specific default engine, not necessary that they just exist. We might want to ajust the name or the test

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -50,6 +50,25 @@ export type TrRoutingConfig = {
     // support it in both batch and single calculation
 };
 
+export type OsrmRoutingConfig = {
+    /**
+     * Set to null when using remote osrm server
```
greenscientist, line 55: I think the comment are wrong here. 
We need the port if we use a remote OSRM

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -50,6 +50,25 @@ export type TrRoutingConfig = {
...
+     */
+    port: number | null;
+    /**
+     * Set to null when using remote osrm server
```
greenscientist, line 59: same

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -50,6 +50,25 @@ export type TrRoutingConfig = {
...
+     */
+    osrmPath: string | null;
```
greenscientist, line 61: Can we change this to hostname ?

Or have both an OSRMPath and hostname field ? (Or name it OsrmPathOrHostname)

packages/chaire-lib-common/src/config/__tests__/Preferences.test.ts
```diff
@@ -44,11 +44,13 @@ test('Test current match attributes', () => {
 
 test('Test initialize PreferencesClass with custom attributes', () => {
     const preferences = new PreferencesClass({ data: { foo: 'bar' }, osrmRouting: { directoryPrefix: 'newPrefix' } });
+    // TODO: Decide if this value should be a config or a preference
```
greenscientist, line 47: It should definitely be a config not a preference. 
What would be the case for a user to configure this ?

greenscientist (changes requested):
A few small comment and one bit one. 

My main issue is the use of OSRMPath in the internal structure instead of host. 
We should either have both of them, or even maybe have 2 different struct (A local OSRM could be a different engine that a remote OSRM.)

Or we name the variable PathOrHost

(I think the path is used when we want to do the autostart one so it's really a different use case and simply a hostname)

packages/chaire-lib-backend/src/config/server.config.ts
GabrielBruno24: I renamed it to osrmPathOrHostname

packages/chaire-lib-backend/src/config/server.config.ts
GabrielBruno24: I copied the comments from config_test.js. I'm not sure why the original says that.

packages/chaire-lib-common/src/config/__tests__/Preferences.test.ts
GabrielBruno24: I meant it more in the way that osrmRouting.directoryPrefix is still in defaultPreferences, so in the future we should decide wether we move it to server.config like the rest. I added similar comments to the other uses of osrmRouting from Preferences.

examples/config.js
```diff
@@ -18,27 +18,99 @@ module.exports = {
...
-    //     }
-    //   }
+    //     },
+    //     // Configuration for simple routing modes, using OSRM-like engines 
+    //     driving: {
+    //         defaultEngine: 'osrmRouting',
+    //         engines: {
+    //             osrmRouting: { 
+    //                 port: 7000, // Port used to access OSRM, either locally or remotely
+    //                 host: null, // If set to null, localhost will be used. Ignored if autoStart set to true
```
tahini, line 48: Ici donc, ça ne devrait pas être osrmPathOrHost?

packages/chaire-lib-common/src/config/__tests__/Preferences.test.ts
greenscientist: The question is not if we should or not. We definitely should. If you don't want to migrate it now, the comment should read along the lines of "This value also need to be migrated to a config like we did for ..."

greenscientist:
> @GabrielBruno24 je te suggère de tester plusieurs configs localement:
> 
> 1- Une par défaut: où tu mets pas de ocnfig custom 2- Une où tu déplaces ton osrm-routed vers un autre répertoire pour lui mettre un path custom local 3- Et un utilisant un serveur osrm externe (voir leur démo, ils doivent en avoir un?)
2- Sois tu le déplace, mais tu peux aussi juste pointer à ton build directory and enlever le binaire du PATH
3- Tu peux facilement rouler osrm-routed avec un container docker en local si tu veux, si tu veux pas installer au complet, mais ca reste facile à compiler et à rouler.

packages/chaire-lib-backend/src/utils/processManagers/OSRMProcessManager.ts
```diff
@@ -61,28 +59,34 @@ const availableRoutingModes = async function (): Promise<RoutingMode[]> {
...
-        const modeConfig = osrmServerPrefs.modes[routingMode];
+        const modeConfig = ServerConfig.getRoutingEngineConfigForMode(routingMode, 'osrmRouting');
         // TODO Do we want to accept camelCase equivalent of each routing modes?
         if (routingModes.includes(routingMode)) {
             // Only configure mode that are enabled
             if (modeConfig.enabled === true) {
                 const port = modeConfig.port;
                 // Use the host in the config even for local server starts as it allows to fine-tune the url to contact (for example ipv4 vs ipv6 names)
-                const host = modeConfig.host !== null ? modeConfig.host : '';
+                const host = !_isBlank(modeConfig.host) ? modeConfig.host : '';
```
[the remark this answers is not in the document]
GabrielBruno24: Effectivement.

examples/config.js
GabrielBruno24: Oui, j'aurais aussi du le changer a osrmPath avant

GabrielBruno24:
In the end, I replaced all instances of osrmPath with just host, since osrmPath is never called anywhere in the code.

tahini (approved):
I'll let @greenscientist do a second review

greenscientist:
Can you elaborate longer in the commit message, explaining the rationale of this change.

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -191,6 +353,35 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
             }
         });
     }
+
+    // If osrmRouting is configured using the old defaultPreferences format and not the new one in server.config, set it to the new format
+    // FIXME Remove once the old osrmRouting format is fully deprecated
```
greenscientist, line 358: Could be nice to have an issue number here, to make it easier to find the code to remove.

packages/chaire-lib-backend/src/utils/processManagers/OSRMServicePath.ts
```diff
@@ -11,7 +11,8 @@ import { RoutingMode } from 'chaire-lib-common/lib/config/routingModes';
 
 const defaultDirectoryPrefix = process.env.OSRM_DIRECTORY_PREFIX
     ? process.env.OSRM_DIRECTORY_PREFIX
-    : Preferences.get('osrmRouting.directoryPrefix', '');
+    : // TODO: Migrate this value from preferences to config like the osrm modes
+    Preferences.get('osrmRouting.directoryPrefix', '');
```
greenscientist, line 15: Just a thought for the future, I think that this can go into the new Engine OsrmConfig that you just added. 
Currently we have one global path for all, but I don't see why we could not have a configuration for each. (but maybe we can still keep a default global one. But that might be superfluous

packages/chaire-lib-common/src/services/routing/OSRMRoutingService.ts
```diff
@@ -88,6 +88,7 @@ export default class OSRMRoutingService extends RoutingService.default {
             const origin = waypoints[0];
             const destination = waypoints[waypoints.length - 1];
 
+            // TODO: TODO: Migrate this value from preferences to config like the osrm modes
```
greenscientist, line 91: Maybe, but maybe, we don't need the double TODO here :)

greenscientist: ok, but this one I'm not 100% sure what's this one is used for, if it's a runtime thing or not. But we can see later

packages/chaire-lib-common/src/services/routing/OSRMRoutingService.ts
```diff
@@ -190,6 +191,7 @@ export default class OSRMRoutingService extends RoutingService.default {
             annotations: _isBlank(params.annotations) ? true : params.annotations,
             // gaps      : 'ignore',
             continue_straight:
+                // TODO: TODO: Migrate this value from preferences to config like the osrm modes
```
greenscientist, line 194: TODO TODO

tests/config_test.js
```diff
@@ -48,82 +48,82 @@ module.exports = {
           directoryPrefix: 'test',
           modes: {
             driving: {
-                // set host to null when you want node to start osrm locally with the provided osrmPath and port
+                // set host to null when you want node to start osrm locally with the provided port
```
greenscientist, line 51: Should we remove all this from the preference? Or only when we remove the deprecated part? @tahini  ? 

If only the deprecated part, we should add a comment to it. (And maybe an issue number like I've said ealier

greenscientist (changes requested):
Nothing major, just a few small nits and things about comments. 

Still requesting changes, mostly for the commit message.

examples/config.js
```diff
@@ -18,27 +18,99 @@ module.exports = {
...
-    //           debug: false,
-    //           logs: { maxFileSizeKB: 5120, nbFiles: 3 }
-    //         },
-    //         batch: {
-    //           port: 14000,
-    //           // Enable caching of connections for all scenarios in trRouting. Will use more memory and may not be necessary for batch calculations as currently there's only one scenario per task
-    //           cacheAllScenarios: false,
-    //           debug: false,
-    //           logs: { maxFileSizeKB: 5120, nbFiles: 3 }
+    //     transit: {
```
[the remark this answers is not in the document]
GabrielBruno24: The rest of the file uses half spacing, so I can just change it to that. Is there a reason we specifically want half spacing though?

tests/config_test.js
tahini: Indeed, the test config should ideally be with the new format and not keep the deprecated one.

GabrielBruno24:
I added two issues, one for changing all the old osrm preferences to the new config format, and the other to migrate the remaining osrm settings still in defaultPreferences.

greenscientist (changes requested):
you are still missing a more elaborate commit message

GabrielBruno24:
> you are still missing a more elaborate commit message

Changed the message

greenscientist (approved):
Merci

Issue #709, opened 2023-09-08
batchRouting: trRouting is sometimes left in defunct state


tahini:
I've seen it locally once or twice and it happens on the server too, trRouting remains defunct, so all batch queries after have errors becuse trRouting cannot be started as it is considered as still running but it doesn't answer...

No idea why, the trRouting's last logs don't show anything. And so much happened on the server that logs from that moment are lost.

greenscientist:
Cannot be started from the Transition perspective or because the socket cannot be bind to?

greenscientist:
One "short term" solution would be to check the state of the process pointed by the PID file and wipe the file if the process is defunct. That should allow for a restart.

greenscientist:
Might be related to issues like https://github.com/chairemobilite/trRouting/issues/255

greenscientist:
Yes, that's what I meant. The defunct state is probably cause by some race condition during a crash or something like that.

greenscientist:
Log potentiellement du process defunct
[trRouting.log](https://github.com/chairemobilite/transition/files/15041370/trRouting.log)

tahini:
À noter aussi le flux haut niveau de ce qui a été fait sur Transition au moment où cette situation s'est produite:

```
1. Supprimé la tâche en cours
2. Annulé la même tâche en cours en me disant que j'aurais peut-être dû faire ça en premier avant de la supprimer
3. Supprimé des [lignes]
4. Mis à jour le cache
5. Démarré un nouveau calcul multiple
```

greenscientist:
Y'a peut-etre quelque chose dans la séquence suppression/annulation qui mêler Node.js et qui l'empeche d'acknowledger son child

greenscientist:
A thought: when we cancel a batch calculation we need to kill the running trRouting. But maybe we have some request still in flight. We might need to first stop the calculation, maybe sure we terminate all in flight request before we go and kill trRouting. If we kill trRouting first, maybe we leave some socket in a weird state. (Should not happen, but we never know)

greenscientist:
I think that the p-queue.clear() on cancellation is too  agressive in TrRoutingBatch:
                    if (this.options.isCancelled()) {
                        promiseQueue.clear();
                    }
This will immediately empty the queue, then the await onIdle will return. Some pending query might still be pending, we might then kill trRouting before those queries are done.

greenscientist:
ok, I don't see anything that we could do with the p-queue directly. 
[...]

PR #1092, opened 2024-11-07
batchRoute: Better handle job cancellation and deletion

tahini:
fixes #710

* Cancelled jobs sometimes were not really cancelled if a checkpoint was set after the job status was set to cancelled in the main thread, but before the automatic task refresh every 5 seconds, so we refresh the task before updating it in the checkpoint update.
* Also, refreshing and saving the job would throw an error if the job is deleted from the DB and would kill the running thread, which would cause the trRouting process to be left defunct. So we add a catch when refreshing/updating the job and stopping the trRouting process is done in the finally block, to make sure it is always stopped, even in case of an exception.

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
+++ b/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
@@ -195,12 +195,9 @@ class TrRoutingBatch {
             checkpointTracker.completed();
 
             this.options.progressEmitter.emit('progress', { name: 'BatchRouting', progress: 1.0 });
-            this.options.progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 0.0 });
 
-            const stopStatus = await TrRoutingProcessManager.stopBatch();
-
-            this.options.progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 1.0 });
-            console.log('trRouting multiple stopStatus', stopStatus);
+            // FIXME Should we return here if the job is cancelled? Or we still
+            // generate the results that have been calculated since now?
 
             // Generate the output files
             this.options.progressEmitter.emit('progress', { name: 'GeneratingBatchRoutingResults', progress: 0.0 });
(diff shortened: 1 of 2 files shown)


greenscientist:
@tahini , I think you refered to the wrong issue here, 710 is a PR.

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -195,12 +195,9 @@ class TrRoutingBatch {
             checkpointTracker.completed();
 
             this.options.progressEmitter.emit('progress', { name: 'BatchRouting', progress: 1.0 });
-            this.options.progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 0.0 });
-
-            const stopStatus = await TrRoutingProcessManager.stopBatch();
 
-            this.options.progressEmitter.emit('progress', { name: 'StoppingRoutingParallelServers', progress: 1.0 });
-            console.log('trRouting multiple stopStatus', stopStatus);
+            // FIXME Should we return here if the job is cancelled? Or we still
```
greenscientist, line 199: Do you have an issue to track this? That seems important to get an answer to that

packages/transition-backend/src/tasks/TransitionWorkerPool.ts
```diff
@@ -65,7 +71,7 @@ const getTaskCancelledFct = (task: ExecutableJob<JobDataType>) => {
                     clearInterval(intervalObj);
                 }
             })
-            .catch(() => (refreshError = true));
+            .catch(() => (refreshError = true)); // This will catch deleted jobs
```
greenscientist, line 74: Just checking, you added the same comment as before without changing the code. This is accurate and not just a bad copy/paste ?

greenscientist (approved):
Sounds good. 
Just check my notes about the comments.

tahini:
> @tahini , I think you refered to the wrong issue here, 710 is a PR.

It's #709... how could I mix 709 and 710?

packages/transition-backend/src/tasks/TransitionWorkerPool.ts
tahini: It's accurate, it adds an explanation of which kind of use case would be caught here.

Issue #1111, opened 2024-11-21
Add job to compile latex documentation into PDF to validate it.


GabrielBruno24:
After the merge of #1094, add a job to the pipeline that will compile the latex documents into a pdf, as a way of checking that the latex is valid and does not have any errors.

PR #1116, opened 2024-11-27
Build latex definition in CI

greenscientist:
To insure that the latex file are not corrupted, let's build them in the CI

Closes: #1111

### DIFF
--- a/.github/workflows/latex.yml
+++ b/.github/workflows/latex.yml
@@ -1,0 +1,25 @@
+# Workflow that validate latex files
+name: latex definition files
+
+on:
+  push:
+    branches: [ main ]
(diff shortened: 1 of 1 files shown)


GabrielBruno24:
With the way this is set up this doesn't actually run the action on this PR. Can you make a small trivial change in the tex files so that it will actually run and we can check it works correctly?

kaligrafy (approved):

GabrielBruno24 (approved):

greenscientist:
@GabrielBruno24 , I added a commit with a broken tex file. It was run and detected properly. I'll remove the commit, but you can see it with hash 24ffc70

tahini (approved):

PR #1122, opened 2024-11-28
Clean up withTranslation()

GabrielBruno24:
Many of the uses of withTranslation() had outdated arguments, either missing translation files or not including needed ones.

This PR updates them to correctly match the current content of the files. Also removes brackets in case of 1 file in the argument.

Also removes the instance of withTranslation() that have 0 arguments, updating some props types that extend WithTranslation along the way.

### DIFF
--- a/packages/chaire-lib-frontend/src/components/pageParts/Header.tsx
+++ b/packages/chaire-lib-frontend/src/components/pageParts/Header.tsx
@@ -74,7 +74,7 @@ const UserMenuButton: React.FunctionComponent<UserMenuButtonProps & WithTranslat
         </React.Fragment>
     );
 };
-const TranslatableUserMenuButton = withTranslation('menu')(UserMenuButton);
+const TranslatableUserMenuButton = withTranslation()(UserMenuButton);
 
 interface UserMenuProps {
     user: CliUser;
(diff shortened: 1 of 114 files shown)


greenscientist:
Definitely more a review for @tahini , maybe @kaligrafy . I have no idea what is going on :)

packages/chaire-lib-frontend/src/components/pageParts/Header.tsx
```diff
@@ -115,7 +115,7 @@ const UserMenu: React.FunctionComponent<UserMenuProps & WithTranslation> = (prop
         </div>
     );
 };
-const TranslatableUserMenu = withTranslation('menu')(UserMenu);
+const TranslatableUserMenu = withTranslation()(UserMenu);
```
tahini, line 118: Pourquoi laisser des parenthèses vides? Si on mets les namespaces, ne faudrait-il pas les mettre partout?

tahini (commented):
Merci pour le nettoyage! J'ai juste un questionnement sur les valeurs laissées vides. Qu'est-ce que ça signifie? Est-ce des composants qui n'ont pas de traductions finalement? Si oui, on peut enlever le withTranslation complètement?

tahini (approved):
excellent merci!

kaligrafy (approved):

PR #1136, opened 2024-12-06
doc tooltip: Fetch on demand and cache definition results

tahini:
The `DocumentationTooltip` component now requests the definition only if the user clicks to display the tooltip. The loader is shown while it is loading.

The `DefinitionService` caches the results from previous server calls, such that it is loaded only once. If a specific definition request is ongoing, it does not call the server again but waits for the server to return a value or error. Upon error, the next call to get the definition will try to get it again.

Add unit tests for the `getDefinitionFromServer` function

### DIFF
--- a/packages/transition-frontend/src/services/definitions/DefinitionsService.ts
+++ b/packages/transition-frontend/src/services/definitions/DefinitionsService.ts
@@ -5,25 +5,56 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 
-import React from 'react';
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import * as Status from 'chaire-lib-common/lib/utils/Status';
+import { de } from 'date-fns/locale';
 import { Dictionary } from 'lodash';
+
+// Cache the definitions once they are fetched. This is to avoid fetching the
+// same definition multiple times. 'loading' means a request to the server is
+// ongoing and further requests for the same label can wait for it to return.
+// 'error' means the server returned an error for this definition. If the
+// definition is not in this cache, it has not been fetched yet.
+const definitionCache: { [label: string]: Dictionary<string> | 'loading' | 'error' } = {};
 
 export const getDefinitionFromServer = (
     label: string,
-    setDefinition: React.Dispatch<React.SetStateAction<Dictionary<string>>>,
-    setGotError: React.Dispatch<React.SetStateAction<boolean>>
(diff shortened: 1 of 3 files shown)


packages/transition-frontend/src/services/definitions/DefinitionsService.ts
```diff
@@ -5,23 +5,49 @@
...
+                        setGotError(true);
+                    } else {
+                        setDefinition(definitionCache[label] as Dictionary<string>);
+                    }
+                }
+            }, 500);
+        } else {
+            setDefinition(definitionCache[label] as Dictionary<string>);
+        }
+        return;
```
greenscientist, line 37: instead of return here, maybe we could simply have an else for the rest? The nesting is not too deep in the function.

packages/transition-frontend/src/services/definitions/DefinitionsService.ts
```diff
@@ -5,23 +5,49 @@
...
  */
 
-import React from 'react';
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import * as Status from 'chaire-lib-common/lib/utils/Status';
 import { Dictionary } from 'lodash';
 
+const definitionCache: { [label: string]: Dictionary<string> | 'loading' | 'error' } = {};
+
 export const getDefinitionFromServer = (
```
GabrielBruno24, line 14: Maybe just add a comment that explains the definition will only be gotten from the server the first time and after that is taken from the cache?

GabrielBruno24 (approved):
The code looks good and works on my end. I just have one comment that the cache could use more explanation.

packages/transition-frontend/src/services/definitions/DefinitionsService.ts
```diff
@@ -5,25 +5,55 @@
...
-            if (Status.isStatusOk(definitionFromServer)) {
-                setDefinition(Status.unwrap(definitionFromServer));
-            } else {
-                setGotError(true);
+    if (definitionCache[label] === 'loading') {
+        // Wait for the previous call to complete
+        const interval = setInterval(() => {
+            if (definitionCache[label] !== 'loading') {
+                clearInterval(interval);
+                if (definitionCache[label] === 'error') {
```
greenscientist, line 30: We won't do it now, but maybe something to think about later. 
In case of error, we should note the timestamp and retry at a moment later.

Issue #408, opened 2022-08-09
transition: Get minimal data in geojson collections and fetch when required


tahini:
Ideally, we should fix chairemobilite/transition#287 to reduce memory footprint, but a good first step would be to return only the data from the nodes and path that is required for display purposes: id, name, code, etc. All other data can be fetched on demand.

For example, by just removing the `data` fields from the nodes and paths, my local memory went from 237MB to 119.

- [ ] But to avoid side effects, it would require first to move to the backend anything making use of the node and path's data:
- [ ] Also, some fields of the PathAttributes and NodeAttributes would not be present in some cases. We should have 2 different types and use them where each is required. The Path and Node classes may need to be modified to either match the full class or the partial geojson object, or have 2 different classes for both.

PR #1138, opened 2024-12-06
nodes: remove the `data` field from geojson collection

tahini:
This improves the time and memory necessary to retrieve the node geojson collection and it does not appear to have side effects.

On a large instance (all the greater Montreal area) this divides by 3 the query time to get the node geojson collection and can reduce the memory usage in the browser by 100 MB.

Tested in the UI:

* Node edits
* Multiple node edits
* Path creation and routing

### DIFF
--- a/packages/transition-backend/src/models/db/transitNodes.db.queries.ts
+++ b/packages/transition-backend/src/models/db/transitNodes.db.queries.ts
@@ -94,7 +94,6 @@ const geojsonCollection = async (
             'name', name,
             'description', description,
             'geography', ST_AsGeoJSON(geography)::jsonb, 
-            'data', data,
             'is_enabled', is_enabled,
             'integer_id', integer_id,
             'routing_radius_meters', routing_radius_meters,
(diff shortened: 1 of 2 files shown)


packages/transition-backend/src/models/db/transitNodes.db.queries.ts
```diff
@@ -94,7 +94,6 @@ const geojsonCollection = async (
             'name', name,
             'description', description,
             'geography', ST_AsGeoJSON(geography)::jsonb, 
-            'data', data,
```
greenscientist, line 97: Is the // TODO above related to this field or to "data" in general??

tahini: It refers the the content in general, we may have a bit too much data still, but that saves a lot already

tahini: The TODO refers to https://github.com/chairemobilite/transition/issues/408

PR #1139, opened 2024-12-06
path: let the path geojson collection db query return the features

tahini:
It seems that typescript is much more performant than the database to coalesce json objects.

For a large transit network (the whole Montreal area), the time for this query is divided by 3 if typescript is responsible to create the feature collection instead of the DB.


greenscientist (commented):
I don't understand the changes, I don't see how both version are equivalent. This might need more details in the commit message or in the comment.

greenscientist (commented):
don't we just miss an index to make the DB query more performant? Have you try an explain on it ?

greenscientist:
On the other side, your code is simpler, but I'm not sure why!

tahini:
> don't we just miss an index to make the DB query more performant? Have you try an explain on it ?

I did do an explain and no, we don't miss an index, even removing the join, where and order by clauses, it still takes pretty much the same time. It's the `jsonb_agg`  that takes the most time and no index can improve that. Anyway, a DB is not at its best when playing with json data, while typescript is. I could probably try the same thing with the nodes query and see if it improves the time too. I'll add more comments in the message later

greenscientist (approved):
typo in collection :)

And it would be clearer with a 2nd entry

greenscientist (approved):

PR #1143, opened 2024-12-13
accessibility map: check that data is defined for accessiblePlaces

tahini:
Now that the NodeCollection does not return the 'data' field, when nodes are retrieved from this collection, the data field will be undefined, so we make sure to undefined-check it before trying to read accessiblePlaces data.


tahini:
#1144 tracks what was broken

kaligrafy (approved):

PR #1148, opened 2024-12-16
add a readme to explain why icons.ai

kaligrafy:


samuel-duhaime (approved):

greenscientist (commented):
Can you elaborate just a bit more . 

Like give an example of file that was created with this

kaligrafy:
> Can you elaborate just a bit more .
> 
> Like give an example of file that was created with this

Added more info

greenscientist (approved):

PR #1150, opened 2024-12-16
Add a warning message that import dir does not exist in downloadOsmNe…

greenscientist:
…tworkData

Closes: #739


greenscientist:
Let's do this instead of #740

kaligrafy (approved):

tahini (approved):

PR #1155, opened 2024-12-16
Remove Lint warnings from transition-common

GabrielBruno24:
The lint warnings from transition-common (other than "unexpected any") have been removed.

65 warnings got removed (143 to 78)


tahini (commented):
Un petit truc à enlever de la doc, sinon, c'est bon

tahini (approved):

Issue #951, opened 2024-05-21
Penalize ways with bicycle=discouraged


davidmurray:
See discussion here: https://github.com/chairemobilite/transition/pull/950

With the function `way:get_value_by_key("bicycle")`

Possible example (untested!)

```diff
function safety_handler(profile, way, result, data)
  -- convert duration into cyclability
  if profile.properties.weight_name == 'cyclability' then
    local safety_penalty = profile.unsafe_highway_list[data.highway] or 1.
    local is_unsafe = safety_penalty < 1

    -- primaries that are one ways are probably huge primaries where the lanes need to be separated
    if is_unsafe and data.highway == 'primary' and not data.is_twoway then
      safety_penalty = safety_penalty * 0.5
    end
    if is_unsafe and data.highway == 'secondary' and not data.is_twoway then
      safety_penalty = safety_penalty * 0.6
    end

    local forward_is_unsafe = is_unsafe and not data.has_cycleway_forward
    local backward_is_unsafe = is_unsafe and not data.has_cycleway_backward
    local is_undesireable = data.highway == "service" and profile.service_penalties[data.service]
    local forward_penalty = 1.
    local backward_penalty = 1.
    if forward_is_unsafe then
      forward_penalty = math.min(forward_penalty, safety_penalty)
    end
    if backward_is_unsafe then
      backward_penalty = math.min(backward_penalty, safety_penalty)
    end

    if is_undesireable then
      forward_penalty = math.min(forward_penalty, profile.service_penalties[data.service])
      backward_penalty = math.min(backward_penalty, profile.service_penalties[data.service])
    end

+   -- Add penalty for bicycle=discouraged tag
+   local bicycle_tag = way:get_value_by_key("bicycle")
+   if bicycle_tag == "discouraged" then
+     forward_penalty = forward_penalty * 0.5
+     backward_penalty = backward_penalty * 0.5
+   end

    if result.forward_speed > 0 then
      -- convert from km/h to m/s
      result.forward_rate = result.forward_speed / 3.6 * forward_penalty
    end
    if result.backward_speed > 0 then
      -- convert from km/h to m/s
      result.backward_rate = result.backward_speed / 3.6 * backward_penalty
    end
    if result.duration > 0 then
      result.weight = result.duration / forward_penalty
    end
  end
end

```

_Originally posted by @davidmurray in https://github.com/chairemobilite/transition/issues/950#issuecomment-2122688947_

tahini:
This is fixed by #1159, we can close

PR #1159, opened 2024-12-16
Add a weight for bicycle=discouraged tag in OSRM profile

greenscientist:
In general the OSM community do not encourage the use for this tag, but it can be a good indication of dangerous bike routes. (We have a few street marked as such in Montreal)

This add a penalty to routing on those segments.

This was tested on De La Verandrye street, after setting the secondary default weight to 0.9 to have OSRM first route on it. After adding this code with the 0.9 weight, we see that the road is avoided.


greenscientist:
Closes #951

kaligrafy (approved):

PR #1169, opened 2024-12-19
nodes db: return an empty data field for the geojson collection

tahini:
The `data` field is mandatory in the `NodesAttributes`, but it was just dropped from the geojson collection. We instead return an empty object for data, to make sure it exists.


kaligrafy (approved):

greenscientist:
Instead of this, should we make it optional and have the people who access it check if it exist?

greenscientist:
If we do this, should we undo https://github.com/chairemobilite/transition/commit/ee4cddb602d37090b24315c3c044ad741cde0bac ???

tahini:
we can't make it optional, this is a field in GenericObject !!! (but this may also evolve, with the change of uuid to numeric IDs, we won't be able to extend GenericObject as is anymore and we can think of a new and improved way to implement common features in object, using composition and interfaces instead of inheritance, but that's in the long term)

But indeed, we can revert ee4cddb

greenscientist (approved):

PR #1173, opened 2024-12-19
Remove Lint warnings from chaire-lib-frontend

GabrielBruno24:
The lint warnings from chaire-lib-frontend (other than "unexpected any") have been removed.

29 warnings got removed (122 to 93).

### DIFF
--- a/packages/chaire-lib-frontend/src/components/pages/LoadingPage.tsx
+++ b/packages/chaire-lib-frontend/src/components/pages/LoadingPage.tsx
@@ -7,7 +7,7 @@
 import React from 'react';
 import Loader from 'react-spinners/HashLoader';
 
-const LoadingPage = function (props) {
+const LoadingPage = function (props: { message?: string }) {
     return (
         <div className="_fill" style={{ flexDirection: 'column' }}>
             {typeof props.message !== 'undefined' && <h1>{props.message}</h1>}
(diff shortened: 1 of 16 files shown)


packages/chaire-lib-frontend/src/components/pages/LoadingPage.tsx
```diff
@@ -7,10 +7,10 @@
 import React from 'react';
 import Loader from 'react-spinners/HashLoader';
 
-const LoadingPage = function (props) {
+const LoadingPage = function (loadingPageProps) {
```
tahini, line 10: Maybe here it's rpeferable to add the correct type instead of silencing the warning by renaming the props. For simplicity, you can inline the type `props: { message?: string }`

packages/chaire-lib-frontend/src/components/routers/PublicRoute.tsx
```diff
@@ -21,7 +21,7 @@ type PublicRouteProps = RouteProps & {
     config?: { [key: string]: unknown };
 } & PropsWithChildren;
 
-const PublicRoute = ({ component: Component, children, ...rest }: PublicRouteProps) => {
+const PublicRoute = ({ component: Component, ...rest }: PublicRouteProps) => {
```
tahini, line 24: I think the idea here was to remove the children from the props that you pass to the component (I'm not sure why), but the effect of this is that now Component will receive the children if there are any.

To fix this warning, you can do `children: _children` to rename the prop that is called `children` (I think)

tahini (changes requested):

tahini (approved):

PR #1174, opened 2024-12-19
public API: Add new endpoints

tahini:
Add endpoints to get the agencies, services and lines

Add an endpoint to get the supported versions

Add the x-added-in-version extension to the yaml documentation

### DIFF
--- a/docs/APIv1/API.yml
+++ b/docs/APIv1/API.yml
@@ -102,6 +102,19 @@ paths:
               schema:
                 $ref: 'agencies.yml#/agencies'
 
+  /api/v1/services:
+    get:
+      summary: Get all services
+      description: Get all services which are currently available in Transition
+      x-added-in-version: 1.1
+      responses:
+        '200':
+          description: Returns a list of all services currently available in Transition
+          content:
+            application/json:
+              schema:
+                $ref: 'services.yml#/services'
+
   /api/v1/routing-modes:
     get: 
       summary: Get all routing modes
--- a/docs/APIv1/API.yml
+++ b/docs/APIv1/API.yml
@@ -115,6 +115,19 @@ paths:
               schema:
                 $ref: 'services.yml#/services'
 
+  /api/v1/lines:
+    get:
+      summary: Get all lines
+      description: Get all lines which are currently available in Transition
+      x-added-in-version: 1.1
+      responses:
+        '200':
+          description: Returns a list of all lines currently available in Transition
+          content:
+            application/json:
+              schema:
+                $ref: 'lines.yml#/lines'
+
   /api/v1/routing-modes:
     get: 
       summary: Get all routing modes
--- a/docs/APIv1/API.yml
+++ b/docs/APIv1/API.yml
@@ -34,7 +34,24 @@ paths:
           content:
             text/plain:
               schema:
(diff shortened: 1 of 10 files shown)


packages/transition-backend/src/api/__tests__/public.routes.test.ts
```diff
@@ -360,82 +423,174 @@ describe('Testing API endpoints', () => {
...
-            (calculateAccessibilityMap as jest.Mock).mockResolvedValue(result);
+            mockedCalculateAccessibilityMap.mockResolvedValue(calculationResult);
             (TransitAccessibilityMapRouting as any).mockImplementation(() => mockTransitAccessibilityMapRouting);
-            (AccessibilityMapAPIResponse as any).mockImplementation((input) => ({
-                getResponse: () => input.resultParams
-            }));
 
             const response = await request(app)
-                .post('/api/v1/accessibility?withGeojson=false')
+                .post(`/api/v1/accessibility?withGeojson=false`)
```
greenscientist, line 573: Might not need to change to backticks

packages/transition-backend/src/api/__tests__/public.routes.test.ts
```diff
@@ -360,82 +423,174 @@ describe('Testing API endpoints', () => {
...
+            };
+
+            const expectedOutput = {
+                result: {
+                    nodes: resultByNode.nodes,
+                    polygons: {
+                        type: 'FeatureCollection',
+                        features: [{
+                            type: 'Feature',
+                            geometry: calculationResult.polygons.features[0].geometry,
```
greenscientist, line 614: Probably fine, but if the feature change somehow, we won't detect the change

docs/APIv1/API.yml
```diff
@@ -72,6 +72,18 @@ paths:
               schema:
                 $ref: 'scenarios.yml#/scenarios'
 
+  /api/v1/agencies:
+    get:
+      summary: Get all agencies
+      description: Get all agencies which are currently available in Transition
```
greenscientist, line 95: Do you want to add (and bump the X-version tag)

greenscientist: ok, was added in next commit

packages/transition-backend/src/api/__tests__/public.routes.test.ts
```diff
@@ -825,6 +825,14 @@ describe('Testing API endpoints', () => {
         });
     });
 
+    test('/api/v1/version endpoint', async () => {
+
+        const response = await request(app).get('/api/versions');
+
+        expect(response.status).toStrictEqual(200);
+        expect(response.body).toStrictEqual(['1.1']);
```
greenscientist, line 877: Do we want to mock the version or we want to change the test with each version bump?

greenscientist (commented):
Nothing major so far

packages/transition-backend/src/api/__tests__/public.routes.test.ts
tahini: we want to be aware of a version change and make sure it is correctly returned, so it's hard-coded. I'll add a comment.

greenscientist (approved):

PR #1176, opened 2024-12-20
Upgrade lint warnings to errors.

GabrielBruno24:
Configures lint so that every warning (expect for no-explicit-any) will now be considered errors. Also removes the remaining  warnings missed by the previous PRs.


packages/chaire-lib-backend/src/config/auth/index.ts
```diff
@@ -84,12 +84,14 @@ export default <U extends IUserModel>(authModel: IAuthModel<U>): PassportStatic
...
     }
     if (config.auth && config.auth.passwordless !== undefined) {
+        /* eslint-disable-next-line */
         const passwordlessConfig = require('./passwordless.config');
         // FIXME It used to work without the next line, not anymore... probably some compilation issue
         const pwdlessConfig = passwordlessConfig.default ? passwordlessConfig.default : passwordlessConfig;
         pwdlessConfig(passport, authModel);
     }
     if (config.auth && config.auth.anonymous === true) {
+        /* eslint-disable-next-line */
```
greenscientist, line 94: Since it's everywhere, should we keep this one about require as a warning instead of silencing them?

tahini (approved):

PR #1180, opened 2025-01-08
paths: update layers when TransitPathEdit component is mounted

### LINKED ISSUE #1178: paths: the path is not highlighted when opened in edit mode
It used to show an animation with arrows when editing a path, but currently, the path style is not updated. If the path is modified/edited, the animation comes back again.

A recent change probably introduced this bug

tahini:
fixes #1178

The upgrade to React 19 seemingly changed the handling of events in such a way that the path selection event is now handled before the `TransitPathEdit` component is mounted, while it was handled after in the React 16 version. So we call the `updateLayers` method at mount to display the selected path.


greenscientist:
Kind of curious, would other things be impacted by something like that?

tahini:
> Kind of curious, would other things be impacted by something like that?

I'd say probably, but there's no way to find out until our users report bugs. But nothing critic that could lead to bad data or calculation errors (at least nothing worse than our current "bad data path" where schedules are not updated when paths are updated)

greenscientist (approved):

PR #1206, opened 2025-01-21
Update log message when sending account creation email

greenscientist:
Update the message to make it clear that we send both a user and admins messages

### DIFF
--- a/packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
+++ b/packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
@@ -151,10 +151,14 @@ export const sendConfirmationEmail = async (
 ): Promise<void> => {
     try {
         const emails = await getConfirmEmailsToSend(user, options.strategy);
-        //TODO Claude says: forEach with async operations doesn't wait for all promises to complete. We should use Promise.all with map instead.
-        emails.forEach(async (email) => {
-            await sendEmail(email, { userConfirmationUrl: { url: options.confirmUrl }, usermail: user.email });
-        });
+        await Promise.all(
+            emails.map((email) =>
+                sendEmail(email, {
+                    userConfirmationUrl: { url: options.confirmUrl },
+                    usermail: user.email
+                })
+            )
+        );
         console.log(`${emails.length} email(s) (admin and/or user) sent for account confirmation`);
     } catch (error) {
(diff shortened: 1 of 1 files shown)


samuel-duhaime (approved):

packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
```diff
@@ -154,7 +154,7 @@ export const sendConfirmationEmail = async (
         emails.forEach(async (email) => {
             await sendEmail(email, { userConfirmationUrl: { url: options.confirmUrl }, usermail: user.email });
         });
-        console.log('Email sent for account confirmation');
+        console.log('Emails (admins and user) sent for account confirmation');
```
tahini, line 157: En fait, on ne sait pas si c'est admin _and_ user, ça dépend de la config, ça peut être juste le user si on a confirmByUser. Donc `admin and/or user` serait plus juste.

tahini (changes requested):

packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
greenscientist: Bon, encore mieux, je devrais checker le length de emails, et dire XX Emails (admins and/or user) sent for account confirmation

packages/chaire-lib-backend/src/services/auth/userEmailNotifications.ts
```diff
@@ -151,10 +151,11 @@ export const sendConfirmationEmail = async (
 ): Promise<void> => {
     try {
         const emails = await getConfirmEmailsToSend(user, options.strategy);
+        //TODO Claude says: forEach with async operations doesn't wait for all promises to complete. We should use Promise.all with map instead.
```
tahini, line 154: Well, if Claude is right, then we do not know that the emails were really sent, so, in theory, the message would be wrong. Since it was already there, I could approve as it improves a wrong message, but it would be quite straightforward to make it right.

greenscientist:
Add the switch from foreach to promise.all map as suggested by claude.

greenscientist:
rebased and formated

greenscientist:
fixed {}

tahini (approved):

PR #1214, opened 2025-01-23
Streamify import and process osm data task

GabrielBruno24:
Change the import and process osm data task so that it reads the source files with an asynchronous stream. Also adds an option in one of the functions used by the task so that it will not halt if just one geojson is missing compared to the raw OSM data.

### DIFF
--- a/packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
+++ b/packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
@@ -6,6 +6,9 @@
  */
 import { DataBase } from './dataBase';
 import GeoJSON from 'geojson';
+import fs from 'fs';
+import { pipeline } from 'node:stream/promises';
+import JSONStream from 'JSONStream';
 
 export class DataGeojson extends DataBase<GeoJSON.Feature> {
     private _data: GeoJSON.Feature[];
@@ -65,5 +68,66 @@ export class DataFileGeojson extends DataGeojson {
             }
         }
         return this._fileData || [];
+    }
+}
+
+// Instead of reading the entire file at once, this class streams it asynchronously. This allows for large files to be read without crashing the application.
+export class DataStreamGeojson extends DataGeojson {
+    private _fileData: GeoJSON.Feature[] | undefined = undefined;
+    private _filename: string;
+    private _dataInitialized: boolean;
+
+    constructor(filename: string) {
+        super({ type: 'FeatureCollection', features: [] });
+        this._filename = filename;
+        this._dataInitialized = false;
+    }
+
+    // Factory method so that we can create the class while calling an async function.
+    static async Create(filename: string): Promise<DataStreamGeojson> {
+        const instance = new DataStreamGeojson(filename);
+        await instance.streamDataFromFile();
+        return instance;
+    }
+
+    protected getData(): GeoJSON.Feature[] {
+        if (!this._dataInitialized) {
+            console.error('The GeoJSON data has not been properly initialized.');
+        }
+        return this._fileData || [];
+    }
+
+    private async streamDataFromFile(): Promise<void> {
+        try {
+            this._fileData = await this.readGeojsonData();
+            this._dataInitialized = true;
+        } catch (error) {
+            console.error('Error reading GeoJSON data file ' + this._filename, error);
+        }
+    }
+
+    private async readGeojsonData(): Promise<GeoJSON.Feature[]> {
+        console.log('Start streaming GeoJSON data.');
+        const readStream = fs.createReadStream(this._filename);
+        const jsonParser = JSONStream.parse('features.*');
+        const features: GeoJSON.Feature[] = [];
+
+        return new Promise((resolve, reject) => {
+            jsonParser.on('error', (error) => {
+                console.error(error);
+                reject(error);
+            });
+
+            jsonParser.on('data', (feature) => {
+                features.push(feature);
+            });
+
+            jsonParser.on('end', () => {
+                console.log('End of reading GeoJSON data.');
+                resolve(features);
+            });
+
+            pipeline(readStream, jsonParser);
+        });
     }
 }
--- a/packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
+++ b/packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
@@ -73,18 +73,20 @@ export class DataFileGeojson extends DataGeojson {
 
 // Instead of reading the entire file at once, this class streams it asynchronously. This allows for large files to be read without crashing the application.
 export class DataStreamGeojson extends DataGeojson {
-    private _fileData: GeoJSON.Feature[] | undefined = undefined;
+    private _fileData: GeoJSON.Feature[] = [];
     private _filename: string;
     private _dataInitialized: boolean;
 
-    constructor(filename: string) {
+    private constructor(filename: string) {
         super({ type: 'FeatureCollection', features: [] });
         this._filename = filename;
         this._dataInitialized = false;
     }
 
     // Factory method so that we can create the class while calling an async function.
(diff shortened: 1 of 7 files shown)


packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+    private _dataInitialized: boolean;
+
+    constructor(filename: string) {
+        super({ type: 'FeatureCollection', features: [] });
+        this._filename = filename;
+        this._dataInitialized = false;
+    }
+
+    // Factory method so that we can create the class while calling an async function.
+    static async Create(filename: string): Promise<DataStreamGeojson> {
```
tahini, line 87: lower case 'c' for the function name here

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+        const instance = new DataStreamGeojson(filename);
+        await instance.streamDataFromFile();
```
tahini, line 89: Do you actually read the whole file in the factory method, effectively adding all features in memory right now? I'm not sure this will solve the problem. See comment in the `readGeojsonData` method below.

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+        } catch (error) {
+            console.error('Error reading GeoJSON data file ' + this._filename, error);
+        }
+    }
+
+    private async readGeojsonData(): Promise<GeoJSON.Feature[]> {
+        console.log('Start streaming GeoJSON data.');
+        const readStream = fs.createReadStream(this._filename);
+        const jsonParser = JSONStream.parse('features.*');
+        const features: GeoJSON.Feature[] = [];
```
tahini, line 113: Here, you seem to fill the features array with the file content. If the file content is too big for memory, so will be the `features` array, no? Ideally, each feature should be "processed" (whatever processed means to the consumer of this class) and dropped after processing to avoid filling the memory.

tahini (changes requested):
Let me know if I am mistaken in my review, but I'm not sure the intended goal of streaming the operation works correctly.

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
greenscientist: No, because the main original limitation was that the whole file was put into a big string and this had a max size. 

Yes, this way takes a lot of memory, but Node seem to be able to cope since it's in a small chunk. We will need to refactor that further, but I don't think we have time for that at the moment.

tahini: Ok, so it is less limited than it used to be and an intermediary step for later when all is in postgis? fair enough. then just fix the lowercase 'c' in the function name and it's good

greenscientist: We'll wait for confirmation from @GabrielBruno24 that it actually work on a region as big as Montreal. 
Would be interesting to see if there's a limit

GabrielBruno24: It takes forever because the time is O(N^2) but it does work if you give node enough memory yes. It would work better if all the write actions were part of a stream like I did with task 1 and 1b but the logic here is a lot more complex, so it would be better to just rewrite the function from scratch.

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+    private _filename: string;
+    private _dataInitialized: boolean;
+
+    constructor(filename: string) {
+        super({ type: 'FeatureCollection', features: [] });
+        this._filename = filename;
+        this._dataInitialized = false;
+    }
+
+    // Factory method so that we can create the class while calling an async function.
```
greenscientist, line 86: Can we elaborate a bit more in the comment here to explain that this is a async/not async work around to avoid having to rewrite everything. 
Maybe in a TODO: form

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+    static async Create(filename: string): Promise<DataStreamGeojson> {
+        const instance = new DataStreamGeojson(filename);
+        await instance.streamDataFromFile();
+        return instance;
+    }
+
+    protected getData(): GeoJSON.Feature[] {
+        if (!this._dataInitialized) {
+            console.error('The GeoJSON data has not been properly initialized.');
```
greenscientist, line 95: Could we throw an exception here?

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,64 @@ export class DataFileGeojson extends DataGeojson {
...
+        }
+        return this._fileData || [];
```
greenscientist, line 97: Is it still possible for _fileData to be false.  (Do we still need the || []

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
greenscientist: what is enough memory? (ballpark)

packages/chaire-lib-common/src/tasks/dataImport/data/osmGeojsonService.ts
```diff
@@ -298,13 +301,17 @@ const getGeojsonsFromRawData = (
                     features[i].type,
                     features[i].id
                 );
-                throw 'Missing OSM geojson';
+                if (options.continueOnMissingGeojson) {
```
greenscientist, line 304: Maybe we should adapt the error message above? Should it says also "Or verify the OSM data"

greenscientist (changes requested):
a few small things on my side

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
GabrielBruno24: About 8 GB

greenscientist:
looks good to me

packages/chaire-lib-common/src/tasks/dataImport/data/dataGeojson.ts
```diff
@@ -67,3 +70,69 @@ export class DataFileGeojson extends DataGeojson {
...
+        super({ type: 'FeatureCollection', features: [] });
+        this._filename = filename;
+        this._dataInitialized = false;
+    }
+
+    // Factory method so that we can create the class while calling an async function.
+    // The proper way to do this would be to do it in getData(), but making that method async would force us to modify every class this inherits from, so we use this factory workaround.
+    // TODO: Rewrite the class from scratch so that it accepts an async getData().
+    static async create(filename: string): Promise<DataStreamGeojson> {
+        const instance = new DataStreamGeojson(filename);
```
tahini, line 90: Idéalement, avec l'approche factory, les données auraient pu être lues avant d'appeler le constructeur, ainsi l'objet aurait toujours été initialisé. Mais comme on va réécrire tout ça from scratch bientôt, ça va être correct. Le commentaire est plus pour une autre éventuelle factory que tu aurais à écrire un jour futur.

tahini (approved):

PR #1215, opened 2025-01-24
Fix enhance polygon task 

GabrielBruno24:
Change the enhance polygon task so that it will read its source files with an asynchronous stream, allowing it to work on large files. It also fixes a bug where the levels of buildings was always set to 1. Now, the level is only set to 1 if it does not already exist, and the levels:imputed attributed is marked as true in that case.

### DIFF
--- a/packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
+++ b/packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
@@ -4,18 +4,17 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
-import GenericDataImportTask from './genericDataImportTask';
 import GeoJSON from 'geojson';
-import { DataFileGeojson } from './data/dataGeojson';
-import { SingleGeoFeature } from '../../services/geodata/GeoJSONUtils';
 import { area as turfArea } from '@turf/turf';
+import { pipeline } from 'node:stream/promises';
+import JSONStream from 'JSONStream';
+import fs from 'fs';
+
+import GenericDataImportTask from './genericDataImportTask';
+import { SingleGeoFeature } from '../../services/geodata/GeoJSONUtils';
 import { ignoreBuildings } from '../../config/osm/osmBuildingQuerySetup';
 import { queryResidentialBuildings } from '../../config/osm/osmResidentialBuildingTags';
 import { findOverlappingFeatures } from '../../services/geodata/FindOverlappingFeatures';
-
-type PolygonGeojson =
-    | GeoJSON.Feature<GeoJSON.Polygon, GeoJSON.GeoJsonProperties>
-    | GeoJSON.Feature<GeoJSON.MultiPolygon, GeoJSON.GeoJsonProperties>;
 
 export default class enhanceAndSaveOsmPolygonData extends GenericDataImportTask {
     private _warnings: string[] = [];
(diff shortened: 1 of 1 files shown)


packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
```diff
@@ -55,149 +54,194 @@ export default class enhanceAndSaveOsmPolygonData extends GenericDataImportTask
         console.error('ERROR: %s, %s', error, this._geojsonOutputter.toString(geojson));
     }
 
-    private readAndEnhancePolygons(
-        osmGeojsonFeatures: GeoJSON.Feature<GeoJSON.Geometry, GeoJSON.GeoJsonProperties>[]
-    ): PolygonGeojson[] {
-        const polygons: PolygonGeojson[] = [];
+    private async findBuildingFeatures(
+        sourceFile: string
+    ): Promise<GeoJSON.Feature<GeoJSON.Geometry, GeoJSON.GeoJsonProperties>[]> {
```
tahini, line 59: This function should probably stream the buildings instead of returning an array

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
```diff
@@ -4,19 +4,18 @@
...
+import JSONStream from 'JSONStream';
+import fs from 'fs';
+
+import GenericDataImportTask from './genericDataImportTask';
+import { SingleGeoFeature } from '../../services/geodata/GeoJSONUtils';
 import { ignoreBuildings } from '../../config/osm/osmBuildingQuerySetup';
 import { queryResidentialBuildings } from '../../config/osm/osmResidentialBuildingTags';
 import { findOverlappingFeatures } from '../../services/geodata/FindOverlappingFeatures';
 
-type PolygonGeojson =
```
tahini, line 16: Why remove this type and use the less precise `Geometry` instead?

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
```diff
@@ -55,149 +54,194 @@ export default class enhanceAndSaveOsmPolygonData extends GenericDataImportTask
...
+                    }
+                } else {
+                    feature.properties['polygon:area'] = areaSqMeters;
                 }
-                feature.properties['building:levels'] = feature.properties['building:levels'] || 1;
-                feature.properties['building:floor_area:calculated'] = true;
-                floorArea = areaSqMeters; // use building area like there is only one floor as default floor area value
-            }
-            feature.properties['building:floor_area'] = floorArea ? floorArea : undefined;
+                // if no floor area available or calculated and the polygon is a building and non residential and non ignored, trigger warning:
```
tahini, line 186: Une chance qu'il y a un trigger warning, sinon, je pèterais ma coche avec un `if` de même! :D Je suis probablement à blâmer...

tahini (commented):
On va devoir passer à du postgis très bientôt pour ces requêtes, mettons qu'avec des grandes zones, ça risque d'exploser quand même.

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
GabrielBruno24: Oui, il faudrait probablement bouger ca dans sa propre fonction...

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
```diff
@@ -152,11 +152,14 @@ export default class enhanceAndSaveOsmPolygonData extends GenericDataImportTask
...
-                feature.properties['building:levels'] = feature.properties['building:levels:imputed']
-                    ? 1
-                    : feature.properties['building:levels'];
+                // set building:levels to 1 if they do not already have one
+                // TODO, check if there could be building:levels = 0 for small buildings and deal with it if so
+                if (feature.properties['building:levels']) {
+                    feature.properties['building:levels:imputed'] = false;
+                } else {
+                    feature.properties['building:levels:imputed'] = true;
+                    feature.properties['building:levels'] = 1;
```
greenscientist: Can you add a //TODO here : impute number of level based on height if available. (Maybe you have a task number for that)

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
GabrielBruno24: This was used as the return type for readAndEnhancePolygons(). Since we don't return the polygons anymore and just directly write inside the file we have no need for this type.

samuel-duhaime: We use this type in Evolution sometimes. I think we should keep it.

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
GabrielBruno24: Not sure what you want me to change here. The array returned by this is used as a parameter for several functions, and I'm not sure if there's a way to streamify it further without having to rewrite all those functions from scratch.

packages/chaire-lib-common/src/tasks/dataImport/enhanceAndSaveOsmPolygonData.ts
GabrielBruno24: > We use this type in Evolution sometimes. I think we should keep it.

Are you sure? I don't see in anywhere in base Evolution, and it doesn't get exported.

samuel-duhaime: No, actually, after checking again, you are right. I did check the Mtl, MJ, and Longue distance survey, and we were not importing it. 

It was only there from this Transition package file of the Survey.

tahini (approved):

tahini:
Bon, j'ai pesé sur le mauvais bouton et ça a ajouté un merge commit, tu peux faire un rebase sur `main` pour l'enlever?

GabrielBruno24:
> Bon, j'ai pesé sur le mauvais bouton et ça a ajouté un merge commit, tu peux faire un rebase sur `main` pour l'enlever?

C'est fait

PR #1216, opened 2025-01-28
batch: Use the max calculation if the parallel calculation is unset

tahini:

This will allow long started jobs to be restarted on a server with more CPUs available without having to recreate the job.


packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -354,6 +360,17 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
         });
     }
 
+    // Make sure the maximum number of parallel calculators is a positive value, but does not exceed the CPU count
+    if (newConfig.maxParallelCalculators !== undefined) {
+        if (newConfig.maxParallelCalculators <= 0) {
+            console.warn('maxParallelCalculators must be a positive number. Using the number of CPUs instead');
```
GabrielBruno24, line 366: Print the values of maxParallelCalculators and the number of CPUs

packages/transition-common/src/services/batchCalculation/types.ts
```diff
@@ -8,7 +8,10 @@ import { ErrorMessage } from 'chaire-lib-common/lib/utils/TrError';
 import { validateTrQueryAttributes } from '../transitRouting/TransitRoutingQueryAttributes';
 import { TransitRoutingQueryAttributes } from 'chaire-lib-common/lib/services/routing/types';
 
-export const isBatchParametersValid = (parameters: BatchCalculationParameters) => {
+export const isBatchParametersValid = (
```
GabrielBruno24, line 11: This function could use a jsdoc comment, especially to explain what a negative maxParallelCalculations does.

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -354,6 +360,22 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
...
+        if (newConfig.maxParallelCalculators <= 0) {
+            console.warn(
+                `maxParallelCalculators (${newConfig.maxParallelCalculators}) must be a positive number. Using the number of CPUs instead: ${availableCPUs}`
+            );
+            newConfig.maxParallelCalculators = availableCPUs;
+        } else if (newConfig.maxParallelCalculators > availableCPUs) {
+            console.warn(
+                `maxParallelCalculators (${newConfig.maxParallelCalculators}) cannot exceed the number of CPUs. Using the number of CPUs instead: ${availableCPUs}`
+            );
+            newConfig.maxParallelCalculators = availableCPUs;
```
greenscientist, line 375: We should let users set a value higher than the currently available CPU. 
Could be test purposes or other. 
We can keep the warning, but not override the setting

packages/chaire-lib-backend/src/utils/processManagers/TrRoutingProcessManager.ts
```diff
@@ -188,7 +187,7 @@ const startBatch = async function (
 ) {
     // Ensure we don't use more CPU than configured
     // TODO The os.cpus().length should move to a "default config management class"
```
greenscientist, line 189: Need to update the TODO

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -351,11 +352,16 @@ class TrRoutingBatch {
         return { odTrips, errors };
     };
 
+    private getParallelCalculationCount = (): number =>
```
greenscientist, line 355: Can we have a comment on this function with a description

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -351,11 +352,16 @@ class TrRoutingBatch {
...
+        typeof this.batchRoutingQueryAttributes.parallelCalculations === 'number'
```
greenscientist, line 356: Could we use "if" instead of ? : 
just to make it easier to read

greenscientist (changes requested):

greenscientist (approved):

PR #1231, opened 2025-02-05
routes: refactor private and public route elements

tahini:
The routes are now components that wrap a route to make sure required permissions are granted. These components can use the auth state directly to know if the person is authenticated or get the current user. This avoids having to pass them as props and simplifies the components.


samuel-duhaime:
@tahini Will it fix this bug if we do that in Evolution?

https://github.com/chairemobilite/evolution/issues/641

tahini:
> @tahini Will it fix this bug if we do that in Evolution?
> 
> [chairemobilite/evolution#641](https://github.com/chairemobilite/evolution/issues/641)

Not at all, that bug has everything to do with the preload functions. Other refactoring will fix that, but not this. This is only for the React-router update in Evolution (and to simplify everything)

kaligrafy (approved):

PR #1238, opened 2025-02-11
Add a Docker environment for development (#45)

KevinPeymani:
New docker development environment for quick code updates.
>Note: Debugging with breakpoints has not been tested yet

Created a new `docker-dev/` folder with its own `Dockerfile` and `docker-compose.yml`
* The new `Dockerfile` is optimized to rebuild the image faster in case of a code change 
  * Moved `COPY . /app` to the last possible moment. This way, the docker image doesn't need to be rebuilt entirely for every file change. Some directories did still need to be copied earlier though.
  * Moved `yarn compile` right before 1. . This also keeps from rebuilding almost the entire stack when editing code.
  * Changed the `CMD` that starts the app. It now executes the file `entrypoint-dev.sh` (see below)
* `docker-dev/docker-compose.yml` has new bind-mounts to map volumes from the source code of the host to the container. This allows for quick code changes without rebuilding the image 

Added a separate start bash script to keep from opening 4 terminals to launch the app (`entrypoint-dev.sh`)

Added new documentation in `docs/devWithDocker.md` containing a tutorial on setting up the dev environment

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,18 +28,8 @@ COPY packages/transition-frontend packages/transition-frontend
 RUN yarn config set network-timeout 300000
 RUN yarn install
 
-# Copy the rest. (node_modules are excluded in .dockerignore)
-COPY . /app
-
 # Setup the example as a default configuration for the image
 COPY .env.docker /app/.env
-
-#TODO evaluate if any of those commands are necessary
-#RUN yarn setup
-#RUN yarn setup && yarn migrate && yarn create-user
-#RUN yarn compile && yarn build:dev
-RUN yarn compile
-#RUN yarn compile && yarn build:prod
 
 #TODO We probably need to do something different for the projects configuration directories
 # the docker-compose file have an example of using volume for part of a project
@@ -64,6 +54,14 @@ RUN /usr/local/bin/osrm-extract --help && \
     /usr/local/bin/osrm-contract --help && \
     /usr/local/bin/osrm-partition --help && \
     /usr/local/bin/osrm-customize --help
+
+# Copy the necessary files for compilation   
+COPY ./configs /app/configs
+COPY ./packages /app/packages
+RUN yarn compile
+
+# Copy the rest. (node_modules are excluded in .dockerignore)
+COPY . /app
 
 # Start json2capnp -> Relies on manually creating cache directory before
 # Start Node app
--- a/docker-dev/docker-compose.override.yml
+++ b/docker-dev/docker-compose.override.yml
@@ -1,0 +1,78 @@
+# Docker compose override file to map the code directories inside the 
+# container, allowing the app to rebuild dynamically
+
+# To start the docker dev environment, run this command from the root folder : 
+# docker compose -f docker-compose.yml -f docker-dev/docker-compose.override.yml up
+
+services:
+  transition-www:
+    command:
+      - /bin/sh 
+      - /app/docker-dev/entrypoint-dev.sh
+    volumes:
+      # Bind mounts on the package directories. These enable instant code changes on the container without entirely rebuilding the image   
+      # It is necessary to bind individually every file in every package, and not just the package. This is because during compilation,
+      # when each package is compiled, the container creates a lib/ folder that's needed to build and run the app. 
+      # Binding the whole package would overwrite this lib/ folder.
+      - ${PWD}/packages/chaire-lib-backend/src:/app/packages/chaire-lib-backend/src
+      - ${PWD}/packages/chaire-lib-backend/.eslintignore:/app/packages/chaire-lib-backend/.eslintignore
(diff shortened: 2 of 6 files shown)


docker-dev/docker-compose.yml
```diff
@@ -0,0 +1,125 @@
...
+      - transition-project-import:/app/examples/runtime/imports
+      - transition-project-osrm:/app/examples/runtime/osrm
+      # TODO Remove this once the cache is actually a cache
+      - transition-project-cache:/app/examples/runtime/cache
+      # CUSTOM For custom project with config file and runtime directory, add
+      # additional volumes here. The following lines adds the local
+      # `/home/myHome/transition-example` in the container's `/config` directory
+      # - "/home/myHome/transition-example/:/config"
+
+      # Bind mounts on the package directories. These enable instant code changes on the container without entirely rebuilding the image
```
greenscientist, line 47: Any reason to bind mount each file instead of the directories?

docker-dev/Dockerfile
```diff
@@ -0,0 +1,78 @@
+# Transition Docker file
```
greenscientist, line 1: Feels like we could find a way to just reuse most of the main file directly. Maybe with something that autogen this "dev" file

greenscientist:
Also, maybe some of the changes could be put in the main docker file ? (Like moving the copy around ?)

docker-dev/docker-compose.yml
KevinPeymani: Because in my experimentation, each package creates the `lib/` folder at experimentation. If I bind the whole package, the app crashes.

image

docker-dev/Dockerfile
KevinPeymani: I agree, but I've never done that before. Do you have any reference I could use to write the script ?
It's essentially the same Dockerfile as the one in the root, except for the entrypoint. So I could just remove my Dockerfile, and specifiy in the `docker-docs/docker-compose.yml` an `entrypoint:` field I think?

KevinPeymani: I could also use the first Dockerfile as a intermediary image, and then add the last layers in a separate Dockerfile

greenscientist: Since you need to have the docker-compose anyway, it could be a good option to just reuse the main Dockerfile

KevinPeymani:
@greenscientist Now the dev environment uses the root Dockerfile and docker-compose.yml. The bind mounts and dev entrypoint are now specified in `docker-dev/docker-compose.override.yml`. I've added my optimizations in the root Dockerfile.

docker-dev/docker-compose.override.yml
```diff
@@ -0,0 +1,73 @@
+# Docker compose override file to map the code directories inside the
```
greenscientist, line 1: I know it's in the doc, but can you add here how to run with this override file

Dockerfile
```diff
@@ -29,18 +29,13 @@ RUN yarn config set network-timeout 300000
 RUN yarn install
 
 # Copy the rest. (node_modules are excluded in .dockerignore)
```
greenscientist, line 31: Need to update the comments, probably where you put the "copy all" and add a comment about why we do thoses firsts.

greenscientist (changes requested):
This looks good. A few things other that the small update to comments requested. 

Please rework the commits so that the first commit is the optimisation of the Dockerfile and the second commit is the addition of the docker-dev instruction. 

Also, I,m wondering if we should move the docs/devWithDocker.md directly inside the docker-dev directory as this directory README file. (@tahini thoughts?)

KevinPeymani:
Done!

greenscientist:
@KevinPeymani can you rework the commit history to maybe have only 2 commits: one with the changes to the docker file and the other with the "dev" stuff

KevinPeymani:
Done! The branch should now have only two commits

docker-dev/docker-compose.override.yml
```diff
@@ -0,0 +1,76 @@
...
+# docker compose -f docker-compose.yml -f docker-dev/docker-compose.override.yml up
+
+services:
+  transition-www:
+    command:
+      - /bin/sh 
+      - /app/docker-dev/entrypoint-dev.sh
+    volumes:
+
+      # Bind mounts on the package directories. These enable instant code changes on the container without entirely rebuilding the image
```
greenscientist, line 14: Can you add the comment here about why we need to add all those file. So that we know if we add modules or other things what we need to add here.

greenscientist (changes requested):
Just a couple of small things and we'll be good to go. In addition to the comment in the code:
* Can you update the Dockerfile commit title to something more explicit that Edit Dockerfile, maybe something like optmize Dockerfile runs or something like that
* Rebase the stack on the latest changes in the repo

KevinPeymani:
Done!

greenscientist (approved):

PR #1240, opened 2025-02-12
path: fix schedule update on line when a path is deleted

### LINKED ISSUE #1237: path: schedules not correctly updated when path is deleted
Bug to be investigated: we do not know exactly, but here is what [we think] happened

Using the STM green line, the line was copied, the outbound path was deleted, nothing else was done, the data was saved to the cache and trRouting broke.

Somehow, the schedules were not properly re-generated and some lingering old data remained in the cache and trRouting crashed at startup.

tldr for this issue: copying a line and deleting an outbound path should work out of the box and simply

tahini:
fixes #1237 

Add a "set null" contraints on the schedule period's outbound_path_id and inbound_path_id to avoid deleting all trips if one of the path is deleted.

Refresh a line's schedule after deleting a path

Read the object from database after saving it.

### DIFF
--- a/packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
+++ b/packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
@@ -188,8 +188,16 @@ function createDataHandlers(): Record<string, TransitObjectDataHandler> {
                     }
                     if (transitClassConfig.cacheQueries.objectToCache) {
                         try {
+                            // Re-read the object from the database to update the cache
+                            // FIXME: Make sure that all read objects contain
+                            // all necessary to send to cache. Currently, only
+                            // the lines use this code path, nodes use the
+                            // `transitNode.save` socket route instead. And
+                            // lines read the schedule so it's ok. But should anything
+                            // else implement this, it should be checked.
+                            const updatedObject = await transitClassConfig.dbQueries.read(updatedId);
                             await transitClassConfig.cacheQueries.objectToCache(
-                                attributes,
+                                updatedObject.attributes ? updatedObject.attributes : updatedObject,
                                 attributes.data.customCachePath
                             );
                             return {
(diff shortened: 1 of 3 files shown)


packages/transition-common/src/services/path/Path.ts
```diff
@@ -1373,10 +1373,13 @@ export class Path extends MapObject<GeoJSON.LineString, PathAttributes> implemen
                             line.attributes.path_ids = line.attributes.path_ids.filter((pathId) => {
                                 return pathId !== this.id;
                             });
+                            // FIXME This should be one method only, to refresh the line from server
```
greenscientist, line 1376: Would probably be good to have an issue created for this!

packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
```diff
@@ -188,8 +188,16 @@ function createDataHandlers(): Record<string, TransitObjectDataHandler> {
                     }
                     if (transitClassConfig.cacheQueries.objectToCache) {
                         try {
+                            // Re-read the object from the database to update the cache
```
greenscientist, line 191: Cache invalidation is a difficult problem!

greenscientist (commented):
So with the Set NULL instead of delete cascade, is there a risk that we leave data in the DB ? Like path all set to NULL line and live there foveever?

tahini:
> So with the Set NULL instead of delete cascade, is there a risk that we leave data in the DB ? Like path all set to NULL line and live there foveever?

The risk is to end up with a schedule with periods, but without trips, which is something that trRouting will give a warning for apparently, from what I saw in the logs, but transition will not complain. I could open an issue. We don't necessarily want to automatically delete the schedule, as we may have configured the periods and want to go and update with new paths. But we could issue a warning when saving the line that there are empty services and propose to just delete them.

PR #1243, opened 2025-02-14
auth: type the `google` and `facebook` auth config

tahini:
These options are optional boolean options added to the `auth` object of the project.config's file.

Also hide the corresponding routes and auth methods conditional to the enablement of those methods instead of being available all the time.

Display an error message if one of those auth method is enabled but the required environment variables are not set.

### DIFF
--- a/packages/chaire-lib-backend/src/api/auth.routes.ts
+++ b/packages/chaire-lib-backend/src/api/auth.routes.ts
@@ -40,32 +40,34 @@ const defaultFailureCallback = (err, _req: Request, res: Response, _next) => {
 };
 
 export default <U extends IUserModel>(app: express.Express, authModel: IAuthModel<U>, passport: PassportStatic) => {
-    app.get(
-        '/googlelogin',
-        passport.authenticate('google', { scope: 'https://www.googleapis.com/auth/userinfo.email' })
-    );
-    app.get('/googleoauth', passport.authenticate('google', { failureRedirect: '/login' }), (req, res) => {
-        //console.log('GOOGLE LOGIN');
-        if (req.user) {
-            res.redirect('/survey');
-        } else {
-            return res.status(200).json({
-                error: 'User not authenticated'
-            });
-        }
-    });
+    if (config.auth && config.auth.google === true) {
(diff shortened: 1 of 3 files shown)


tahini:
This will allow evolution to import the actual project.config with all correctly typed properties (instead of any currenctly) for the `auth` object.

kaligrafy (approved):

packages/chaire-lib-backend/src/api/auth.routes.ts
```diff
@@ -40,32 +40,36 @@ const defaultFailureCallback = (err, _req: Request, res: Response, _next) => {
...
-        } else {
-            return res.status(200).json({
-                error: 'User not authenticated'
-            });
-        }
-    });
+    if (config.auth && config.auth.facebook === true) {
+        app.get('/facebooklogin', passport.authenticate('facebook'));
+        app.get('/facebookoauth', passport.authenticate('facebook', { failureRedirect: '/login' }), (req, res) => {
+            //console.log('FACEBOOK LOGIN');
```
greenscientist, line 63: While you are there. should we keep this console.log?

packages/chaire-lib-backend/src/api/auth.routes.ts
```diff
@@ -40,32 +40,36 @@ const defaultFailureCallback = (err, _req: Request, res: Response, _next) => {
...
-            });
-        }
-    });
+    if (config.auth && config.auth.google === true) {
+        app.get(
+            '/googlelogin',
+            passport.authenticate('google', { scope: 'https://www.googleapis.com/auth/userinfo.email' })
+        );
+        app.get('/googleoauth', passport.authenticate('google', { failureRedirect: '/login' }), (req, res) => {
+            //console.log('GOOGLE LOGIN');
```
greenscientist, line 49: While you are there. should we keep this console.log?

packages/chaire-lib-backend/src/api/auth.routes.ts
tahini: Removed

PR #1244, opened 2025-02-14
Add github workflow to build docker images

greenscientist:
First workflow will test docker build on each PR
Second will build the docker image and push to both docker hub and github package repo


greenscientist:
Yes, we'll need to do this for each repo!

tahini (approved):

PR #1246, opened 2025-02-14
Better logging and hard failure on config errors

greenscientist:


packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -418,8 +418,39 @@ try {
         );
     }
     setProjectConfiguration(configFromFile);
-} catch {
-    console.error(`Error loading server configuration in file ${configFileNormalized}`);
+} catch (error: unknown) {
```
GabrielBruno24, line 421: Add a comment to explain why we have such complex error handling here and not elsewhere.

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -418,8 +418,39 @@ try {
...
+            console.error('The configuration file does not exist or cannot be accessed');
+        } else if (error instanceof SyntaxError) {
+            console.error('The configuration file contains invalid JSON or JavaScript syntax');
+        }
+
+        // Application-specific error handling
+        if (process.env.NODE_ENV === 'production') {
+            // In production, you might want to use a default configuration
+            // or terminate the process
+            process.exit(1);
```
GabrielBruno24, line 444: Maybe add a console log before the process exit?

PR #1266, opened 2025-02-28
map layers: remove all references to aggregatedOd layer

### LINKED ISSUE #1248: Decide what to do about aggregatedOD
There are several references to "aggregatedOD" in the code, particularly in the layers config in defaultPreferences. However, this feature is not currently implemented. We should either implement it or remove it.

tahini:
fixes #1248

This feature is not supported yet in Transition. When we want to support it, we can add it back as specified.


kaligrafy (approved):

tahini:
Will be merged after deck.gl

Issue #1265, opened 2025-02-28
map: the layers should not be part of the preferences


tahini:
Currently, the layers to display in each section are configured in the `defaultPreferences.ts` file, which eventually gets copied entirely in to the user's preferences object. Once that is done, there is absolutely no way to unset these. That means that if we add a layer or change a layer name, the current users will never have access to it!

This data should be part of the configuration, not the preferences.

Eventually, a user will be able to activate/deactivate some layers and that would be a preference, but that is a different issue

PR #1267, opened 2025-02-28
fix: move map layers config from preferences to dedicated config

kaligrafy:
Problem:
Currently map layer visibility settings are stored in user preferences, making it impossible for new layers to be visible to existing users, as the app copies the defaults into user preferences which are then never updated.

Solution:
- Move layer visibility configurations from defaultPreferences.config.ts to a dedicated sectionLayers export in layers.config.ts
- Update TransitionMainMap to use this configuration first, with fallback to user preferences for backward compatibility
- Remove layer configuration from default preferences
- Add unit tests to validate the changes

This change ensures new layers will be visible to all users while maintaining backward compatibility.

🤖 Generated with Claude Code

### DIFF
--- a/packages/chaire-lib-common/src/config/defaultPreferences.config.ts
+++ b/packages/chaire-lib-common/src/config/defaultPreferences.config.ts
@@ -11,7 +11,6 @@ import constants from './constants';
 interface SectionDescription {
     localizedTitle: string;
     icon: string;
-    hasMapLayers: boolean;
     showMap?: boolean;
     showFullSizePanel?: boolean;
     enabled?: boolean;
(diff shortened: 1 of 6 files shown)


packages/transition-frontend/src/components/map/__tests__/TransitionMainMap.test.tsx
```diff
@@ -0,0 +1,167 @@
+/*
+ * Copyright 2022, Polytechnique Montreal and contributors
```
tahini, line 2: 2025

packages/transition-frontend/src/components/map/TransitionMainMap.tsx
```diff
@@ -70,8 +70,13 @@ class MainMap extends React.Component<MainMapProps & WithTranslation & PropsWith
     constructor(props: MainMapProps & WithTranslation) {
         super(props);
 
+        // Get layers for section from config first, fall back to user preferences for backward compatibility
```
kaligrafy, line 73: Here we don't need backward compatiobility because as soon as this PR is merged, preferences will never be read

packages/transition-frontend/src/components/map/__tests__/TransitionMainMap.test.tsx
```diff
@@ -0,0 +1,167 @@
...
+import React from 'react';
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import Preferences from 'chaire-lib-common/lib/config/Preferences';
+import { sectionLayers } from '../../../config/layers.config';
+
+import TransitionMainMap from '../TransitionMainMap';
+
+// Mocks
+jest.mock('mapbox-gl', () => ({
```
tahini, line 16: Adding tests at this moment in the main map is not a good idea as it will only make it more difficul to migrate to deck.gl. We can ask Claude to generate the tests at the end of the deck.gl branch

packages/transition-frontend/src/config/layers.config.ts
```diff
@@ -4,6 +4,79 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
+
+// Define which layers should be visible for each section
+export const sectionLayers = {
```
tahini, line 9: Do we want the layers configuration file to be aware of the sections? Technically, sections are provided by "modules" (transit offer, transit demand). But for that matter, so are layers probably and that too should not necessarily be a single layer config file... sounds like we'll need to rework our modular approach eventually!

tahini (commented):
Also please add the fixes #1265 to the commit message

packages/chaire-lib-common/src/config/defaultPreferences.config.ts
```diff
@@ -100,75 +100,9 @@ const defaultPreferences: PreferencesModel = {
...
-                'transitPathsSelected',
-                'transitPathWaypoints',
-                'transitPathWaypointsSelected',
-                'transitNodes',
-                'transitNodesSelected'
-            ]
-        }
+        // Layers configuration has been moved to transition-frontend/src/config/layers.config.ts
+        // This is kept here for backward compatibility with existing user preferences
+        layers: {}
```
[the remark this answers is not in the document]
kaligrafy: removed

packages/transition-frontend/src/components/map/__tests__/TransitionMainMap.test.tsx
kaligrafy: removed

packages/transition-frontend/src/components/map/TransitionMainMap.tsx
kaligrafy: removed

packages/transition-frontend/src/components/map/__tests__/TransitionMainMap.test.tsx
kaligrafy: this file has been removed

packages/transition-frontend/src/config/layers.config.ts
kaligrafy: We will see later (dixit @tahini)

tahini (changes requested):
Missing change in TransitionDashboard

kaligrafy:
added needed changes to TransitionDashboard

tahini (approved):

PR #1280, opened 2025-03-12
Docker dev environment : 2 small fixes related to the instructions in the documentation

KevinPeymani:
1. The documentation mentions `docker-dev-transition-www-1` as the container name, even though when you follow the instructions, the created container is called `transition-transition-www-1`. The name has been corrected accordingly.
2. When trying to setup the dev environment following the documentation, the command  `docker compose ... up --build` would fail because it would try to pull the app's image. Now the docker compose override file changes the build context and uses the root Dockerfile to build the image.

### DIFF
--- a/docker-dev/docker-compose.override.yml
+++ b/docker-dev/docker-compose.override.yml
@@ -6,6 +6,10 @@
 
 services:
   transition-www:
+    # This line is necessary to specify that the build context is at the root of the app,
+    # otherwise the Dockerfile will not be detected. The docker-compose command *must* be
+    # ran from the root folder of the app, as mentioned above
+    build: .
     command:
       - /bin/sh 
       - /app/docker-dev/entrypoint-dev.sh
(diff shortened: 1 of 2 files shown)


docker-dev/docker-compose.override.yml
```diff
@@ -6,6 +6,7 @@
 
 services:
   transition-www:
+    build: .
```
greenscientist, line 9: Can you add a comment here to explain what it does, it's not explicit

KevinPeymani: done!

greenscientist: ok. This is valid only if we call the script from the root directory. I guess it's fine, since that's what the instruction says to do.

KevinPeymani: Should I add a sentence to explain that `docker compose` must be ran from the root folder ?

greenscientist: Never hurt to have more comments :)

KevinPeymani: Done! Let me know if my explanation could be better written

greenscientist (approved):

Issue #1262, opened 2025-02-26
batch routing: Add punctual calculation speeds in log message (calc/sec)


tahini:
Currently, in the batch routing job, we have a log line that write the number of calculations per second, but this is the global calculation speed for the task since the beginning of the job. This may hide a slow start, or a faster block of calculations. We may also wish to add the current calculation speed.

PR #1284, opened 2025-03-12
Add current calculation rate to log message

greenscientist:
When running batch routing, we have a log which show the rate of calculation since the beginning of the run. Since the rate can be non constant, we want to see what is the current rate.
We calculate the rate between 2 log entries, by saving what were the values when we previously displayed the log.

New log:
 calc/sec: 2.23 (current: 2.23)
 calc/sec: 3.14 (current: 4.91)
 calc/sec: 3.35 (current: 3.8)

Closes: #1262

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
+++ b/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
@@ -136,6 +136,8 @@ class TrRoutingBatch {
             // Log progress at most for each 1% progress
             const logInterval = Math.ceil(odTripsCount / 100);
             const benchmarkStart = performance.now();
+            let lastLogTime = performance.now();
+            let lastLogCount = startIndex;
             const logOdTripBefore = (index: number) => {
                 if ((index + 1) % logInterval === 0) {
                     console.log(`Routing odTrip ${index + 1}/${odTripsCount}`);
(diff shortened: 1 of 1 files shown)


packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -144,13 +146,21 @@ class TrRoutingBatch {
...
                 if (benchmarkStart >= 0 && index > 0 && index % 100 === 0) {
                     // Log the calculation speed every 100 calculations. Divide the number of completed calculation (substract startIndex if the task was resumed) by the time taken in seconds. Round to 2 decimals
-                    console.log(
-                        'calc/sec',
+                    // Calculate rate since the last logged line
+                    const currentRate =
+                        Math.round(
+                            (100 * (completedRoutingsCount - lastLogCount)) / ((performance.now() - lastLogTime) / 1000)
+                        ) / 100;
+                    lastLogTime = performance.now();
```
tahini, line 154: On pourrait extraire le `performance.now` dans une variable locale pour éviter de faire 3 fois `performance.now()` dans le bloc. Je ne sais pas si c'est juste un simple getter sans overhead ou s'il y en a un léger, mais tant qu'à...

tahini (approved):

PR #1293, opened 2025-03-14
osm: add task to import poi and residential data in database

### LINKED ISSUE #1289: places: support the import of geojson places in Transition
This feature is still only in transition-legacy: https://github.com/chairemobilite/transition-legacy/blob/dev/src/tasks/transition/import/importPlacesFromGeojson.task.js

To support evolutionary algorithm with accessibility map feature, it will need to be ported to Transition

tahini:
fixes #1289

This tasks is the last task of the OSM import that was missing, to import the generated and cleaned data saved in a geojson file to the database.

This adds a task, `4_importPlacesToDb.task.ts` in `transition-backend` that will automatically read the POIs and residential entrances file and create/update 2 data sources, identified by the data source direction, followed by ' Places' and ' Residences' respectively.

This task uses an import function that takes a point geojson file and data source name and does the actual import, asking the user whether to overwrite any previous data source of the same name. This import function can later be re-used by other code paths to import places.


kaligrafy (approved):

packages/transition-backend/src/services/importers/PlacesDataSourceImporter.ts
```diff
@@ -0,0 +1,106 @@
...
+import dataSourcesDbQuery from 'chaire-lib-backend/lib/models/db/dataSources.db.queries';
+import { PlaceAttributes } from 'transition-common/lib/services/places/Place';
+import DataSource from 'chaire-lib-common/lib/services/dataSource/DataSource';
+import { fileManager } from 'chaire-lib-backend/lib/utils/filesystem/fileManager';
+import PQueue from 'p-queue';
+
+const pipelineAsync = promisify(pipeline);
+
+// FIXME This prompt works only on command line where the user can answer. Need to support arbitrary prompts, like going through a web interface.
+async function promptOverwriteDataSource(dataSourceName): Promise<boolean> {
```
greenscientist, line 24: Should this prompt go into the task itself and leave this code path with only parameter that are feeded by various UIs?

packages/transition-backend/src/services/importers/PlacesDataSourceImporter.ts
```diff
@@ -0,0 +1,106 @@
...
+                { name: 'Yes', value: true },
+                { name: 'No', value: false }
+            ]
+        }
+    ]);
+    return answers.overwrite;
+}
+
+/**
+ * Import the points from the geojson file into the database, in the data source with the given name. If the data source already exists, the user will be prompted to confirm overwriting the existing data.
```
greenscientist, line 40: Do we define the file format somewhere? "geojson" is quite generic no?

packages/transition-backend/src/tasks/dataImport/importPlacesFromGeojson.ts
```diff
@@ -0,0 +1,37 @@
...
+     *
+     * @param dataSourceDirectory The directory containing the data sources
+     */
+    protected async doRun(dataSourceDirectory: string): Promise<void> {
+        // Import from the point of interest file
+        const absoluteDsDir = this._importDir + dataSourceDirectory + '/';
+        const POIgeojsonFilePath = absoluteDsDir + GenericDataImportTask.POINT_OF_INTEREST_FILE;
+        const dataSourceName = `${dataSourceDirectory} Places`;
+
+        await importGeojsonPlaces(POIgeojsonFilePath, dataSourceName);
```
greenscientist, line 30: Do those 2 imports need to be sequential ? Or can we do them in "parallel"?  We could "await all" at the end of the function (Maybe with a flag in the task to select selection if we prefer, for debug purpose)

greenscientist (commented):
A few comments, most of them could maybe be saved as TODO for later

packages/transition-backend/src/services/importers/PlacesDataSourceImporter.ts
tahini: > Should this prompt go into the task itself and leave this code path with only parameter that are feeded by various UIs?

I suppose it's included in the fixme above. We would need to verify if a datasource already exists first, then request the user somehow if he wants to overwrite it, then call the import. Or else, always ask if we want to overwrite if the datasource exists, without checking its existence first.

packages/transition-backend/src/tasks/dataImport/importPlacesFromGeojson.ts
tahini: For now, they need to be sequential. The import function asks to overwrite the data source and if they run in parallel, there will be 2 questions at the same time and you don't know what you're answering to. If we eventually split the data source creation/overwrite from the import, then the import part can be done in parallel. But as currently, this task is the only one doing the import, the split can wait for another use case.

greenscientist (approved):

Issue #424, opened 2022-10-31
chaire-lib: Frontend config should be fetched from server instead of webpack


tahini:
Now, the default configuration is wrapped in the build by webpack, so it's available right from the start. It should be something that is fetched from the server upon application start. If the server restarts with new configuration options, the client does not need to be built again.

This can be done after configuration/preferences refactoring, as it will make clearer what is needed in the frontend and how.

PR #1295, opened 2025-03-17
config: get the section configuration out of preferences

### LINKED ISSUE #1286: config: Move section config out of preferences
Similarly to #1265 for the map layers, the enabled section and configuration should not be in the defaultPreferences.config file. It should be instance-specific and if the instance changes, the user should automatically be able to access new sections without having to update its preferences in the database!

When fixing this, make sure that all previous section configurations still work (for example, we have calculation instances who disabled the agencies, nodes, etc. sections, they should still work even if the configuration was not ported to the new approach.)

tahini:
fixes #1286

Add a `sections` configuration option in the config of chaire-lib, to be used by applications like Transition or Octavi that use the dashboard functionality of chaire-lib.

Add a `project.config` file in transition-common with the transition section default configuration.

Update the `webpack.config.js` file to make sure the default section config is imported before reading the config file. That is the only change required to account for the new section config option.

Update usages in the frontend to get the enabled sections from the config instead of preferences.

** Breaking change **

Section configurations on the config file's `defaultPreferences` section is now deprecated. They will still work for a while, but they need to be updated.

Configuration files can overwrite section configuration options, directly, not in the preferences, as follows:

```
module.exports = {
   ...myConfig,
   sections: {
      simulations: {
         enabled: true
      },
      routing: {
         enabled: false
      }
   }
}
```

### DIFF
--- a/packages/chaire-lib-backend/src/config/server.config.ts
+++ b/packages/chaire-lib-backend/src/config/server.config.ts
@@ -405,6 +405,22 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
         }
     }
 
+    // If there are still section configurations in the defaultPreferences section of the project config, set it as section config.
+    // FIXME Remove this if a reasonable time after march 2025, to give time to instances to update their configuration
+    const defaultPreferences = (newConfig as any)?.defaultPreferences;
+    if (defaultPreferences && defaultPreferences.sections) {
+        console.warn(
+            'The `defaultPreferences`\'s `sections` configuration is deprecated and will be removed in the future. Please use `sections.[sectionName]` configuration options instead, with the same format as the defaultPreferences.'
+        );
+        Object.keys(defaultPreferences.sections).forEach((appName) => {
+            Object.keys(defaultPreferences.sections[appName]).forEach((sectionName) => {
+                _merge(newConfig, {
+                    sections: { [sectionName]: defaultPreferences.sections[appName][sectionName] }
+                });
+            });
+        });
+    }
+
     setProjectConfigurationCommon(newConfig);
 };
 
--- a/packages/transition-frontend/src/components/dashboard/TransitionMenuBar.tsx
+++ b/packages/transition-frontend/src/components/dashboard/TransitionMenuBar.tsx
@@ -9,7 +9,7 @@ import { WithTranslation, withTranslation } from 'react-i18next';
 
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import { LayoutSectionProps } from 'chaire-lib-frontend/lib/services/dashboard/DashboardContribution';
-import Preferences from 'chaire-lib-common/lib/config/Preferences';
+import config from 'chaire-lib-common/lib/config/shared/project.config';
 
 // TODO Menu items should not be provided directly by widgets, it should be
 // built from descriptive elements, or contributions should register their own
@@ -17,7 +17,7 @@ import Preferences from 'chaire-lib-common/lib/config/Preferences';
 const MenuBar: React.FunctionComponent<LayoutSectionProps & WithTranslation> = (
     props: LayoutSectionProps & WithTranslation
 ) => {
-    const sectionsConfig = Preferences.get('sections.transition');
+    const sectionsConfig = config.sections;
 
     const onClickHandler = function (e) {
         e.preventDefault();
--- a/packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
+++ b/packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
@@ -6,6 +6,7 @@
  */
 import React from 'react';
 import Collapsible from 'react-collapsible';
+import moment from 'moment';
 import { WithTranslation, withTranslation } from 'react-i18next';
 import _toString from 'lodash/toString';
 import PreferencesResetToDefaultButton from '../PreferencesResetToDefaultButton';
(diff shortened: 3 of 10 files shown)


tahini:
@GabrielBruno24 dépendamment de l'ordre d'entrée des PRs, il va falloir faire des changement soit à celle-ci, soit à celle de la comparaison de scénario pour pouvoir tenir compte de la nouvelle section.

packages/chaire-lib-backend/src/config/server.config.ts
```diff
@@ -405,6 +405,19 @@ export const setProjectConfiguration = (newConfig: Partial<ProjectConfiguration<
         }
     }
 
+    // If there are still section configurations in the defaultPreferences section of the project config, set it as section config.
+    // FIXME Remove this if a reasonable time after march 2025, to give time to instances to update their configuration
+    const defaultPreferences = (newConfig as any)?.defaultPreferences;
+    if (defaultPreferences && defaultPreferences.sections) {
```
greenscientist, line 411: Can we have a console.warn() here that says that we are using a deprecated feature. 
(To help make sure we have migrated everything)

packages/transition-frontend/webpack.config.js
```diff
@@ -18,6 +18,9 @@ if (!process.env.NODE_ENV) {
     process.env.NODE_ENV = 'development';
 }
 
+// Make sure Transition specific config are loaded
+require('transition-common/lib/config/project.config');
```
greenscientist, line 22: isn't a bad idea to expose the project config directly into the frontend???

greenscientist (changes requested):

kaligrafy (approved):

packages/transition-frontend/webpack.config.js
tahini: > isn't a bad idea to expose the project config directly into the frontend???

yeah, and we've known it for a while now: #424 At least we [try to] keep in the backend the more sensible stuff.

greenscientist: Should we try to not add more then? :)

packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
```diff
@@ -14,25 +15,28 @@ import { roundToDecimals } from 'chaire-lib-common/lib/utils/MathUtils';
 import InputStringFormatted from 'chaire-lib-frontend/lib/components/input/InputStringFormatted';
 import InputWrapper from 'chaire-lib-frontend/lib/components/input/InputWrapper';
 import InputSelect from 'chaire-lib-frontend/lib/components/input/InputSelect';
+import config from 'chaire-lib-common/lib/config/shared/project.config';
```
greenscientist, line 18: At least, can we have a //TODO with each of these new imports to point to the task about removing them

packages/transition-frontend/src/components/dashboard/TransitionMenuBar.tsx
```diff
@@ -9,15 +9,15 @@ import { WithTranslation, withTranslation } from 'react-i18next';
 
 import serviceLocator from 'chaire-lib-common/lib/utils/ServiceLocator';
 import { LayoutSectionProps } from 'chaire-lib-frontend/lib/services/dashboard/DashboardContribution';
-import Preferences from 'chaire-lib-common/lib/config/Preferences';
+import config from 'chaire-lib-common/lib/config/shared/project.config';
```
greenscientist, line 12: See later comment

packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
tahini: This will never be removed, as we still need the list of enabled sections to list them for the user to select the default section. Same applies to the other comment

packages/chaire-lib-backend/src/config/server.config.ts
tahini: done

tahini:
> So I imagine one of the original intent was to allow users to hide some of the section if they did not want it. Granted, we don't have a UI to let users do that at the moment. And it's good to first have the server config to let a base sélection of which section is available. Do we have an issue to let users hide some section if they want ?

I haven't found such an issue, as nobody asked for it. And for now, we are interested in instance-enabled sections (like simulation is disabled by default, and the calculation instance only enables the calculation sections).

If we ever implement user-enablement/disablement of sections, it will be a different option as a user should not be able to enable a section not available in the instance. An issue that *should* be opened now, though, is to make sure that the UI cannot "accidentally" enable a section that the instance does not "support". (but it was already the case, no regression here)

packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
greenscientist: This should be served by an API to the frontend then, not as an hardcoded list for that purpose ?

greenscientist (approved):
I'll accept. 

But I think we should rework the preference part to better handle things to/from the backend

packages/transition-frontend/src/components/forms/preferences/sections/PreferencesSectionGeneral.tsx
greenscientist: Sure, it's an improvement, but I will not miss a chance ask for it. It might end up being done at some point... :)

greenscientist: And yes conditional webpack compile might be a good way to do. I just don't like have just a generic import and having every bit of the config easily available... We never know what we might leak!

tahini: > Sure, it's an improvement, but I will not miss a chance ask for it. It might end up being done at some point... :)

There's an issue for it now fwiw!

PR #1300, opened 2025-03-17
dep: Update `@babel/*` packages

tahini:
Delete all the `@babel/*` data from `yarn.lock` and run `yarn` again to get the latest versions. This should fix a few dependabot issues.


greenscientist (approved):
Et y'a moins de dépendences!

PR #1301, opened 2025-03-19
Add a couple more continue if not found flag in the poi import section

greenscientist:


GabrielBruno24 (approved):
You should add a description to the PR and the commit message but I approve otherwise.

greenscientist:
@GabrielBruno24 what more would you like to see in a description?

GabrielBruno24:
> @GabrielBruno24 what more would you like to see in a description?

Just explain that the task failed due to edge cases on a particular dataset and so you changed the flags so it can continue.

Issue #725, opened 2023-10-03
Enter key in login screen must trigger the email confirm button


millette:
Vient de https://github.com/chairemobilite/evolution/issues/232

PR #1302, opened 2025-03-24
Fix bug with passwordless login form when pressing 'enter' or 'space'.

samuel-duhaime:
# Pull request
fix: #725 
Close https://github.com/chairemobilite/transition/pull/726 when this is merged. It's a duplicate of an old PR.

## Description
Fix bug with passwordless login form when pressing 'enter' or 'space'.
Remove onKeyUp on the button because the default button already handles that. On the email input, put onKeyUp with handleKeyUp.
So now, the user can press 'enter' or 'space' directly from the email input and go to the survey directly. Remove e.witch that is deprecated. The new method is e.key.


greenscientist (approved):
lgtm

greenscientist:
Something to put on the stack of automated test when we have them :) CC @GabrielBruno24

Issue #1142, opened 2024-12-13
Possible double login caused by dual onKeyUp/onKeyPressed


kaligrafy:
We have a onKeyUp on some submit buttons and also on the form itself (onKeyPressed), so this can cause issues with double clicking. See LoginForm for example. This needs cleaning/testing.

samuel-duhaime:
Actually, onKeyUp is not needed on the Button input. The default button is enough to handle these cases. I just fixed one of these problems with the PR https://github.com/chairemobilite/transition/pull/1302. 

We should use the same method (remove onKeyUp from the Button and put it on the input).

PR #1304, opened 2025-03-24
Fix double submit when 'Enter' press inside the form.

samuel-duhaime:
# Pull request
fix: #1142 

## Description
Fix double submit when 'Enter' press inside the form.
We remove onKeyUp inside a form element because this could send the form twice when clicking 'enter' with the submit button focus. 
We move onKeyUp on the last input of the form so the user can submit the form when pressing 'enter' from there. 
We did this correction for LoginForm, RegisterForm, and PwdLessLoginForm. 
There were the same errors for all of them.
Update the packages/chaire-lib-frontend/src/components/pages/__tests__/__snapshots__/LoginPage.test.tsx.snap snapshot with buttons with type 'submit'

### DIFF
--- a/packages/chaire-lib-frontend/src/components/forms/auth/localLogin/LoginForm.tsx
+++ b/packages/chaire-lib-frontend/src/components/forms/auth/localLogin/LoginForm.tsx
@@ -69,20 +69,15 @@ const LoginPage: React.FC<LoginPageProps> = ({ withForgotPassword = false, heade
         );
     };
 
-    const handleKeyPress = (e: React.KeyboardEvent<HTMLFormElement>) => {
-        if (e.key === 'Enter' || e.which === 13) {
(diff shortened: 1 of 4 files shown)


greenscientist:
@samuel-duhaime , to confirm, the enter behavior will properly be: if enter is pressed in the first fields, it will go to the next, until the last one, if we press enter there, the form will be submitted?

samuel-duhaime:
> @samuel-duhaime , to confirm, the enter behavior will properly be: if enter is pressed in the first fields, it will go to the next, until the last one, if we press enter there, the form will be submitted?

No, if you press 'enter' on the first fields, nothing will happen. Only 'tab' will take you to the next field. However, if you press 'enter' on the last field, the form will submit.

greenscientist:
ok. I guess it's ok. . I'll tag @kaligrafy so he double check when he's back

greenscientist (approved):

PR #1306, opened 2025-03-25
Compare accessibility maps

GabrielBruno24:
image

Adds the new accessibility map comparison tab, which allows you to visualize two accessibility maps that use different scenarios at the same time.

Also applies two fixes to the routing comparison tab:
1- Limit the displayed name of the chosen scenarios to two lines
2- Allow all inputted parameters in the form to affect both routing engines used to calculate the two paths (for example, changing the starting time would not update the starting time for the scenario 2 path)


packages/transition-frontend/src/components/forms/comparison/ScenarioComparisonPanel.tsx
```diff
@@ -295,7 +295,11 @@ class ScenarioComparisonPanel extends ChangeEventsForm<ComparisonPanelProps, Com
     private getScenarioNameById(id: string): string {
         for (const scenario of this.state.scenarioCollection.features) {
             if (scenario.id === id) {
-                return scenario.toString(false);
+                let scenarioName = scenario.toString(false) as string;
+                if (scenarioName.length >= 30) {
```
greenscientist, line 302: Can you add a comment that says something like: 
// We are not simply doing a substr(0,30) as we want to add "..." when we have to cut it down

greenscientist, line 302: Should the 30 be configurable ? (You probably want to put it in a constant anyway to make sure we use the same one in the if and in the substr in the next line)

packages/transition-frontend/src/components/forms/comparison/ScenarioComparisonPanel.tsx
```diff
@@ -305,13 +308,25 @@ class ScenarioComparisonPanel extends ChangeEventsForm<ComparisonPanelProps, Com
...
+    };
+
     render() {
         if (!this.state.scenarioCollection) {
             return <LoadingPage />;
         }
 
         const routing = this.state.object;
-        const alternateRouting = this.state.alternateScenarioRouting;
+        const altRouting = this.state.alternateScenarioRouting;
```
greenscientist, line 329: Why reduce the variable name ? Its usually better to keep full names, character space is cheap thse days :)

packages/transition-backend/src/api/services.socketRoutes.ts
```diff
@@ -151,6 +151,48 @@ export default function (socket: EventEmitter, userId?: number) {
         }
     );
 
+    socket.on(
```
greenscientist, line 154: Can we have comment describing those new routes?

packages/transition-backend/src/api/services.socketRoutes.ts
```diff
@@ -151,6 +151,48 @@ export default function (socket: EventEmitter, userId?: number) {
...
+        'accessibiliyMap.getIntersection',
```
greenscientist, line 155: Do we want to have those 2 call separated or they would always be called together ??

Also would it be better to have a flag for when we request the data initially to have the accessbility map already compared?
What I mean, it seem that we first need to do a request to the backend, get some maps and then we send those maps to the backend again to get a different. 
Would be it better to have a call that request accessbilityMapCompairaison with both scenario as a param and return directly the full set of data? 
I might not be right, but just an idea to think about...

packages/transition-frontend/src/services/routing/RoutingUtils.ts
```diff
@@ -86,3 +86,43 @@ export const calculateAccessibilityMap = async (
         );
     });
 };
+
+export const calculateMapIntersection = async (
```
greenscientist, line 90: Is this really belong to a RoutingUtil object ? 
See my previous comment, since this might not be the best flow for this.

greenscientist (commented):
The code seems quite devoid of comment, we probably need a bit more. 

I'll set @samuel-duhaime comment on the React frontend stuff...

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -10,7 +10,9 @@ import {
     multiLineString as turfMultiLineString,
     circle as turfCircle,
     area as turfArea,
-    polygonToLine as turfPolygonToLine
+    polygonToLine as turfPolygonToLine,
```
samuel-duhaime, line 13: Small note for the future, we might be overusing 'as'. This is a strong operation in TypeScript that will override the type (for the better and the worse), if we don't pay attention, it could lead to a false sense of security. Also, I'm not sure why turfPolygonToLine doesn't start with a capital letter.

We are better off typing something like this. 
`const polygonToLine: TurfPolygonToLine`

But in this case, all the imports are like that, so it's okay.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -253,6 +255,101 @@ export class TransitAccessibilityMapCalculator {
...
+            const difference = turfDifference(polygons) as Feature<MultiPolygon> | null;
+            if (difference === null) {
+                resolve(null);
+            }
+
+            let polygonStroke = turfPolygonToLine(difference!);
+            if (polygonStroke.type === 'Feature') {
+                polygonStroke = turfFeatureCollection([polygonStroke]);
+            }
+            const polygonStroke2 = turfFeatureCollection([]) as FeatureCollection<GeoJSON.LineString>;
```
samuel-duhaime, line 272: Don't use 'as'. Use `const polygonStroke2: Fea...` This will help TypeScript catch any error.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
samuel-duhaime, line 1: Let's remove all occurrences of 'as' in this commit. We should only use' as' when working with 'externate' packages, and it is the only way to compile. When writing our own code, 'as' should not be used. We want TypeScript to catch compilation errors; otherwise, we should just write in JavaScript and assume that everything is fine. 

From someone on Reddit.
> In some ways as is equivalent to any- you're telling the TypeScript compiler to trust that you know information about a type that it was not able to statically infer.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const AccessibilityComparisonStatsComponent: React.FunctionComponent<AccessibilityComparisonStatsComponentProps> = (
+    props: AccessibilityComparisonStatsComponentProps
+) => {
+    const { t } = useTranslation(['transit', 'main']);
+
+    const sortByDuration = (features: GeoJSON.Feature[]) => {
+        features.sort((feat1, feat2) => {
+            const duration1 = feat1.properties?.durationMinutes;
+            console.log(duration1);
```
samuel-duhaime, line 26: Do we still want this console.log?

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const features2 = props.accessibilityPolygons.result2.features;
+    sortByDuration(features2);
+
+    const intersections = props.accessibilityPolygons.intersection.features;
+    sortByDuration(intersections);
+
+    const combinedFeatures: { result1: GeoJSON.Feature; result2: GeoJSON.Feature; intersection: GeoJSON.Feature }[] =
+        [];
+
+    for (let i = 0; i < features1.length; i++) {
```
samuel-duhaime, line 46: By the way, it might be my preference, but the forEach method is perfect for this kind of code. 
`features.forEach((feature) > {­­...})`

You don't need to change this if you don't want to.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
```diff
@@ -0,0 +1,157 @@
...
+        combinedFeatures.push({
+            result1: features1[i],
+            result2: features2[i],
+            intersection: intersections[i]
+        });
+    }
+
+    const featureStats = combinedFeatures.map((features) => {
```
samuel-duhaime, line 54: For `combinedFeatures.map((features)`, are there many 'features' or just one 'feature' ? Normally, it should be 'feature'.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
```diff
@@ -0,0 +1,157 @@
...
+        const properties1 = features.result1.properties;
+        const properties2 = features.result2.properties;
+        const intersection = features.intersection.properties;
+        if (!properties1 || !properties2 || !intersection) return null;
+
+        // TODO: POI import does not work correctly right now, so there is no point in having this section at the moment.
```
samuel-duhaime, line 60: It should be a FIXME.

greenscientist: Not sure we have an agreement/standard here. I personnally prefer and use TODO

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
GabrielBruno24: Use of 'as' is necessary here since the turf functions can return multiple types. We use the if else to treat the different types separately, but we have no way to check the types such that vscode will understand without the as.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
GabrielBruno24: Since each item in the array is a combination of three features I think it should be kept plural.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
GabrielBruno24: Never really though about it, I think I've used them interchangeably in the past.

packages/transition-backend/src/api/services.socketRoutes.ts
GabrielBruno24: I combined the intersection and difference calls into one, that way we can remove the loop in the frontend as well. That said I decided to keep the initial calls for the single accessibility maps separate. We do some intermediary steps that alter and depend on the state of the react component and if I combined all three calls it would lead to a lot of arguments and a bloated return type, so I thought it was more trouble than it was worth. We still receive the maps only to send them back to the backend immediatly after but I feel that's a necessary evil, and could be useful in the future if we ever want to do something like toggle the visibility of one or the other.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -294,6 +353,61 @@ export class TransitAccessibilityMapCalculator {
...
+
+            let intersection = await this.getPolygonsIntersection(combinedPolygons, 'rgba(47, 138, 243, 0.4)');
+            if (intersection === null) {
+                intersection = {
+                    polygons: turfFeatureCollection([]),
+                    strokes: turfFeatureCollection([])
+                };
+            }
+
+            let difference1 = await this.getPolygonsDifference(combinedPolygons, 'rgba(255, 0, 0, 0.4)');
```
samuel-duhaime, line 378: let difference1 et let difference 2 aurait intéret à être dans une fonction (pour aider à maintenir).
let difference 1 = getDifference(...)

packages/transition-frontend/src/components/forms/comparison/ScenarioComparisonAlternativesSelect.tsx
```diff
@@ -113,11 +113,22 @@ const AlternativesSelect: React.FunctionComponent<AlternativesSelectProps> = (pr
...
+                    <div>
+                        {t('transit:transitComparison:ScenarioAlternatives', {
+                            scenarioNumber: '1'
+                        }) + ','}
+                    </div>
+                    {/* If the scenario's name takes 3 lines or more, it will truncate itself with an ellipsis. */}
+                    <div
+                        style={{
+                            overflow: 'hidden',
+                            display: '-webkit-box',
```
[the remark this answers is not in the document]
samuel-duhaime: And also, you use it 2 times.

samuel-duhaime (dismissed):

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
```diff
@@ -0,0 +1,685 @@
...
+        const routing = this.state.object;
+        const maxDuration = routing.attributes.maxTotalTravelTimeSeconds;
+        const numberOfPolygons = routing.attributes.numberOfPolygons;
+        if (typeof numberOfPolygons !== 'number' || numberOfPolygons <= 1) {
+            this.setState({ displayMaxTimeSelect: false });
+            this.setState({ possibleMaxTimes: [] });
+            return;
+        }
+
+        if (typeof maxDuration !== 'number' || maxDuration < numberOfPolygons * 60) {
```
tahini, line 333: Qu'est-ce que numberOfPolygons * 60?

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
tahini: Then something might need to be renamed, maybe resultFeatures or combinedResultsFeatures

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
tahini: Or maybe just totally drop this code for now and add a TODO to support it, with maybe links to where the base code should be, as this commented code will likely not match the actual repaired code once it is repaired...

packages/transition-frontend/src/config/layers.config.ts
```diff
@@ -57,6 +57,12 @@ export const sectionLayers = {
         'accessibilityMapPolygonStrokes',
         'accessibilityMapPoints'
     ],
+    accessibilityComparison: [
+        'aggregatedOD',
```
tahini, line 61: I think we got rid of the aggregatedOD now. Or is it in the deck-gl branch. Anyway, you might want to remove it from here.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -294,6 +353,61 @@ export class TransitAccessibilityMapCalculator {
...
+        const finalMap: TransitAccessibilityMapWithPolygonAndTimeResult[] = [];
+
+        for (let i = 0; i < numberOfPolygons; i++) {
+            const polygons1 = result1.polygons.features[i];
+            const polygons2 = result2.polygons.features[i];
+
+            const combinedPolygons = turfFeatureCollection([polygons1, polygons2]);
+            const combinedPolygonsReverseOrder = turfFeatureCollection([polygons2, polygons1]);
+
+            let intersection = await this.getPolygonsIntersection(combinedPolygons, 'rgba(47, 138, 243, 0.4)');
```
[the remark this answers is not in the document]
tahini: J'ai quand même une préférence pour que cette fonction soit plutôt un polygonDiff comme dans `diff fichier1 fichier2` avec +, - et égal au lieu de retourner des polygones dont seule la couleur ou leur position dans l'array permet de savoir ce que ça représente exactement. Les consommateurs de cette fonction pourrait ensuite faire ce qu'ils veulent avec le résultat pour l'afficher et/ou faire des calculs, mais au moins, la réponse au retour de cette fonction sera claire.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -251,6 +261,55 @@ export class TransitAccessibilityMapCalculator {
...
     }
 
+    private static async getPolygonsDifference(
+        polygons: FeatureCollection<MultiPolygon>,
+        color: string
+    ): Promise<TransitAccessibilityMapWithPolygonResult | null> {
+        return new Promise((resolve) => {
+            const difference = turfDifference(polygons) as Feature<MultiPolygon> | null;
+            if (difference === null) {
+                resolve(null);
```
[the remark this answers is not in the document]
GabrielBruno24: Didn't know that about resolve, I thought it was equivalent to return inside an promise. You learn something knew every day.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -251,6 +261,55 @@ export class TransitAccessibilityMapCalculator {
         }
     }
 
+    private static async getPolygonsDifference(
+        polygons: FeatureCollection<MultiPolygon>,
+        color: string
+    ): Promise<TransitAccessibilityMapWithPolygonResult | null> {
+        return new Promise((resolve) => {
+            const difference = turfDifference(polygons) as Feature<MultiPolygon> | null;
```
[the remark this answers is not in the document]
GabrielBruno24: I'm not sure what happens when there are more than two polygons, the turf documentation does not say. I could throw an error if it isn't 2.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
GabrielBruno24: C'est parce que le temps est en secondes, et on veut pas que le nombre de polygones excede le nombre de minutes. Mais ça fait plus de sens de diviser le temps par 60 a la place.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonStatsComponent.tsx
GabrielBruno24: I renamed featureStats as combinedFeaturesStatsTable

packages/transition-frontend/src/config/layers.config.ts
GabrielBruno24: We still have aggregatedOD on the other layers. However it looks like it doesn't do anything when I remove it for accessibilityComparison so I removed it.

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
tahini: > C'est parce que le temps est en secondes, et on veut pas que le nombre de polygones excede le nombre de minutes. J'avais copié ça d'ailleurs dans le code mais effectivement ça fait plus de sens de diviser le temps par 60 a la place.

Alors ça vaudrait la peine de mettre un commentaire pour expliquer cela

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -294,6 +353,61 @@ export class TransitAccessibilityMapCalculator {
...
 
+    static async getMapComparison(
+        result1: TransitAccessibilityMapWithPolygonResult,
+        result2: TransitAccessibilityMapWithPolygonResult,
+        numberOfPolygons: number
+    ): Promise<TransitAccessibilityMapWithPolygonAndTimeResult[]> {
+        const finalMap: TransitAccessibilityMapWithPolygonAndTimeResult[] = [];
+
+        for (let i = 0; i < numberOfPolygons; i++) {
+            const polygons1 = result1.polygons.features[i];
```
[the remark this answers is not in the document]
GabrielBruno24: Le truc c'est que il y a pas vraiment de "base" et "comparateur". Les deux polygons ont le même statut, et si tu échange les scénarios, ca va juste changer la couleur de la carte et le signe des nombres dans la 3eme collone des résultats.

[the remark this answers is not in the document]
tahini: > Le truc c'est que il y a pas vraiment de "base" et "comparateur". Les deux polygons ont le même statut, et si tu échange les scénarios, ca va juste changer la couleur de la carte et le signe des nombres dans la 3eme collone des résultats.

Pas vraiment. En rouge, c'est la portion qui était dans le polygon de base et qui n'est plus dans le polygone comparateur et en vert, c'est ce qu'il y a de nouveau dans le 2e polygone (comparateur), comme dans un PR sur github. Si ce n'est pas le cas, ce sera pas clair, donc le scénario 1 est le baseline et le scénario 2 est celui avec lequel on compare.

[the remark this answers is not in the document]
greenscientist: Hum, est-ce qu'on est sur que c'est ça qu'on veut?  Ca revient peut-etre au même, mais je ne voyait pas ça comme un diff.  (et nos utilisateurs ne vont pas avoir la référence de voir des diff non plus).  Je voyais ça comme étant une zone commune et ensuite les zone exclusive au scénario 1 et au scénario 2. Probablement que choisir du rouge et du vert est une mauvaise idée.  (En fait, c'est doublement une mauvaise idée, car les colors blind ne vont voir que la meme couleur)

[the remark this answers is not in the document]
tahini: haha, alors c'est clair que c'est pas clair, car je voyais ça totalement comme un diff, comme un differential flamegraph ou qqc du genre. Mais comme discuté sur Teams avec Gabriel, de la façon que je voie ça, il y a le calcul et la présentation du résultat. 

Le calcul se fait comme un diff: il y a une portion commune, une portion enlevée (dans le premier, mais pas dans le 2e) et une portion ajoutée (dans le 2e et pas le premier).

La présentation utilise cette information comme elle veut. Si on ne veut pas présenter ça comme un diff, il faut juste que ce soit très clair ce que chaque couleur signifie. Et si c'est pas présenté comme un diff, ça ne doit surtout pas être rouge et vert! Pour des palettes de couleurs, voir https://colorbrewer2.org/ et choisir les paramètres appropriés.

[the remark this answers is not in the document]
GabrielBruno24: J'ai change les couleurs. Dans le future il faudrait bouger ca dans les preferences.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -458,4 +550,36 @@ export class TransitAccessibilityMapCalculator {
...
+        const polygonStroke2: FeatureCollection<LineString> = turfFeatureCollection([]);
+
+        for (let i = 0, countI = polygonStroke.features.length; i < countI; i++) {
+            const feature = polygonStroke.features[i];
+            if (feature.geometry.type === 'MultiLineString') {
+                // this is a polygon with hole, we need to separate into two LineStrings.
+                for (let j = 1, countJ = feature.geometry.coordinates.length; j < countJ; j++) {
+                    polygonStroke2.features.push(turfLineString(feature.geometry.coordinates[j]));
+                }
+                // TODO Copied from original code, but should it be .push instead of features[i] ?
```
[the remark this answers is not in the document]
GabrielBruno24: C'était un bug effectivement, j'ai fixé et rajouté un unit test.

packages/transition-common/src/services/accessibilityMap/TransitAccessibilityMapResult.ts
```diff
@@ -27,7 +27,18 @@ export interface TransitAccessibilityMapResult {
...
-    resultByNode: TrRoutingResultAccessibilityMap | undefined;
+    resultByNode?: TrRoutingResultAccessibilityMap;
+}
+
+export interface TransitAccessibilityMapComparisonResult {
+    polygons: GeoJSONComparison<GeoJSON.MultiPolygon>;
+    strokes: GeoJSONComparison<GeoJSON.MultiLineString>;
+}
+
+interface GeoJSONComparison<T extends GeoJSON.MultiPolygon | GeoJSON.MultiLineString> {
```
tahini, line 38: Juste MultiPolygon ou ça pourrait être Polygon aussi?

packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
```diff
@@ -0,0 +1,718 @@
...
+        newValue: { value: any; valid?: boolean },
+        alternateRouting: TransitAccessibilityMapRouting
+    ) => {
+        this.onValueChange(path, newValue);
+        if (newValue.valid || newValue.valid === undefined) {
+            alternateRouting.attributes[path] = newValue.value;
+        }
+    };
+
+    private convertToRGBA = (rgbValue: string, alpha: number) => {
```
tahini, line 393: Je pourrais dire qqc, mais tout ça va changer dans la branche deck.gl et demandera refactoring de toute façon, alors je dis rien...

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -294,6 +371,47 @@ export class TransitAccessibilityMapCalculator {
         }
     }
 
+    static async getMapComparison(
+        result1: FeatureCollection<MultiPolygon>,
+        result2: FeatureCollection<MultiPolygon>,
+        numberOfPolygons: number,
+        colors: TransitMapColorOptions
```
tahini, line 378: Je persiste à penser que cette fonction dans le backend n'a pas à avoir de notion de couleur, mais bon, on peut fixer ça plus tard.
D'ailleurs le numberOfPolygons devrait venir de soi, ie juste traiter le nombre minimal contenu dans chacun des features collections. ça enlèverait 2 paramètres à cette fonction, mais ça peut être une autre amélioration subséquente.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
```diff
@@ -294,6 +371,47 @@ export class TransitAccessibilityMapCalculator {
...
+                colors.scenario2Minus1Color
+            );
+
+            finalMap.push({
+                polygons: {
+                    intersection: intersection.polygons.features,
+                    scenario1Minus2: scenario1Minus2.polygons.features,
+                    scenario2Minus1: scenario2Minus1.polygons.features
+                },
+                strokes: {
```
tahini, line 404: À voir avec deck.gl, est-ce qu'on aura encore besoin des strokes? On a éliminé la couche stroke des trajets, peut-être qu'on peut le faire aussi avec les polygones, mais c'est un problème de deck.gl..

tahini (approved):
Good enough pour rendre disponible aux utilisateurs

packages/transition-common/src/services/accessibilityMap/TransitAccessibilityMapResult.ts
GabrielBruno24: Non, les cartes d'accessibilité sont toujours retournées en multipolygons.

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
GabrielBruno24: J'envoie les couleurs parce que les fonctions turf d'intersection et de soustraction ont une fonctionalité qui permet de passer des propriétés aux polygones retournés, et je trouvait ça plus clean que de faire ça manuellement coté frontend. Pour numberOfPolygons... pas d'excuses, je sais pas pourquoi je l'ai fait comme ça. Je peux m'en occuper après le merge.

greenscientist: Ouais, un peu d'accord avec ce commentaires sur les couleurs

packages/transition-backend/src/services/accessibilityMap/TransitAccessibilityMapCalculator.ts
GabrielBruno24: Visuellement ça serait mieux, non? Est-ce que deck.gl gènère des outlines visuelles tout seul?

greenscientist (changes requested):
We need proper source, licence information about the icon. 

We probably need this in the commit message and a README file in the icon directory to note the details, specially if we need attribution.

greenscientist (approved):
THanks for adding the source information

Issue #1065, opened 2024-09-23
Add Playwright to transition


GabrielBruno24:
Make transition able to run playwright tests, and add a few simple ones.

PR #1308, opened 2025-03-26
Add Playwright

GabrielBruno24:
Add Playwright to Transition so that we can run UI tests, and add tests to log in and navigate through all the tabs.
Fix: #1065

### DIFF
--- a/.github/workflows/transition.yml
+++ b/.github/workflows/transition.yml
@@ -33,8 +33,6 @@ jobs:
       run: yarn build:prod
     - name: Unit Test
       run: yarn test
-    - name: UI Test
-      run: yarn test:ui
   
   code-lint:
     runs-on: ubuntu-latest
--- a/packages/transition-frontend/package.json
+++ b/packages/transition-frontend/package.json
@@ -18,7 +18,7 @@
     "test": "cross-env NODE_ENV=test jest --config=jest.config.js",
     "test:unit": "cross-env NODE_ENV=test jest --config=jest.config.js",
     "test:sequential": "echo 'cross-env NODE_ENV=test jest --config=jest.sequential.config.js --runInBand'",
-    "test:ui": "echo 'cross-env NODE_ENV=test jest --config=jest.ui.config.js'",
+    "test:ui": "LOCALE_DIR=$(pwd)/locales npx playwright test",
     "lint": "eslint .",
     "format": "prettier-eslint $PWD/'src/**/*.{ts,tsx}' --write"
   },
@@ -66,6 +66,7 @@
   },
   "devDependencies": {
     "@alienfast/i18next-loader": "^2.0.2",
+    "@playwright/test": "^1.50.0",
     "@testing-library/react": "^16.1.0",
     "@types/geojson": "^7946.0.15",
     "@types/jest": "^29.5.14",
--- a/packages/transition-frontend/playwright-example.config.ts
+++ b/packages/transition-frontend/playwright-example.config.ts
@@ -1,0 +1,81 @@
+import dotenv from 'dotenv';
+import path from 'path';
+import { defineConfig, devices } from '@playwright/test';
+
+dotenv.config({ path: path.resolve(__dirname, '../../.env') });
+
+export default defineConfig({
+  testDir: './ui-tests',
+  /* Run tests in files in parallel */
+  fullyParallel: true,
+  /* Fail the build on CI if you accidentally left test.only in the source code. */
+  forbidOnly: !!process.env.CI,
+  /* Retry on CI only */
+  retries: process.env.CI ? 2 : 0,
+  // Each test is given 15 seconds.
+  timeout: 15000,
+  /* Opt out of parallel tests on CI. */
+  workers: process.env.CI ? 1 : undefined,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter:
+    process.env.CI === 'true'
+        ? [['list', { printSteps: true }]]
+        : [['html', { outputFolder: 'playwright-report', open: 'always' }]],
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: 'http://localhost:8080',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+    screenshot: 'only-on-failure'
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    // {
+    //   name: 'chromium',
+    //   use: { ...devices['Desktop Chrome'] },
+    // },
+    {
+        name: 'Google Chrome',
+        use: { ...devices['Desktop Chrome'], channel: 'chrome' }
+    }
+    // {
+    //   name: 'firefox',
+    //   use: { ...devices['Desktop Firefox'] },
+    // },
+
+    // {
+    //   name: 'webkit',
+    //   use: { ...devices['Desktop Safari'] },
+    // },
+
+    /* Test against mobile viewports. */
+    // {
+    //   name: 'Mobile Chrome',
+    //   use: { ...devices['Pixel 5'] },
+    // },
+    // {
+    //   name: 'Mobile Safari',
+    //   use: { ...devices['iPhone 12'] },
+    // },
+
+    /* Test against branded browsers. */
+    // {
+    //   name: 'Microsoft Edge',
+    //   use: { ...devices['Desktop Edge'], channel: 'msedge' },
+    // },
+    // {
+    //   name: 'Google Chrome',
+    //   use: { ...devices['Desktop Chrome'], channel: 'chrome' },
+    // },
+  ],
+
+  /* Run your local dev server before starting the tests */
+  // webServer: {
+  //   command: 'npm run start',
+  //   url: 'http://127.0.0.1:3000',
+  //   reuseExistingServer: !process.env.CI,
+  // },
+});
--- a/packages/transition-frontend/ui-tests/login-tests.spec.ts
+++ b/packages/transition-frontend/ui-tests/login-tests.spec.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2025, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { test } from '@playwright/test';
+import * as testHelpers from './testHelpers';
+import * as loginTestHelpers from './loginTestHelpers';
+
+const context = {
+    page: null as any,
+    widgetTestCounters: {}
+};
+
+// Configure the tests to run in serial mode (one after the other)
+test.describe.configure({ mode: 'serial' });
+
+test.beforeAll(async ({ browser }) => {
+    context.page = await testHelpers.initializeTestPage(browser);
+});
+
+// Login using all methods enabled in the transition login page
+// We switch to english only on the first login, since the language is kept after that
+loginTestHelpers.startAndLoginAnonymously({ context, title: 'Transition', loginMethod: 'clickLoginButton', changeToEnglish: true });
+testHelpers.logoutTest({ context });
+loginTestHelpers.startAndLoginAnonymously({ context, title: 'Transition', loginMethod: 'enterOnPasswordField', changeToEnglish: false });
+testHelpers.logoutTest({ context });
+loginTestHelpers.startAndLoginAnonymously({ context, title: 'Transition', loginMethod: 'enterOnUsernameField', changeToEnglish: false });
+testHelpers.logoutTest({ context });
--- a/packages/transition-frontend/ui-tests/loginTestHelpers.ts
+++ b/packages/transition-frontend/ui-tests/loginTestHelpers.ts
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2025, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import * as testHelpers from './testHelpers';
+import * as languageTestHelpers from './languageTestHelpers';
+
+/**
+ * Test the login page, change to the right language, and login to the test account.
+ * @param {Object} options - The options for the test.
+ * @param {string} options.title - The title of the page.
+ * @param {testHelpers.LoginMethods} options.loginMethod - The method used to log in. By default, presses the login button.
+ * @param {boolean} options.changeToEnglish - Whether to change the language to english or not. Used in tests that require multiple logins, as the page will still be in english after the first.
+ */
+export const startAndLoginAnonymously = ({
+    context,
+    title,
+    loginMethod = 'clickLoginButton',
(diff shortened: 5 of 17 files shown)


.env.example
```diff
@@ -70,3 +70,8 @@ MAIL_TRANSPORT_SMTP_AUTH_PWD=password
 
 # From email
 MAIL_FROM_ADDRESS=example@example.com
+
```
greenscientist, line 73: I don't think we should put those in the main example file. We should put them elsewhere in another file or in the env definition of the github workflow.

.github/workflows/transition.yml
```diff
@@ -12,11 +12,32 @@ on:
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
```
greenscientist, line 14: I think we should create a separate workflow file for the playwright test. This add a lot of things to the main file which already contains several tests. it makes it unclear what is essential for which part of the test.

README.md
```diff
@@ -165,6 +165,38 @@ For example, to run yarn test in the chaire-lib backend:
 You can also run the app this way with:
 `docker run -a STDOUT -it -v "${PWD}:/home/project" -w=/home/project/ testtransition yarn start`
 
+### UI testing with Playwright
```
greenscientist, line 168: THis is the docker section, we probably need to put this into it's own section. (Maybe even more to a different file, this is developpement specific, maybe in the contributing file. We'd need to check where we have the most developpement related instructions.

examples/config_github.js
```diff
@@ -0,0 +1,181 @@
+// This config file is used during the github pipeline
```
greenscientist, line 1: This could go into the .github directory or into a test directory with the playwirght stuff. 

For this test file, you could removed the commented out stuff.

packages/transition-frontend/package.json
```diff
@@ -18,7 +18,7 @@
     "test": "cross-env NODE_ENV=test jest --config=jest.config.js",
     "test:unit": "cross-env NODE_ENV=test jest --config=jest.config.js",
     "test:sequential": "echo 'cross-env NODE_ENV=test jest --config=jest.sequential.config.js --runInBand'",
-    "test:ui": "echo 'cross-env NODE_ENV=test jest --config=jest.ui.config.js'",
+    "test:ui": "LOCALE_DIR=$(pwd)/locales npx playwright test",
```
greenscientist, line 21: I'm gessing we are removing the old UI test?

packages/transition-frontend/playwright-example.config.ts
```diff
@@ -0,0 +1,78 @@
+import dotenv from 'dotenv';
```
greenscientist, line 1: With a name with example, I'm guessing this file is not really meant to be used and we have another actual file for the test?

packages/transition-frontend/ui-tests/login-tests.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+};
+
+// Configure the tests to run in serial mode (one after the other)
+test.describe.configure({ mode: 'serial' });
+
+test.beforeAll(async ({ browser }) => {
+    context.page = await testHelpers.initializeTestPage(browser);
+});
+
+// Login using all available methods
```
greenscientist, line 25: COmment is a bit missleading, we have several other login method available in chaire-lib :)  
Maybe more details in the comment to say that it's for the method enabled in transition?

packages/transition-frontend/ui-tests/loginTestHelpers.ts
```diff
@@ -0,0 +1,35 @@
...
+ */
+export const startAndLoginAnonymously = ({
+    context,
+    title,
+    loginMethod = 'clickLoginButton',
+    changeToEnglish = true
+}: { title: string, loginMethod?: testHelpers.LoginMethods, changeToEnglish?: boolean } & testHelpers.CommonTestParameters) => {
+    testHelpers.hasTitleTest({ context, title });
+    testHelpers.hasUrlTest({ context, expectedUrl: '/login' });
+    if (changeToEnglish) {
```
greenscientist, line 25: This would need a comment of why it's part for a login test.

packages/transition-frontend/ui-tests/test-left-menu.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+test.describe.configure({ mode: 'serial' });
+
+test.beforeAll(async ({ browser }) => {
+    context.page = await testHelpers.initializeTestPage(browser);
+});
+
+loginTestHelpers.startAndLoginAnonymously({ context, title: 'Transition' });
+
+// Click all the sections in the left menu and check that the right panel becomes the correct one. 
+testHelpers.clickLeftMenuTest({ context, section: 'nodes' });
```
greenscientist, line 28: Would this break every time we add a new section ?

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -0,0 +1,226 @@
...
+    test(`Click the ${section} section of the left menu - ${getTestCounter(context, `${section}`)}`, async () => {
+        //test.setTimeout(60000);
+        const leftMenu = context.page.locator("//nav[@id='tr__left-menu']/ul[@class='tr__left-menu-container']");
+        const sectionButton = leftMenu.locator(`//li/button[@data-section='${section}']/span/img`);
+        await sectionButton.click();
+        const rightPanel = context.page.locator("//section[@id='tr__right-panel']/div[@class='tr__right-panel-inner']");
+        const rightPanelTitle = rightPanel.locator("//h3").nth(0);
+        let expectedRightPanelTitle;
+        switch(section) {
+            case 'agencies':
```
greenscientist, line 190: This part seems a bit hard to maintain as it depends on translation and configuration. Not fully sure if there's a value on checking the whole list.

greenscientist (changes requested):
A few things to change, put in different places.

packages/transition-frontend/playwright-example.config.ts
GabrielBruno24: Yes, the file is copied and renamed to playwright.config.ts. We do the same thing in evolution

packages/transition-frontend/ui-tests/loginTestHelpers.ts
samuel-duhaime: Yes, let's only login in this function. We could test the language switch in another test file.

packages/transition-frontend/ui-tests/loginTestHelpers.ts
```diff
@@ -0,0 +1,35 @@
...
+        testHelpers.isLanguageTest({ context, expectedLanguage: 'fr' });
+        testHelpers.switchLanguageTest({ context, languageToSwitch: 'en' });
+        testHelpers.isLanguageTest({ context, expectedLanguage: 'en' });
+    }
+    testHelpers.loginTest({ context, loginMethod });
+};
+
+export const logout = ({ context }: testHelpers.CommonTestParameters) => {
```
samuel-duhaime, line 33: What is the meaning of this function ? It's already there with testHelpers.logoutTest({ context });. I don't think we need this, it will be harder to maintain everything.

packages/transition-frontend/package.json
GabrielBruno24: There wasn't an old test, the test:ui command in this package would just write "echo 'cross-env NODE_ENV=test jest --config=jest.ui.config.js'" with echo. I'm not sure why actually, and the chaire-lib-frontend package does the same thing. Do you know why, @tahini ?

packages/transition-frontend/ui-tests/test-left-menu.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as loginTestHelpers from './loginTestHelpers';
+
+const context = {
+    page: null as any,
+    title: '',
+    widgetTestCounters: {}
+};
+
+// Configure the tests to run in serial mode (one after the other)
+test.describe.configure({ mode: 'serial' });
```
samuel-duhaime, line 19: I'm wondering if we should really do Transition test in 'serial' mode. 

We did this in Evolution, because the survey is a long succession of actions, but Playwright recommends parallel test, which will be faster to do, and we could test only the part that we want easily. 

@greenscientist Do you have an opinion on that?

samuel-duhaime: We could do a beforeAll that just login inside the application (without any test), and do all action one by one. (We could group some actions together, for example all one panel actions together).

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -0,0 +1,226 @@
...
+ * parameters to add to the URL as query string question.
+ * @param {boolean} options.ignoreHTTPSErrors - Whether to ignore HTTPS errors.
+ * These can happen if running the tests on a remote server with HTTPs (for
+ * example test instances)
+ */
+export const initializeTestPage = async (
+    browser: Browser,
+    options: { urlSearchParams?: { [param: string]: string }, ignoreHTTPSErrors?: boolean } = {}
+): Promise<Page> => {
+    const context = await browser.newContext({ ignoreHTTPSErrors: options.ignoreHTTPSErrors === true });
```
samuel-duhaime, line 53: Can you add a comment why we ignore HTTPS Errors

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -0,0 +1,226 @@
...
+};
+
+// Close the browser after all the tests
+test.afterAll(async ({ browser }) => {
+    browser.close;
+});
+
+const getTestCounter = (context: CommonTestParameters['context'], testKey: string) => {
+    const testIdx = context.widgetTestCounters[testKey] || 0;
+    context.widgetTestCounters[testKey] = testIdx + 1;
```
samuel-duhaime, line 79: Oh, maybe we need a widget test counter. 
@tahini Do you have an opinion on that ?

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -0,0 +1,226 @@
...
+        await expect(context.page).toHaveURL('/login');
+    });
+};
+
+/**
+ * Click on one of the sections on the left menu, and check that the right panel is the correct one.
+ * @param {Object} options - The options for the test.
+ * @param {string} options.section - The section we click.
+ */
+export const clickLeftMenuTest: LeftMenuTest = ({ context, section }) => {
```
samuel-duhaime, line 180: General comment, when possible, it's better to use the locator by role. With this solution, if we change a small 'li' it could break everything. 

```
// Good solution
page.getByRole('button', { name: 'submit' });

// Bad solution
page.locator('button.buttonIcon.episode-actions-later');
```

Source: https://playwright.dev/docs/best-practices#use-locators

yarn.lock
```diff
@@ -1246,6 +1246,13 @@
     "@types/geojson" "^7946.0.7"
     type-fest "^1.2.1"
 
+"@playwright/test@^1.51.1":
```
samuel-duhaime, line 1249: Should we have the same version ("@playwright/test": "^1.50.0" ) as Evolution, for compatibility reasons between Transition and Evolution? I'm not sure. @greenscientist Do you know this?

README.md
GabrielBruno24: There's a "Testing" section in COntributing.md, I'll move it here

.env.example
GabrielBruno24: The user would still need to manually include these in env if they want to do the tests locally so I'm not sure we really want to remove it.

packages/transition-frontend/ui-tests/login-tests.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+import { test } from '@playwright/test';
+import * as testHelpers from './testHelpers';
+import * as loginTestHelpers from './loginTestHelpers';
+
+const context = {
+    page: null as any,
+    title: '',
+    widgetTestCounters: {}
```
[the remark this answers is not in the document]
GabrielBruno24: I think the testCounter is still nice to have in order to avoid duplicate tests, now or in the future.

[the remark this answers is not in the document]
GabrielBruno24: I won't put a type for context in the spec as tests in evolution do not do it either and importing the type from testHelpers leads to a whole mess due to the null.

packages/transition-frontend/ui-tests/testHelpers.ts
GabrielBruno24: I'm not sure why, I just copied that part from the evolution test helpers.

packages/transition-frontend/ui-tests/loginTestHelpers.ts
GabrielBruno24: I decided to keep the language switch in the login function, but made a separate switchFromFrenchToEnglish function. Since we want to test in english to be consistent, I think the language switch should be included by default in the login function. The number of tests where we will need to login multiples times should be low as well, so most of the time we won't have to worry about that parameter.

packages/transition-frontend/package.json
```diff
@@ -40,6 +41,7 @@
     "moment": "^2.30.1",
     "moment-business-days": "^1.2.0",
     "papaparse": "^5.4.1",
+    "playwright": "^1.51.1",
```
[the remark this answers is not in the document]
GabrielBruno24: I removed playwright and only kept playwright/test. As for 1.51.1, I guess it just downloaded the latest version when I added the package.

packages/transition-frontend/ui-tests/testHelpers.ts
GabrielBruno24: I agree, but when searching for an element with a specific class/id, locator remains more convenient.

yarn.lock
GabrielBruno24: I downgraded to 1.50.0 just in case.

.env.example
GabrielBruno24: I removed them and added default values

packages/transition-frontend/ui-tests/testHelpers.ts
GabrielBruno24: I kept the test counter, when I took it off it did an error on the test where I logged in multiple times.

samuel-duhaime (approved):
Dernier commentaire. Peux-tu mettre un username et password avec un nom plus précis. C'est trop général username et password. 

> const username = process.env.PLAYWRIGHT_TEST_USER || 'testUser';
> const password = process.env.PLAYWRIGHT_TEST_PASSWORD || 'testPassword';

yarn.lock
```diff
@@ -6607,11 +6614,16 @@ fs.realpath@^1.0.0:
   resolved "https://registry.yarnpkg.com/fs.realpath/-/fs.realpath-1.0.0.tgz#1504ad2523158caa40db4a2787cb01411994ea4f"
   integrity sha1-FQStJSMVjKpA20onh8sBQRmU6k8=
 
-fsevents@^2.3.2:
+fsevents@2.3.2:
```
greenscientist, line 6617: weird 2 have 2 fsevents, I imagine that playwright need a specific version of it

greenscientist (approved):
2 smalls comments, but only need to be done if you really want to.

PR #1318, opened 2025-04-04
Pin version of rust toolchain to use for json2capnp test

greenscientist:
The latest version of the toolchain does not work with some of the old dependencies that we have in json2capnp. Until we fix/change json2capnp, let's pin the test to the last working of the rust toolchain

### DIFF
--- a/.github/workflows/transition.yml
+++ b/.github/workflows/transition.yml
@@ -77,7 +77,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions-rs/toolchain@v1
       with:
-        toolchain: stable
+        toolchain: 1.85.0 # Last working version with latest version of Rouille. Was stable
     - run: cargo build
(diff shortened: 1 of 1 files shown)


GabrielBruno24 (approved):

samuel-duhaime (approved):

.github/workflows/transition.yml
```diff
@@ -77,7 +77,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: actions-rs/toolchain@v1
       with:
-        toolchain: stable
+        toolchain: 1.85.0 # Last working version with latest version of Rouille. Was stable
```
GabrielBruno24, line 80: Comment sounds a bit weird, did you mean to have "Was stable" at the end?

greenscientist: Yes, it's meant to remind that the value was "stable" before

PR #1327, opened 2025-04-11
Upgrade socket.io to 4.8.1

tahini:
Upgrading the socket.io server and client package requires minimal changes

Tthe file upload to server was to be completely refactored, both on the server and client side as per the documentation in https://socket.io/how-to/upload-a-file.

On the server, we rewrote the logic to have a single 'uploadFile' socket route with the file as binary data and a file data 
object with import type and file information.

On the client side a hook replaces the HOC to provide the upload functionality. The GtfsImportForm was refactored as function component for easier upgrade.

Also add routes to send the file in chunks


kaligrafy (approved):

greenscientist:
Since @kaligrafy have already click the accept button, I want to comment without looking at the code, that we need to restrict the path where the file can be uploaded in the backend side of it. Might be already like that, but wanted to leave it here before I have time to review the code.

tahini:
J'allais pas merger, le commentaire de copilot est pertinent et ressemble à ce qu'on a eu dans la branche deck.gl en mettant la map dans un function component, je veux retravailler. Mais sinon, du côté serveur, c'est envoyé dans le userData dans un répertoire approprié (imports par défaut) On ne peut overwrite

tahini:
Ceci dit, à propos du commentaire de copilot, c'est quand même lui qui a fait une partie de la conversion à la base... ;-)

tahini:
Bon, tout est beau, c'est prêt pour review officiel

greenscientist (changes requested):
Yeah, I'm quite sure an external user could write any file on the system with this if a filename is named ../../../something...

tahini:
Added filename sanitation and a unit tests to prove it

packages/chaire-lib-backend/src/api/uploads.socketRoutes.ts
```diff
@@ -22,53 +39,137 @@ type ObjectImporter =
           type: 'importOnly';
       };
 
+// Remove anything before a slash or backslash in the filename to prevent path traversal attacks
+const sanitizeFileName = (filename: string) => filename.replace(/^.*[\\/]/, '');
```
greenscientist, line 43: @tahini  Just to confirm, at this point, we would not need to handle %2F ? (Like in a param/name with %2e%2e%2f

greenscientist: Right, we are not in an URL at this point, so this is ok

greenscientist (approved):
I think it's ok.

PR #1352, opened 2025-04-28
user: Add possibility to set users as admin

### LINKED ISSUE #654: users: Allow to set the administrator flag of users
Only an administrator should be able to set this flag. It can be set in a similar way than the other roles checkboxes

tahini:
fixes #654

Add a route in the backend such that the is_admin flag can only be set by an administrator. Extract the user update code so that the other user update route can use the same code. It sanitizes this other route to make sure the is_admin flag cannot be set by a non-admin user.

Add a component to edit the admin status of a user in the users table. A user cannot set his own administrator flag, to not lock himself out of the application.


kaligrafy (approved):

greenscientist (commented):
I don't see where we return a 401 if the a user that is not an admin try to set another user properties. Is that already in another file?

Issue #338, opened 2022-01-06
transition: Deprecate GenericObject#getAttributes()


tahini:
getAttributes() and the attributes accessor are equivalent, and are used interchangeably everywhere. To avoid confusion, we should only use one of them and the attributes accessor looks prettier in code.

PR #1354, opened 2025-04-29
Deprecate getAttributes(), replacing it by an attributes accessor.

GabrielBruno24:
Fix: #338


greenscientist (changes requested):
Can the commit message explain why we are doing this and why we are using accessor instead of getAttributes ?
(It might be in the issues, but we should have it in the commit message to understand the change in the future

kaligrafy (approved):

tahini (commented):
I'm fine with this, please rebase and update the commit message as requested

greenscientist (approved):
lol, ok. I was expecting a more complex rationale than "looks prettier", but it is what it is!

PR #1361, opened 2025-05-01
dep: remove un-used dependencies and put them in the right place

### LINKED ISSUE #431: Check all dependencies are in the right place
For each workspace, make sure what is under dependencies is the stuff needed at runtime. Anything required to build and compile should be in the devDependencies. Some packages seem to be misplaced, like types, or babel/webpack stuff

tahini:
fixes #431

Using npm's `depcheck` module, validate that all packages in the `package.json` are used and put in the right `dependencies` or `devDependencies` section.

Notable changes:

* `child_process` package is now built-in in node, so it is not required anymore

* `gtfs-types` is now in the `devDependencies` as it is only used in `import type` statements, so a line to disable eslint has been added.


tahini:
@samuel-duhaime ouais, au prochain update de transition dans evolution, tout ce travail sera à faire là aussi! (yé!...)

kaligrafy (approved):

Issue #1324, opened 2025-04-09
tests: Run sequential tests in the CI


tahini:
We do not run them as they require a test database setup. The json2capnp sequential tests also require a running json2capnp instance with the right parameters. But we recently added UI tests with playwright, who also require a running json2capnp and database setup. So maybe we could add the sequential tests to this CI test run, or with a separate CI, using the same setup.

Note that since those tests are not systematically run by the developers either, in the past, we often found out that they were broken when someone knowingly changed something required to run them and just realized that they had been broken for a while... Would it helpful to know right away.

PR #1373, opened 2025-05-12
tests: destroy connection after db tests

tahini:
By adding `await knex.destroy()` in the `afterAll` block of each database tests, we make sure that the connection is properly destroyed, synchronously, and the tests can exit without open handles.

Some tests already called the `destroy` db query from some tables, but this function was effectively destroying the connection multiple times (as all queries share the same connection) and it was not done synchronously as no callback was passed to the function, so the test was still finishing with open handles.


tahini:
First step before being able to do #1324

kaligrafy (approved):

PR #1376, opened 2025-05-14
file upload: create directory before uploading file

### LINKED ISSUE #1375: userData directory does not exist when uploading files
Maybe a side-effect of the recent passage to socket.io 4 for file upload, but in the infra, the following error occurs when uploading a GTFS for import: `directory does not exists for file path: /app/projects/demo_transition/userData/x/gtfs/import.zip`. The userData for the user does not exist.

It went unnoticed locally as those directories already existed, but deleting it reproduces the error.

tahini:
fixes #1375

The `socket.io-file` package used to create the directory where to upload the file if it did not exist. When this package was replaced by our own code, this behavior was not replicated. We now create the directory if it does not exist before saving the file.

### DIFF
--- a/packages/chaire-lib-backend/src/api/uploads.socketRoutes.ts
+++ b/packages/chaire-lib-backend/src/api/uploads.socketRoutes.ts
@@ -89,23 +89,30 @@ export default function (
         }
     };
 
-    // TODO Handle file chunking
+    // Get the absolute file path to save the file to, and makes sure the directory exists
+    const prepareUploadFileAndDir = (options: FileUploadOptions): string => {
+        // Get the folder where to save the file
+        const saveToFolder = allOptions.uploadDirs[options.uploadType] || allOptions.uploadDirs['imports'];
+        if (!allOptions.uploadDirs[options.uploadType]) {
+            console.error(`No folder set for upload type ${options.uploadType}. Saving to 'imports' directory.`);
+        }
+
+        const importFileName =
+            typeof options.data.filename === 'string' ? sanitizeFileName(options.data.filename) : 'upload.txt';
+
+        const saveToFile = `${saveToFolder}/${importFileName}`;
+
+        // Create the directory if it does not exist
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):

packages/chaire-lib-backend/src/api/uploads.socketRoutes.ts
```diff
@@ -89,23 +89,29 @@ export default function (
...
+            console.error(`No folder set for upload type ${options.uploadType}. Saving to 'imports' directory.`);
+        }
+
+        const importFileName =
+            typeof options.data.filename === 'string' ? sanitizeFileName(options.data.filename) : 'upload.txt';
+
+        const saveToFile = `${saveToFolder}/${importFileName}`;
+
+        // Create the directory if it does not exist
+        fileManager.directoryManager.createDirectoryIfNotExistsAbsolute(saveToFolder);
```
greenscientist, line 105: This is in a "getSavedFilePath" function.  A "get" function should not really have a side effect like this. It should be moved elsewhere.

greenscientist (changes requested):
The getSavedFilePath function does too many things.

packages/chaire-lib-backend/src/api/uploads.socketRoutes.ts
```diff
@@ -89,23 +89,30 @@ export default function (
         }
     };
 
-    // TODO Handle file chunking
+    // Get the absolute file path to save the file to, and amkes sure the directory exists
```
greenscientist, line 92: typo..  makes

greenscientist (approved):
Good enough. I would prefer a separate fonction for the name and one for the directory creation,  but ok. 

One last note: Should we have a test were we don't have to create the directory ?

tahini:
> One last note: Should we have a test were we don't have to create the directory ?

In the tests, we just check that the function that would create the directory is called, it does not actually create a new directory, so we don't know if the directory exists or not (we stub the fs), so no, no need for more tests.

PR #1381, opened 2025-05-14
Use Node 24.x for Docker builds and add it to test matrix

greenscientist:


greenscientist:
I've kept the package.json at >= 22.x, but we could decide to bump it to 24.

tahini (approved):
It's ok to keep 22.x for now. I suppose you tested your node 24 dockerfile and it works fine?

greenscientist:
Yup, tested locally, tested routing and accessbility and was fine.

kaligrafy (approved):

PR #1385, opened 2025-05-16
Duplicate schedules in the backend

### LINKED ISSUE #943: Duplicate schedules in the backend

tahini:
* Add a db query that copies in one query per table all schedules for given line, service, path mappings. This will in time allow to copy all the schedules for an agency in 3 queries
* Add socket routes to duplicate schedules in the backend
* Update the LineDuplicator to use the backend db duplicator
* Unit test everything

fixes #943

### DIFF
--- a/packages/transition-backend/src/api/__tests__/transitObjects.socketRoutes.test.ts
+++ b/packages/transition-backend/src/api/__tests__/transitObjects.socketRoutes.test.ts
@@ -40,6 +40,10 @@ jest.mock('../../services/transitObjects/transitServices/ServiceDuplicator', () 
     duplicateServices: jest.fn()
 }));
 
+jest.mock('../../services/transitObjects/transitSchedules/ScheduleUtils', () => ({
+    duplicateSchedules: jest.fn()
+}));
+
 import { v4 as uuidV4 } from 'uuid';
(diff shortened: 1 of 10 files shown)


tahini:
This branche was lying around my repo, I finally finished it!

kaligrafy (approved):

greenscientist:
The title of the 3rd commit worries me: "schedules: Use the backend schedule duplicator where possible". 

Is there still cases where we duplicate schedule from the frontend ? Can we remove those?

tahini:
This branche was lying around my repo, I finally finished it!

> The title of the 3rd commit worries me: "schedules: Use the backend schedule duplicator where possible".
> 
> Is there still cases where we duplicate schedule from the frontend ? Can we remove those?

You are right, I copied this commit message from the service duplicator, where it was not yet possible to do, or _someone_ was too lazy to make it possible or force the author to make it possible in review, not sure who to throw the rock to ;-) Anyhow, in this specific case, it was possible to move it all to backend, so I can just update the commit message!

greenscientist (approved):
Looks ok. 
The duplication with the mapping is a bit complicated, but there's a good amount of comment.

PR #1398, opened 2025-06-20
jobs: give feedback to users when there are errors or warnings in file

### LINKED ISSUE #825: calculations: No error feedback to the user (disk quota reached, errors in file, etc)
A user reported that new calculation tasks were set as pending but never appeared in the calculation jobs table. Looking at the server, it appears this user had reached the disk usage quota. There should be an explicit message for this case, instead of showing a pending task that does not exist.

tahini:
fixes #825 

See individual commits for details.


packages/transition-backend/src/models/db/jobs.db.queries.ts
```diff
@@ -16,13 +16,29 @@ const tableName = 'tr_jobs';
 const attributesParser = <U extends JobDataType>({
     resources,
     updated_at,
+    status_messages,
     ...rest
-}: JobAttributes<U>): JobAttributes<U> => ({
+}: Omit<JobAttributes<U>, 'statusMessages'> & {
+    status_messages: JobAttributes<U>['statusMessages'];
```
GabrielBruno24, line 22: Can you add a comment explaining why we need this complex type?

packages/transition-common/src/services/jobs/Job.ts
```diff
@@ -21,6 +23,7 @@ export interface JobAttributes<TData extends JobDataType> {
     user_id: number;
     name: TData[JobNameKey];
     status: JobStatus;
+    statusMessages?: { errors?: ErrorMessage[]; warnings?: ErrorMessage[]; infos?: ErrorMessage[] };
```
GabrielBruno24, line 26: Do all of these need the ErrorMessage type? Is there not a more appropriate one?

packages/transition-frontend/src/components/parts/executableJob/ExecutableJobComponent.tsx
```diff
@@ -150,9 +151,16 @@ const ExecutableJobComponent: React.FunctionComponent<ExecutableJobComponentProp
                 accessor: 'status',
                 width: 70,
                 Cell: (cellProps) => (
-                    <div className={`status_${cellProps.value}`}>
-                        {componentProps.t(`transit:jobs:Status_${cellProps.value}`)}
-                    </div>
+                    <>
+                        <div className={`status_${cellProps.value}`}>
+                            {componentProps.t(`transit:jobs:Status_${cellProps.value}`)}
```
GabrielBruno24, line 156: While not the purpose of this pr, since this is a function component maybe you could take the time to update the syntax for the translation?

GabrielBruno24 (commented):
A few comments, but looks good

packages/transition-common/src/services/jobs/Job.ts
tahini: > Do all of these need the ErrorMessage type? Is there not a more appropriate one?

It is the type that we use for error messages that can be translated (either translatable string or a text with additional parameters). These messages can be fed to the `t` function directly.

Do you have a suggestion for a more appropriate type?

GabrielBruno24: I just think it's strange for all three attributes of statusMessage to have an Error type, including infos. But it's also not a big deal if it stays like that.

tahini: J'ai ajouté cet issue: https://github.com/chairemobilite/transition/issues/1400

greenscientist (approved):
letsego

PR #1408, opened 2025-06-23
In the accessibility comparison tab, add a plus to the numbers in the difference collumn if they are positive.

GabrielBruno24:
This makes it clearer to the user what the effect of the different parameters are.

Screenshot from 2025-06-23 15-58-26


greenscientist (commented):
Isn't there an equivalent of an sprintf function of something like that that would handle formating and add this already ? (instead of creating a new function ?

GabrielBruno24:
> Isn't there an equivalent of an sprintf function of something like that that would handle formating and add this already ? (instead of creating a new function ?

I used toLocaleString to format.

PR #1413, opened 2025-07-07
captcha: allow to keep a token after validation

tahini:
By default, once a cap.js token is validated, it is deleted and cannot be reused by any other query. This is not desirable, for example, in case of login, where legitimate errors can occur and the user may need to re-submit a modified form. He should not have to re-do the captcha every time.

This adds an option to the captcha middleware, a boolean `keepToken` value, which can be set per route, such that a query that should succeed everytime (like help request) can delete the token, while others, like login, can keep the token between queries.

Note that cap.js does a garbage collection of the expired token in any case, this option only applies to the current query.

### DIFF
--- a/packages/chaire-lib-backend/src/api/captcha.routes.ts
+++ b/packages/chaire-lib-backend/src/api/captcha.routes.ts
@@ -10,7 +10,7 @@ import config from '../config/server.config';
 import { fileManager } from '../utils/filesystem/fileManager';
 
 // Define the type for the validation function
-export type CaptchaValidationFunction = (token: string) => Promise<boolean>;
+export type CaptchaValidationFunction = (token: string, opts: { keepToken: boolean }) => Promise<boolean>;
 
 // This variable will hold the validation function once the library is loaded
 let validateCaptcha: CaptchaValidationFunction | undefined = undefined;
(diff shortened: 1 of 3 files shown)


packages/chaire-lib-backend/src/api/captcha.routes.ts
```diff
@@ -58,9 +60,9 @@ export default (router: express.Router) => {
                 });
 
                 // Define the validation function
-                validateCaptcha = async (token: string): Promise<boolean> => {
+                validateCaptcha = async (token: string, opts): Promise<boolean> => {
                     try {
-                        const validationResult = await cap.validateToken(token);
+                        const validationResult = await cap.validateToken(token, { keepToken: opts.keepToken });
```
tahini, line 65: it's used here and yes, it's included in cap.js

packages/chaire-lib-backend/src/api/captcha.routes.ts
```diff
@@ -22,14 +22,16 @@ export interface CaptchaMiddlewareOptions {
...
 
-const defaultMiddlewareOptions: CaptchaMiddlewareOptions = {
+const defaultMiddlewareOptions: Required<CaptchaMiddlewareOptions> = {
     tokenField: 'captchaToken',
     queryParam: 'captchaToken',
     headerName: 'X-Captcha-Token',
     errorStatusCode: 403,
-    errorMessage: 'Captcha validation failed'
+    errorMessage: 'Captcha validation failed',
+    keepToken: true // Default to re-using the token
```
tahini, line 34: We can default to `false` here if it makes more sense

tahini (commented):
Also, I'll have to make sure it is possible to reset the captcha widget for cap.js. Currently, it is not reset by default.

packages/chaire-lib-backend/src/api/captcha.routes.ts
```diff
@@ -22,14 +22,16 @@ export interface CaptchaMiddlewareOptions {
...
 
-const defaultMiddlewareOptions: CaptchaMiddlewareOptions = {
+const defaultMiddlewareOptions: Required<CaptchaMiddlewareOptions> = {
     tokenField: 'captchaToken',
     queryParam: 'captchaToken',
     headerName: 'X-Captcha-Token',
     errorStatusCode: 403,
-    errorMessage: 'Captcha validation failed'
+    errorMessage: 'Captcha validation failed',
+    keepToken: false // Default to re-using the token
```
greenscientist, line 34: Need to update the comment after switching the default. 
Also while you are at it, could you add a comment to that explain the impact of setting it to true. 
Took me a bit of time to fully grasp it. 
If I understand properly, basically the idea, when set to true, is that the captcha validation will be valid for the lifetime of the session (or until it times out)

greenscientist (approved):
Approving, but comment need to be added/updated.

PR #1415, opened 2025-07-08
db: Add `scenarioId` parameter to collection queries

tahini:
Agencies, lines, services and paths can now be queried with an optional scenario ID. This will fetch the objects actually used by the scenario, including the exclusion/includes of lines/agencies.

Also fixes the display of the paths used by a scenario, in the scenario edit form


tahini:
I've put a single scenarioId in the queries, but should it be an array instead? is there a use case to support such a query? Or we keep the single scenario filter?

kaligrafy (approved):

greenscientist (commented):
I think we might need more comment in the DB query, specially about the new filters. It's getting a bit hard to follow.

tahini (commented):
I added separate functions for collectionForScenario in all db queries file. Base queries are extracted to a common function that both `collection` and `collectionForScenario` can call.

greenscientist (commented):
ok, sorry to come back to it, but having a single entry point/function which could take optioanl parameters could be interesting in some case. (Thinking about the capnp cache generation, where would could something want to have a scenario filter). 

But as discussed, we could keep the function you divided and hide them behind a frontend one which would select the right one based on filter.

tahini:
> ok, sorry to come back to it, but having a single entry point/function which could take optioanl parameters could be interesting in some case. (Thinking about the capnp cache generation, where would could something want to have a scenario filter).
> 
> But as discussed, we could keep the function you divided and hide them behind a frontend one which would select the right one based on filter.

While we're discussing this, I'm wondering... the `collection` functions were originally designed to return the collection with enough interesting information to display in the panels. That why many of them have joins with other tables, with raw array aggregation in their select, for example to count the number of lines in an agency or using a service, or the number of paths and schedules a line has, etc.

In the use cases where we add a filter (that I think of), we usually just care about the table's own data (with potential joins to apply the filter, like the scenario case). Should we rather distinguish collection functions between collections for a listing to the user, with additional [aggregated] data, and those for the purpose of calculations, with only table's field.

Note that the user may also want to filter by something, so this question or purpose is probably orthogonal to the filter one. But we can think about it for the future.

greenscientist:
This collections concept is probably the less efficient way we could access some data, you are right. We could have better taillored calls to get only the data we need.

tahini:
My 2 cents: KISS, we need simple queries in the backend... As for filters, these may require joins sometimes (like the scenario), but these joins can be done on demand only. I have no preference on whether they are done in separate functions or in the single collection function, except that with a single filter in a function, filters can add up while specific functions can get out of hand.

PR #1416, opened 2025-07-08
db: add queries that will be useful for the transit validation

tahini:

* One query uses the schedules to get the trips during a time range for a set of lines and services
* One query gets the transferrable nodes between 2 paths.


packages/transition-backend/src/models/db/transitSchedules.db.queries.ts
```diff
@@ -800,6 +800,39 @@ const duplicateSchedule = async ({
     }
 };
 
+const getTripsInTimeRange = async ({
+    rangeStart,
```
greenscientist, line 804: Please specifiy format as comment (epoch, time from midnight? Second, minutes?)

greenscientist: Probably also add in the function comment header describing the feature that the time in the range is inclusive

greenscientist (changes requested):
Need more comments for the node transferable one. 

With the comment provider in the commit message, I don't see what is the algo expected. This should go in a comment describing the function. 
Do we do a path by path comparison ? (first one of one array and the first one of the other array?)

PR #1424, opened 2025-07-14
Read `yaml` files on the server for translations

tahini:
And allow user email notification to be overwritten with `customServer` namespace


packages/chaire-lib-backend/src/config/i18next.ts
```diff
@@ -23,6 +23,8 @@ const getTranslationPath = (lng: string, namespace: string) => {
             return `${translationPath}/{{lng}}/{{ns}}.yml`;
         } else if (fileManager.fileExistsAbsolute(`${translationPath}/${lng}/${namespace}.json`)) {
             return `${translationPath}/{{lng}}/{{ns}}.json`;
+        } else if (fileManager.fileExistsAbsolute(`${translationPath}/${lng}/${namespace}.yaml`)) {
```
greenscientist, line 26: Nitpicking, but could you place that right next to the .yml one

greenscientist (commented):
Do we want to support both .yml and .yaml, or just one and add a linter somewhere to block us from doing the wrong one ?

PR #1435, opened 2025-07-17
icons: copy the icon files in the `lib` directory

tahini:
This puts the icons in a directory that is expected to be public and distributed along with the `chaire-lib-frontend` package. It ignores the `sources` directories.

webpack can then copy those files in the `dist` directory to publish them


greenscientist:
WHy do assets needs to be in the lib directory ?

samuel-duhaime (approved):

PR #1439, opened 2025-07-21
gtfs: get nodes in radius concurrently

tahini:
With only one concurrent queue, this operation, when importing ~10K nodes for an agency could take many minutes. With a concurrency of 10, the time is decreased to less than a minute.

This does not solve any issue, but it mitigates the slow node import during GTFS import.


greenscientist (changes requested):
This will be a good headache. I cannot find an easy to make this concurrency safe.

PR #1446, opened 2025-07-24
add ornament asset (circle flowers)

kaligrafy:


greenscientist (changes requested):
Where's the credit/licence info?

kaligrafy:
> Also do we need this in chaire-lib or should it be application specific? (ie in Evolution directly)

I plan to include it in more apps, so it should go into chaire-lib

kaligrafy:
> Where's the credit/licence info?

I will add it (MIT License, made from scratch)

samuel-duhaime (approved):

PR #1453, opened 2025-08-01
Import census canada data

GabrielBruno24:
In progress. Right now, imports the dissemination blocks and their boundaries, with the population and proximity indices being next.

Fix: #1396

Screenshot from 2025-08-01 17-36-42


packages/transition-backend/src/models/db/__tests__/census.db.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+        const collection = await dbQueries.collection();
+        expect(collection.length).toEqual(2);
+
+        expect(collection[0].zone_id).toEqual(id1);
+        expect(collection[0].population).toEqual(123456);
+
+        expect(collection[1].zone_id).toEqual(id2);
+        expect(collection[1].population).toEqual(0);
+
```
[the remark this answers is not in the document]
GabrielBruno24: The census table uses an incrementing id, so the order should be the same everytime.

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
```diff
@@ -164,13 +164,88 @@ const getZonesContaining = async (
     }
 };
 
+const addZonesAndConvertedGeography = async (
```
tahini, line 167: This function deserves more comments. Why `addZonesAndConvertedGeography`? How is that different from the `create` or `createMultiple` functions? What is `srIdArray` and what is the expected format of the `geographyArray`?

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
```diff
@@ -164,13 +164,88 @@ const getZonesContaining = async (
...
+                if (!uuidValidate(id.id)) {
+                    throw new TrError(
+                        `Cannot read object from table ${tableName} because the required parameter id is missing, blank or not a valid uuid`,
+                        'DBQZONE0005',
+                        'MissingZoneIdWhenAddingConvertedGeography'
+                    );
+                }
+                return [id.id, srIdArray[index], geographyArray[index]];
+            });
+            const placeholders = data.map(() => '(?, ?, ?)').join(',');
```
tahini, line 192: Please add some comments in this block

packages/transition-backend/src/models/db/census.db.queries.ts
```diff
@@ -0,0 +1,69 @@
...
+    } catch (error) {
+        throw new TrError(
+            `cannot fetch census collection because of a database error (knex error: ${error})`,
+            'CSDB0001',
+            'CensusCollectionCouldNotBeFetchedBecauseDatabaseError'
+        );
+    }
+};
+
+const addPopulationBatch = async (internalIdArray: string[], populationArray: number[]): Promise<void> => {
```
tahini, line 34: It could also update the population if it already exists for a zone. If so, you can look at the `onConflict` function, but a comment somewhere in the code says it does not work with batch inserts, so maybe it's not possible?

Also, instead of having 2 arrays that you hope match, you could have a single array of objects with `{ zoneId: uuid, population: number }`. Then you know the id of the zone always follows its population. The same comment applies to the function in the `zones.db.queries.ts` file where you had 3 array parameters that you could transform into a single array of objects.

packages/transition-backend/src/models/db/census.db.queries.ts
```diff
@@ -0,0 +1,69 @@
...
+        await knex.transaction((trx) => {
+            const data = internalIdArray.map((id, index) => [id, Number(populationArray[index] ?? 0)]); //In case of a nullish population value, insert a 0 instead.
+            const placeholders = data.map(() => '(?, ?)').join(',');
+            const bindings = data.flat();
+
+            const query = `
+                INSERT INTO ${tableName} (zone_id, population)
+                SELECT z.id, v.population::integer
+                FROM (VALUES ${placeholders}) AS v(internal_id, population)
+                JOIN ${parentTable} z ON z.internal_id = v.internal_id
```
tahini, line 51: What if you try inserting data for a zone that does not exist? There should be a test for this case. that should cancel the whole insert

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
```diff
@@ -0,0 +1,384 @@
...
+    prox_idx_lib: number | null;
+    /** Normalised value of a dissemination block's proximity to parks. */
+    prox_idx_parks: number | null;
+    /** Normalised value of a dissemination block's proximity to transit trips. */
+    prox_idx_transit: number | null;
+}
+
+/**
+ * Task to import data about canada's dissemination blocks into the database.
+ * This is intended to be used with data provided by statcan.
```
tahini, line 63: Can you provide links to the data and how to get the files or the specs of the file to import?

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
```diff
@@ -0,0 +1,384 @@
...
+ *
+ * task options are:
+ * * --boundaries-file: Absolute path to the GML file defining the boundaries
```
tahini, line 66: Boundaries file could be at any level, right? not just dissemination blocks? Like if we want Census divisions or subdivisions, it's ok.

Also, you can add a fixme to support also other file types for boundaries, like geojson directly, if available for some levels of precision.

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
```diff
@@ -0,0 +1,384 @@
...
+                // The id of the coordinate system used. Necessary to later convert into the coordinates used by Transition.
+                let srId: string = '';
+
+                // The GML file uses two types of geometry for the blocks: simple surfaces and multi surfaces, with slightly different structures that need their own parsing logic
+                // Most of them are simple surfaces, with multi surfaces having multiple non connected polygons, such as a block that includes islands
+                const surfaceProperty = fme.children.find((element) => element.name === 'gml:surfaceProperty');
+                if (surfaceProperty !== undefined) {
+                    const surface = surfaceProperty.children[0];
+                    const srsName = surface.attributes.srsName;
+                    srId = srsName.split('EPSG:')[1];
```
tahini, line 147: What's `srId`? Usually, in Octavi for instance, it refers to "Secteur de recensement". Is that it? We have the secteur this block belongs to?

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
```diff
@@ -0,0 +1,384 @@
...
+        let i = 0;
+
+        const batchSize = 1000;
+        let batch: ZoneAttributes[] = [];
+        let srIdArray: string[] = [];
+        let geographyArray: string[] = [];
+
+        console.log('Parsing through GML FILE:');
+
+        parser.on('tag:gml:featureMember', (feature: XmlNode) => {
```
tahini, line 128: This blokc is a GML file parser. Have you considered npm packages that do this? like `gml2geojson`? if so, why not use them? You can add a comment why we need to write the import. And a TODO to extract it to a library, so we can eventually support many file types.

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
```diff
@@ -0,0 +1,384 @@
...
+                    .value;
+
+                const multiCoorList: string[] = [];
+
+                // The id of the coordinate system used. Necessary to later convert into the coordinates used by Transition.
+                let srId: string = '';
+
+                // The GML file uses two types of geometry for the blocks: simple surfaces and multi surfaces, with slightly different structures that need their own parsing logic
+                // Most of them are simple surfaces, with multi surfaces having multiple non connected polygons, such as a block that includes islands
+                const surfaceProperty = fme.children.find((element) => element.name === 'gml:surfaceProperty');
```
tahini, line 143: All those fields: `gml:surfaceProperty`, `gml:LinearRing`, are they part of the GML standard or Stat Can's?

packages/transition-backend/src/models/db/migrations/20250722152900_createCensusTbl.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function up(knex: Knex): Promise<unknown> {
+    if (await knex.schema.hasTable(tableName)) {
+        return;
+    }
+    await knex.schema.createTable(tableName, (table: Knex.TableBuilder) => {
+        table.increments();
+        table.uuid('zone_id').index();
+        table.foreign('zone_id').references('tr_zones.id').onDelete('CASCADE');
+        table.integer('population');
```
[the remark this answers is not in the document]
GabrielBruno24: It is individual people. Dissemination blocks are the smallest zones in the census, so in rural areas they can have a very low population (or even zero).

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
GabrielBruno24: Spatial reference ID.

GabrielBruno24: It's all the same for the file I want to import but since it's technically defined in every individual block for the GML I'm reading it each time.

packages/transition-backend/src/tasks/disseminationBlocks/importDisseminationBlocks.ts
GabrielBruno24: It was a combination of several things. I wanted a package that:
1- Supports streaming for reading large files
2- Works on a generalized xml and doesn't require an input with an overly restrictive structure.
I'll have to test this package later but from my own research xml-reader seemed the best.

greenscientist (changes requested):
A few quick things: 
* It would really be great if we could have a doc file describing our scripts like that. Could you start one, we only the task that you are adding?
* I think it would be nice if we could split the "UI" part and the "processing" part of importDessimenationBLocks.ts file. It's all in one class for now, but I think it would be better to have a UI class that use a processor class. If we have to trigger it from the webUI some day we can just use the processor class.

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
```diff
@@ -164,13 +164,88 @@ const getZonesContaining = async (
     }
 };
 
+const addZonesAndConvertedGeography = async (
+    zonesBatch: ZoneAttributes[],
+    srIdArray: string[],
```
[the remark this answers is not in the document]
kaligrafy: Yes the zones table should take any geography data and any id. There should be an internalId for this, which is a string by default to accomodate any id format in the source database.

greenscientist:
Also see #1501 for the upgrade on inquirer and the new usage.

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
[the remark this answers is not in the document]
GabrielBruno24: Sr id is the id of the coordinate system used, so we can then convert them to WSG84. The id of the zone _is_ internalId

packages/transition-backend/src/models/db/census.db.queries.ts
GabrielBruno24: Right now, running the task while there is already data will ask to delete the data source first, so there shouldn't be a situation where you insert data for zones that already exist. I can add a todo in case we call this from elsewhere in the future.

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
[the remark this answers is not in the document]
greenscientist: Maybe a more explicit/different name then?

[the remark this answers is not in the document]
greenscientist: Also, should we instead convert the coordinate directly so we don't have different values?  Or is that something that is done automagically by postgis? This need way more comments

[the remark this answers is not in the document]
GabrielBruno24: The conversion is done bu the postgis function ST_Transform. I can add a few comments to make it clearer

README.md
```diff
@@ -165,6 +165,10 @@ For example, to run yarn test in the chaire-lib backend:
 You can also run the app this way with:
 `docker run -a STDOUT -it -v "${PWD}:/home/project" -w=/home/project/ testtransition yarn start`
 
+## Running tasks in the command prompt
+
+Transition provides a number of tasks that can be run from the command prompt without having to build the application and open the UI. For more information, see the [documentation on tasks](docs/tasks.md).
+
```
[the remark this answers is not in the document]
GabrielBruno24: Am I not already doing that here? I don't see where the issue is.

packages/transition-backend/src/services/importers/ProximityMeasuresAndPopulationImporter.ts
```diff
@@ -0,0 +1,112 @@
...
+    console.log(`There are ${lineCounter} dissemination blocks.`);
+    console.log('Adding proximity indices and population data to database:');
+    let dbCounter = 0;
+    for (const batch of csvData) {
+        console.log(`CSV data added to DB: ${dbCounter * batchSize}/${lineCounter}`);
+
+        const jsonBatch = batch.id.map((id, index) => {
+            return { internalId: id, json: batch.indices[index] };
+        });
+        await zonesQueries.addJsonDataBatch(jsonBatch);
```
tahini: Il n'y a pas de try/catch dans ces fonctions. Est-ce que les erreurs sont correctement attrapées?

yarn.lock
```diff
@@ -12646,11 +12626,26 @@ wsrun@^5.2.4:
...
+  integrity sha512-G0i98epIwiUEiKmMcavmVdhtymW+pCAohMRgybyIME9ygfVu8QheIi+YoQh3ngiThsT0SQzJT4R0sKDEv8Ou0w==
+  dependencies:
+    eventemitter3 "^2.0.0"
+
 xml-name-validator@^4.0.0:
   version "4.0.0"
   resolved "https://registry.yarnpkg.com/xml-name-validator/-/xml-name-validator-4.0.0.tgz#79a006e2e63149a8600f15430f0a4725d1524835"
   integrity sha512-ICP2e+jsHvAj2E2lIHxa5tjXRlKDJo4IdvPvCXbXQGdzSfmSpNVyIKMvoZHjDY9DP0zV17iI85o90vRFXNccRw==
 
+xml-reader@^2.4.3:
```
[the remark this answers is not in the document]
tahini: Potentiellement une information intéressante à ajouter au commit message...

packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
```diff
@@ -164,13 +164,99 @@ const getZonesContaining = async (
...
+    } catch (error) {
+        throw new TrError(
+            `Problem adding new object to table ${tableName} (knex error: ${error})`,
+            'DBQZONE0005',
+            'ProblemAddingObject'
+        );
+    }
+};
+
+const addJsonDataBatch = async (inputArray: { internalId: string; json: { [key: string]: any } }[]): Promise<void> => {
```
greenscientist: Can you add a description for this function ?

docs/tasks.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+Where [filepath] is the relative file path of the task wrapper with the task.ts extension (except with `/src/` replaced by `/lib/` and `.ts` replaced with `.js`) and [parameters] is the parameters of the command, varying from task to task.
+
+You can also run the command `yarn list-tasks` to receive a list of tasks to run, as well as the commands necessary to run them.
+
+Below is a list of the tasks currently implemented, the path to run them, what they do, their parameters, and an example.
+
+
+## Import dissemination blocks
```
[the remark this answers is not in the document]
GabrielBruno24: No, but this task is meant to be run specifically with the two files in the description. If we want something that reuses those functions but for other levels, we can write a new task in the future.

docs/tasks.md
```diff
@@ -0,0 +1,42 @@
+Transition provides a number of tasks that can be run from the command prompt without having to build the application and open the UI.
+
+To run them, you must first compile the transition source code, and run a command at the root of the Transition repo of the form:
+
+```shell
```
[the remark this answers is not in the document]
GabrielBruno24: I have already done this in a previous push.

docs/tasks.md
```diff
@@ -0,0 +1,42 @@
...
+Below is a list of the tasks currently implemented, the path to run them, what they do, their parameters, and an example.
+
+
+## Import dissemination blocks
+
+### File path
+
+```text
+packages/transition-backend/lib/scripts/disseminationBlocks/importDisseminationBlocks.task.js
+```
```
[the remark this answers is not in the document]
GabrielBruno24: I have also done this already.

greenscientist:
Why do we pasrt the GML file directly instead of using PostGIS ST_GeomFromGML ?

PR #1458, opened 2025-08-19
dep: replace `react-loadable` by `lazy` and `suspense`

tahini:
`react-loadable` is a 7 years old package that uses a legacy contextTypes API, removed in React 19.

The same component behavior can be achieved by using React's `lazy` and `Suspense` functionalities instead.

This affects the `UsersPage` component.


samuel-duhaime (changes requested):
It looks good. Can you test that we see the loading component with Google Chrome (or Firefox) in slow mode?

image

PR #1492, opened 2025-09-22
tests: fix `db.config` mock in `ServiceDuplicator` test

tahini:
The mock required the original module to only mock the `transaction` part, but the db.config is already mocked at the package level so should remained mocked, otherwise, we get a `database is not set` error.

## Summary by CodeRabbit

* **Tests**
  * Simplified database transaction mocks in transit services test suites, replacing partial real-module mocks with minimal, focused mocks.
  * Improves test isolation, stability, and consistency across test runs.
  * No impact on production behavior or user-facing functionality.


greenscientist (approved):
Seems fine to me.

PR #1498, opened 2025-10-02
dep: replace `osmtogeojson` with `osm2geojson-lite`

### LINKED ISSUE #1364: osmtogeojson use an old xmldom version
We have this critical dependabot issue. 
https://github.com/chairemobilite/transition/security/dependabot/55

We can try to ask the maintainer of osmtogeojson to update its dependancy or convert our code to use something else. (Maybe convertir it to rust)

tahini:
fixes #1364

This is a drop-in replacement of the original package, with less dependencies and more supported.

While the geojson generated is a bit different, the differences mostly reside in how it handles relations. `osmtogeojson` would create the features for each feature of the relation, creating duplicates in the content, while `osmtogeojson-lite` will create a node where the relation happens and a single feature for the relation itself.

For our use cases, that difference in geojson output does not pose problems. Our import scripts and usages give the same results as previously.

## Summary by CodeRabbit

- Bug Fixes
  - Improved completeness and accuracy when converting OSM data to GeoJSON, reducing missing or incomplete features.
  - Enhanced error logging during data streaming for clearer diagnostics.

- Tests
  - Updated expectations to reflect non-functional JSON property ordering changes.

- Chores
  - Swapped the OSM-to-GeoJSON converter dependency for a lighter, more robust alternative to improve reliability and performance.

### DIFF
--- a/packages/chaire-lib-common/package.json
+++ b/packages/chaire-lib-common/package.json
@@ -31,7 +31,7 @@
     "kdbush": "^3.0.0",
     "lodash": "^4.17.21",
     "node-fetch": "^2.7.0",
-    "osmtogeojson": "^3.0.0-beta.5",
+    "osm2geojson-lite": "^1.1.2",
     "papaparse": "^5.5.2",
     "pbf": "^3.3.0",
     "random": "^5.4.0",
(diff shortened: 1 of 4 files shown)


greenscientist (approved):
Nice reduction in the yarn.lock

tahini:
56 security issues before, let's see how many after

Issue #405, opened 2022-07-15
transition: Ensure data integrity between nodes and paths in the DB


tahini:
Currently, a path's node IDs are in an array field of the path table. This type of data cannot have a foreign key with the nodes table, which means that if a node gets deleted and is used by a path (which is possible through the delete multiple feature and can always be possible from some obscure code path), then the paths has incoherent data, the calculator cannot load the data from cache, so this prevents all calculations from executing correctly.

We should have a table that links the path_id and node_id, with a sequence in path, with foreign keys and on delete cascade. At least the data will remain coherent and it will be easier to validate if a node is used (instead of querying in an array field). Maybe eventually this table can also store the segment data (to or from the node), to make sure the data of the path is also kept coherent.

greenscientist:
@tahini Do you think we could bump the priority of this ? 
The current data structure with the big blob of json makes it harder to extract the information in trRouting directly.

greenscientist:
(Probably to be combined with https://github.com/chairemobilite/transition/issues/247)

tahini:
Indeed, #247 could help with #761 and #802 as well. And it should make the #801 easier... They could all go together in a major path refactoring. And there's probably a few more of those among the issues. We can add all them to a milestone for early next year

Issue #48, opened 2021-03-25
Manage object history at the form level


tahini:
Currently, objects extend a class called ObjectWithHistory that keeps track of the changes in the object, so operations can be undone/redone. But this has a few problems:

~~1. Undo/redo are not atomic, ie a single action by the user may cause multiple variables to be set (see Path for example) and requires more than one click on 'undo' to really undo.~~

~~2. Tracked changes are only those done using the set(name, value) method, which happens to have the disadvantage of not being able to be typed correctly as name can be any string and field values can be set in other manners. Any other change will not cause undo/redo actions.~~

3. Some forms values entered by the user may not result in a field set in the object, so it cannot be tracked as undo-able.

So object/user-action history should be done by a separate component at the form level instead of through the ObjectWithHistory class. This way, many things can be tracked by the history tracker, not just changes to objects.

PR #1507, opened 2025-10-07
transitRouting: Refactor forms using `TransitRouting` objects to function components

tahini:
As part of a refactoring to move away from `GenericObject`s and `ObjectWithHistory` whose history will rather be tracked outside the class itself, using hook, the `ScenarioComparisonPanel` and `TransitRoutingForm` are turned into a function component. The functionalities of the `ChangeEventForm`, from which the class components inherited are temporarily copied into this form and will be replaced by a hook shortly.

Drop unused props from the forms.

First step towards #48, which happens to be a pin in the foot towards solving #297, which is a step in the ultimate goal of reaching #405...

## Summary by CodeRabbit

- Refactor
  - Rewrote Scenario Comparison and Transit Routing panels to modern hook-based components for improved responsiveness and smoother interactions.
  - Simplified translation wiring and internal UI logic while preserving CSV export and batch routing behavior.

- Bug Fixes
  - Improved form validation and error reporting.
  - More consistent map updates when setting origin/destination.
  - Reduced intermittent loading and improved stability during routing calculations.

### DIFF
--- a/packages/transition-frontend/src/components/dashboard/TransitionRightPanel.tsx
+++ b/packages/transition-frontend/src/components/dashboard/TransitionRightPanel.tsx
@@ -46,9 +46,7 @@ const RightPanel: React.FunctionComponent<RightPanelProps> = (props: RightPanelP
                 <TransitRoutingForm availableRoutingModes={props.availableRoutingModes} />
             )}
 
-            {props.activeSection === 'comparison' && (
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):
Ça m'a l'air bien, mais vérifier les commentaires du lapin

PR #1532, opened 2025-10-29
Add option to disable POI calculation in accessibility maps.

GabrielBruno24:
By default, we calculate the number of POIs covered by the accessibility maps. However, this makes the calculation longer. This commit adds a toggle to the accessibility map form that is off by default to calculate them or not. Fix: #1530

image

## Summary by CodeRabbit

* **New Features**
  * Added option to include or exclude points of interest (POIs) in transit accessibility map calculations; option propagated to batch and single-location flows.

* **UI**
  * Added toggle controls in accessibility map and comparison forms to configure POI counting.

* **Localization**
  * Added English and French translations for the POI option.

* **Tests**
  * Updated tests to cover POI-enabled and POI-disabled behaviors.

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
+++ b/packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
@@ -128,6 +128,7 @@ export const batchAccessibilityMap = async (
                     calculationAttributes.arrivalTimeSecondsSinceMidnight = location.timeOfTrip;
                     calculationAttributes.departureTimeSecondsSinceMidnight = undefined;
                 }
+                calculationAttributes.calculatePois = parameters.calculatePois;
 
                 try {
                     const routingResult = await TransitAccessibilityMapCalculator.calculateWithPolygons(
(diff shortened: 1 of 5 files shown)


packages/transition-backend/src/services/accessibilityMap/__tests__/TransitAccessibilityMapCalculator.test.ts
```diff
@@ -304,6 +305,7 @@ describe('Test accessibility map with results', () => {
                 expect(actualString).toContain(expected[j].toString());
             }
         }
+        expect(mockedPlacesPOIsCount).toHaveBeenCalledTimes(0);
```
tahini: Le test ici pourrait explicitement mettre calculatePois à false, sinon, on espère un fallback à `false` (ou alors mentionner qu'on teste le default et ajouter un autre test pour `false`

GabrielBruno24:
> The alignment seems strange in the screenshot, is that expected?

It's also like that in the other tabs, but the surrounding input fields coincidentally align better with it. I agree it looks a little odd.

greenscientist:
> > The alignment seems strange in the screenshot, is that expected?
> 
> It's also like that in the other tabs, but the surrounding input fields coincidentally align better with it. I agree it looks a little odd.

Is that something we can quickly fix here?

GabrielBruno24:
Added another commit to enable and disable the POI in the batch calculation

PR #1550, opened 2025-11-03
batchRoute: handle results through a result visitor

### LINKED ISSUE #1549: batchRouting: support different types of results and decouple its calculation
Currently, after all calculations are done, we generate result files according to what was specified in the parameters (csv, optional detailed and optional geometries). These files are currently the only way to get results from the batch routing job.

But there could be other result processing that could be done. For example, global statistics on the calculation (for example in simulations), various aggregations of results, a posteriori cost functions, etc.

Even files could be generated separately in the future, if, say, someone wants the detailed data after all, instead of automatically from the job.

A result handler could be provided when executing the job, that would be run by default. 

But potentially running result handlers later means they won't be necessarily _in the context_ of a job (ie in the workerpool), unless we enqueue result processing jobs? or something like resuming a completed job at the result generation step? This last possibility needs more discussion and may have its own issue when this one is completed.

tahini:
fixes #1549

Define a `BatchRouteResultVisitor` interface that will visit each result
from the batch calculation and do something about it. The visitor can do
whatever it wants with the result and it does not necessarily involve
generating files.

Move the current result file generation to an implementation of this
visitor: the `BatchRouteFileResultVisitor` class, to generate the
requested files.

The `TrRoutingBatch` class now uses this visitor class to handle results
at the end of the calculation. It does not yet support any other result
handling, but it prepares the way.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal architecture for batch routing result processing to enhance modularity and maintainability. Result handling is now more flexible, allowing for better extensibility while preserving existing functionality and output formats.

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/BatchRoutingJob.ts
+++ b/packages/transition-backend/src/services/transitRouting/BatchRoutingJob.ts
@@ -7,6 +7,7 @@
 import { TransitBatchRoutingDemandAttributes } from 'transition-common/lib/services/transitDemand/types';
 import { TransitBatchCalculationResult } from 'transition-common/lib/services/batchCalculation/types';
 import { BatchCalculationParameters } from 'transition-common/lib/services/batchCalculation/types';
+import { OdTripRouteResult } from './types';
 
 export type BatchRouteJobType = {
     name: 'batchRoute';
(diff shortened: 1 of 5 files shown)


tahini:
By the way, in addition to the simulation vs normal calculation, this result visitor approach opens the door to:

* Cost functions a posteriori on the result, to generate new types of results
* "Profil de charge" on the result data to put people on buses

Calculations are done once, and from the job's result (or jobs', if we eventually support statistical comparisons), any type of calculation can be done above it.

Issue #1553, opened 2025-11-04
network design job: define the types for the Job 


tahini:
Currently, the simulation will describe the basic simulation parameters (what is the expected output, using which network), the algorithm parameters (for example the evolutionary algorithm parameters, like population size, number of generation, crossover parameters, etc) and the transit attributes (transit routing parameters, like access/egress/travel/waiting times).

Then, at runtime when _running_ a simulation, the backend task allows to overwrite the algorithm parameters (to fine-tune and try stuff in various runs).

The run also configures some runtime parameters like:

* The random seed to use (for reproduceability)
* The simulation method details to evaluate the performance of each transit network scenario (using a sample of odTrips and/or based on accessibility calculations)

But,

* is that correct?
* Should we define the simulation methods in the simulation itself (which means we'll have to save the trips and places files somewhere that is not yet associated with a job)
* Which parameter sets should it be possible to modify at runtime? simulation? (I'd say no, just copy and update in another simulation if so), algorithm? (that could be useful for fine-tuning and trying out parameters, for ex with the sample seed and see the effect) transit? (This, I am not sure if it's relevant to modify)

tahini:
About the simulation/evaluation methods, they could be defined at the simulation level (for example to store datasets and their csv definition and avoid having to do this for each run). When configuring the run, we could add the weight of each method, so methods would be a simulation parameters, while the weight would be a runtime one.

kaligrafy:
Simulation = calculate routings for multiple trips/OD etc. in batch
So the whole genetic algorithm should be renamed Optimization or NetworkDesign. Each candidate in each generation will run a simulation to assess its performance

Issue #1533, opened 2025-10-29
genetic algorithm: Let the genetic algorithm use batch routing jobs


tahini:
Instead of the current OD trip version of the trip routing job, we should use batch routing job. The genetic algorithm job will contain the csv file for the OD trips to use.

But since we usually use a sample of the OD trip for simulation purposes (X%), then we should support either sampling in the batch job, or first generate a new od trips file for the corresponding sample of the original file.

PR #1554, opened 2025-11-04
simulation: move method and algorithm types to common

### LINKED ISSUE #1552: simulation: make sure simulation methods and algorithms are described in common and available to frontend
The simulation methods and details on the available algorithms are currently defined in the backend. This means the corresponding descriptors, to get the configuration options, are not available in the frontend, to the UI.

Also, originally, the simulation backend was developed with the idea that we could register new algorithms (other than evolutionary algorithm) and simulation methods (other than od trips and access map) that could come from external modules. This architecture was a bit over-engineered as we do not foresee any of this in a near future (everything would still be implemented in Transition). And it has limitations in our ability to type properly the option. And since the Simulation is still under development and it is not yet a feature available to the user, we might as well consolidate what we have, with the current types and make sure everything works and can be properly typed. Once we have something stable and there's a need for it, we can re-visit and add an algorithm registration API.

tahini:
and drop the registerAlgorithm and registerSimulationMethod functionalities

fixes #1552

For the frontend to be able to configure them, the simulation method types and configuration descriptors should be in the transition-common package.

Also we drop the registration of algorithm and simulation methods. The API of the simulation is not yet stable enough to support dynamically adding methods for now. So we hard-code what is currently implemented in Transition, with the intent of properly typing what is and simplify the upcoming work in making the simulations available from the frontend.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected algorithm type typos in tests and fixed two locale strings.

* **New Features**
  * Added Accessibility Map and OD‑Trip simulation methods with configurable options, async choice loaders, and selectable fitness functions.

* **Refactor**
  * Moved to typed, descriptor-driven registries for algorithms and simulation methods; removed dynamic runtime registration and registration APIs.
  * Simulation runtime options now include thread count.

* **Tests**
  * Updated and added unit tests for descriptors, options, and validation.

### DIFF
--- a/packages/transition-common/src/services/simulation/algorithm/__tests__/EvolutionaryAlgorithm.test.ts
+++ b/packages/transition-common/src/services/simulation/algorithm/__tests__/EvolutionaryAlgorithm.test.ts
@@ -1,0 +1,319 @@
+/*
+ * Copyright 2025, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { EvolutionaryAlgorithmDescriptor, EvolutionaryAlgorithmOptions } from '../EvolutionaryAlgorithm';
+
+describe('EvolutionaryAlgorithmDescriptor', () => {
+    let descriptor: EvolutionaryAlgorithmDescriptor;
+
+    beforeEach(() => {
+        descriptor = new EvolutionaryAlgorithmDescriptor();
+    });
+
+    describe('getTranslatableName', () => {
+        test('should return the correct translatable name', () => {
+            expect(descriptor.getTranslatableName()).toBe('transit:simulation:simulationClasses:LineAndNumberOfVehiclesGASimulation');
+        });
+    });
+
+    describe('getOptions', () => {
+        test('should return all required options', () => {
+            const options = descriptor.getOptions();
+
+            const expectedKeys = [
+                'populationSizeMin',
+                'populationSizeMax',
+                'numberOfElites',
+                'numberOfRandoms',
+                'crossoverNumberOfCuts',
+                'crossoverProbability',
+                'mutationProbability',
+                'tournamentSize',
+                'tournamentProbability',
+                'numberOfGenerations',
+                'shuffleGenes',
+                'keepGenerations',
+                'keepCandidates'
+            ];
+
+            expectedKeys.forEach(key => {
+                expect(options).toHaveProperty(key);
+            });
+        });
+
+        describe('integer options', () => {
+            const integerOptions = [
+                { key: 'populationSizeMin', default: 20, i18nName: 'transit:simulation:PopulationSizeMin' },
+                { key: 'populationSizeMax', default: 20, i18nName: 'transit:simulation:PopulationSizeMax' },
+                { key: 'numberOfElites', default: 2, i18nName: 'transit:simulation:NumberOfElites' },
+                { key: 'numberOfRandoms', default: 0, i18nName: 'transit:simulation:NumberOfRandoms' },
+                { key: 'crossoverNumberOfCuts', default: 1, i18nName: 'transit:simulation:CrossoverNumberOfCuts' },
+                { key: 'tournamentSize', default: 10, i18nName: 'transit:simulation:TournamentSize' }
+            ];
+
+            integerOptions.forEach(({ key, default: defaultValue, i18nName }) => {
+                test(`should configure ${key} correctly`, () => {
+                    const options = descriptor.getOptions();
+                    const option = options[key as keyof typeof options];
+
+                    expect(option.type).toBe('integer');
+                    expect(option.default).toBe(defaultValue);
+                    expect(option.i18nName).toBe(i18nName);
+                    expect(typeof (option as any).validate).toBe('function');
+                });
+            });
+        });
+
+        describe('number options', () => {
+            const numberOptions = [
+                { key: 'crossoverProbability', default: 0.8, i18nName: 'transit:simulation:CrossoverProbability' },
+                { key: 'mutationProbability', default: 0.08, i18nName: 'transit:simulation:MutationProbability' },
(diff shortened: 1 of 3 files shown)


packages/transition-common/src/services/simulation/simulationMethod/OdTripSimulationMethod.ts
```diff
@@ -0,0 +1,85 @@
...
+                // FIXME Still using data source queries. When this code was in the
+                // backend, it used the query to fetch the data source, now let's just
+                // use an empty array (this won't work, but it already doesn't work)
+                const dataSources: { id: string; shortname: string }[] = [];
+                return dataSources.map((ds) => ({
+                    value: ds.id,
+                    label: ds.shortname
+                }));
+            }
+        },
```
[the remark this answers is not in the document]
tahini: We are aware. It was currently not functioning anyway, as there was no way to actually add datasources. Also we plan to move away from the datasource path and use batch calculations instead, so this will be reworked shortly. But thanks for your help.

packages/transition-common/src/services/simulation/simulationMethod/OdTripSimulationMethod.ts
```diff
@@ -0,0 +1,85 @@
...
+        sampleRatio: {
+            i18nName: 'transit:simulation:simulationMethods:OdTripsSampleRatio',
+            type: 'number' as const,
+            validate: (value: number) => value > 0 && value <= 1,
+            default: 1 // right now, it sets the value to NaN if no value are entered... TODO: fix that
+        },
```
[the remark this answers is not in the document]
tahini: removed the comment. It applied to the cli version of the option setter and does not belong in the descriptor

packages/transition-backend/src/services/simulation/methods/AccessibilityMapSimulation.ts
```diff
@@ -8,23 +8,20 @@ import random from 'random';
...
 import { TransitRoutingBaseAttributes } from 'chaire-lib-common/lib/services/routing/types';
 import { SimulationMethodFactory, SimulationMethod } from './SimulationMethod';
-import dataSourceDbQueries from 'chaire-lib-backend/lib/models/db/dataSources.db.queries';
 import placesDbQueries from '../../../models/db/places.db.queries';
 import nodesDbQueries from '../../../models/db/transitNodes.db.queries';
 import { TransitAccessibilityMapCalculator } from '../../accessibilityMap/TransitAccessibilityMapCalculator';
 import TransitAccessibilityMapRouting from 'transition-common/lib/services/accessibilityMap/TransitAccessibilityMapRouting';
 import NodeCollection from 'transition-common/lib/services/nodes/NodeCollection';
+import {
+    AccessMapSimulationDescriptor,
```
kaligrafy: If you want to rename Accessibility by Access, I think you should add some readme somewhere explaining that it is an abbreviation. The term "access" is already used for "access/egress" segments from origin to the first node and from the last node to the destination. I would suggest to use the full Accessibility name instead here for this reason.

packages/transition-common/src/services/simulation/simulationMethod/index.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import type { SimulationAlgorithmDescriptor } from '../SimulationAlgorithm';
+import { AccessMapSimulationDescriptor, type AccessMapSimulationOptions } from './AccessibilityMapSimulationMethod';
+import { OdTripSimulationDescriptor, type OdTripSimulationOptions } from './OdTripSimulationMethod';
+
+export type { AccessMapSimulationOptions } from './AccessibilityMapSimulationMethod';
+export type { OdTripSimulationOptions } from './OdTripSimulationMethod';
+
+// Define the registry of all available simulation methods
+export interface SimulationMethodRegistry {
```
tahini: @kaligrafy @greenscientist question: in #1553 @coderabbitai indirectly suggested to call these `evaluationMethods` instead of `simulationMethods`. Would you agree? I think it makes sense as it is used to rate/evaluate a given transit network scenario. And `Simulation` refers to the high-level concept that uses an algorithm to suggest optimal scenario within certain parameters. Since we're reworking the architecture around the simulation, we might as well work on the correct naming right away, to avoid confusion.

packages/transition-backend/src/services/simulation/SimulationRun.ts
```diff
@@ -93,8 +104,8 @@ export default class SimulationRunBackend extends SimulationRun {
         const methods = Object.keys(runtimeOptions.functions);
         const allResults: { [key: string]: { fitness: number; results: unknown } } = {};
         for (let i = 0; i < methods.length; i++) {
-            const method = methods[i];
-            const factory = SimulationRunBackend.METHODS[method];
+            const method = methods[i] as SimulationMethodType;
+            const factory = METHODS_FACTORY[method];
             if (factory === undefined) {
```
greenscientist: Can this still happen with the way we not fetch the factory method ?

packages/transition-backend/src/services/simulation/SimulationRun.ts
```diff
@@ -108,7 +119,8 @@ export default class SimulationRunBackend extends SimulationRun {
                         );
                     }
                     try {
-                        const simOptions = runtimeOptions.functions[method];
+                        // FIXME Type properly
```
greenscientist: Is it that hard to type properly???

packages/transition-common/src/services/simulation/SimulationRun.ts
```diff
@@ -20,6 +20,10 @@ export type SimulationRuntimeOptions = {
     fitnessSorter: string;
     trRoutingStartingPort: number;
     cachePathDirectory?: string; // the cache directory used to save cache for this simulation run
+    // FIXME type this better (and rename? function is not clear). Also use an
```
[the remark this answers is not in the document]
tahini: It's already included in many issues (#1533, #1553 and #1560)

greenscientist (approved):

PR #1568, opened 2025-11-11
workerpool: add functions to cleanup worker pool

tahini:
Make sure to properly shutdown/terminate workerpool when the main process is terminated.

Also add a worker thread's `onTerminate` handler. It does not do anything yet, just a console.log to see when/if this callback is called in case of hard failure.

## Summary by CodeRabbit

* **New Features**
  * Improved graceful shutdown: process-level termination handlers ensure worker pool and background workers are cleanly stopped on SIGINT/SIGTERM, with errors logged and process exiting after cleanup.
  * Worker runtime now includes an on-terminate cleanup phase so ongoing tasks shut down cleanly without impacting normal startup flow.

### DIFF
--- a/packages/transition-backend/src/server.ts
+++ b/packages/transition-backend/src/server.ts
@@ -15,7 +15,7 @@ import { hideBin } from 'yargs/helpers';
 import express from 'express';
 import { registerTranslationDir, addTranslationNamespace } from 'chaire-lib-backend/lib/config/i18next';
 import config from 'chaire-lib-backend/lib/config/server.config';
-import { startPool } from './tasks/serverWorkerPool';
+import { startPool, terminatePool } from './tasks/serverWorkerPool';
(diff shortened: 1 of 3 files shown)


greenscientist (commented):
Do we need to add a stop action on the Jobs that would act as a cancel, but without cancelling it?

Otherwise looks good, will let you address the rabbit's comments

greenscientist:
> > Do we need to add a stop action on the Jobs that would act as a cancel, but without cancelling it?
> 
> You mean pause a job? #1557? Or you mean when terminating the workerpool for some reason, to "kill" the job? In that case, there's an abort mechanism in the `workerpool` package. I'm not exactly sure how it works, we'll have to experiment with it when we have some time.

Kinda yeah, but with a way to distinguish a pause made from the users or a pause made from the system (for a shutdown for example). So we could restart automatically the task when we can.

PR #1585, opened 2025-11-17
executableJob: support parent/child job relationship

tahini:
Add a `parent_id` field to the Job type

TODO

- [ ] Previously used `on delete cascade` in the migration, but as discussed, there are resources to clean up in the children, so we need to delete them first, so we should use `on delete restrict` instead.
- [ ] In TrRoutingBatch, we should not use the default result visitor if the parent_id is not null, so we can use other visitors instead.

In general, we should support pausing/resuming automatically jobs when they have children to avoid deadlocks. But until we actually have parent/child job, it's not currently necessary.

### DIFF
--- a/packages/transition-backend/src/services/executableJob/ExecutableJob.ts
+++ b/packages/transition-backend/src/services/executableJob/ExecutableJob.ts
@@ -121,6 +121,26 @@ export class ExecutableJob<TData extends JobDataType> extends Job<TData> {
         return job;
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


greenscientist:
> * [ ]  In TrRoutingBatch, we should not use the default result visitor if the parent_id is not null, so we can use other visitors instead.

It should not be automatically based on the parent_id presence. It should be an explicit job parameter. We could have a parent that wants that default behavior.

PR #1595, opened 2025-11-20
db: fix users db test for clean databases

tahini:
Inserting users with previously set ids clash with the auto-increment primary key on first run as IDs have low values.

This removes the pre-set IDs from the user attributes to insert and assigns them at creation time.

Also update deprecated `toThrowError` called to `toThrow`.

## Summary by CodeRabbit

* **Tests**
  * Improved database test suite for greater reliability and clearer structure.
  * Strengthened error expectations and validation of duplicate user constraints.
  * Added coverage for ID handling and case-insensitive email lookups to better reflect real-world flows.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
+++ b/packages/chaire-lib-backend/src/models/db/__tests__/users.db.test.ts
@@ -15,8 +15,7 @@ import { UserAttributes } from '../../../services/users/user';
 
 const testEmail = 'test@test.org';
 
-const user = {
-    id: 1,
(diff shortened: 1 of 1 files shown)


tahini:
> Je suis curieux, au lieu de mettre des ! partout à .id, pourrais-tu le définir first à id: undefined dans la structure originale ?

Si je fais ça, je suis quand même obligée de mettre des ! partout puisque justement, c'est à undefined! Le '!' veut dire "nonon, t'inquiète, je suis sûre que c'est pas undefined à ce point"

PR #1597, opened 2025-11-21
jobs: allow jobs to type their specific internal_data field

tahini:
The internal helps a job manage its internal state. Currently, there is a single `checkpoint` field that allows to resume a `Job` at a specific point. Batch route and accessibility map jobs make use of it.

But there may be additional internal data that a job may store, for example the evolutionary algorithm network design Job may want to save various states and/or calculated cost data for each generation. An additional type gives more flexibility to the job type.

## Summary by CodeRabbit

* **New Features**
  * Jobs now support optional internal state storage with built-in checkpoint tracking to improve recovery and resumability.
  * Job attributes accept custom internal metadata alongside the default checkpoint field for flexible per-job volatile state.

* **Tests**
  * Added tests covering initialization, mutation, and checkpointing of internal job state to validate persistence and recovery behaviors.

### DIFF
--- a/packages/transition-common/src/services/jobs/Job.ts
+++ b/packages/transition-common/src/services/jobs/Job.ts
@@ -10,10 +10,12 @@ export type JobDataType = {
     name: string;
     data: { [key: string]: unknown };
     files: { [fileTitle: string]: boolean };
+    internal_data?: Record<string, unknown>;
 };
 
 export type JobNameKey = 'name';
 export type JobDataKey = 'data';
+export type JobInternalDataKey = 'internal_data';
 export type fileKey = 'files';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/transition-common/src/services/jobs/Job.ts
```diff
@@ -69,11 +72,13 @@ export interface JobAttributes<TData extends JobDataType> {
     status: JobStatus;
     statusMessages?: { errors?: ErrorMessage[]; warnings?: ErrorMessage[]; infos?: ErrorMessage[] };
     /**
-     * Data internal to the job management, that is not relevant to the users
+     * Data internal to the job management, that is not relevant to the users.
+     * Typically volatile during job execution and can be used for example to
```
greenscientist: Not sure the word volatile is the word you want to use here. A checkpoint is definitely something that is not volatile

PR #1598, opened 2025-11-21
batchRoute: use the csv field mapping to describe demand

tahini:
The `TransitOdDemandFromCsv` now inherits the `CsvFileMapping` class
instead of the previous csv file. It calls the constructor with the
desired mapping descriptors. The demand type is updated to be of type
`CsvFileAndMapping`. The backend routes are updated accordingly with the
new types.

The OdTripProvider's option type is updated with the new mapped fields,
but except for naming, it works as before, using the new mappings.

In the frontend, the `BatchAttributesSelection` is not required anymore,
and the `ConfigureDemandFromCsvForm` component can be replaced with the
`GenericCsvImportAndMappingForm`.

## Summary by CodeRabbit

* **Improvements**
  * Updated translation keys to clearer field-oriented labels for id, origin/destination lon/lat and time mapping.
  * Replaced legacy CSV demand workflow with a generic, strongly-typed CSV import & mapping UI; CSVs may come from uploads or previous jobs.
  * Removed calculationName from batch results (cleaner, smaller result payload).

* **New Features**
  * Added explicit delta-related validation messages for batch routing (interval/size constraints).

* **Tests**
  * Tests updated to the new CSV mapping shapes and result expectations.

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
+++ b/packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
@@ -194,8 +194,6 @@ class TrRoutingBatch {
             this.options.progressEmitter.emit('progress', { name: 'GeneratingBatchRoutingResults', progress: 1.0 });
 
             const routingResult = {
-                // FIXME The calculation name is never set anyway, maybe remove? or define?
-                calculationName: 'not set',
                 detailed: this.job.attributes.data.parameters.transitRoutingAttributes.detailed,
                 completed: true,
                 errors: [],
@@ -208,8 +206,6 @@ class TrRoutingBatch {
             if (Array.isArray(error)) {
                 console.log('Multiple errors in batch route calculations for job %d', this.job.id);
                 return {
-                    // FIXME The calculation name is never set anyway, maybe remove? or define?
-                    calculationName: 'not set',
                     detailed: false,
                     completed: false,
                     errors: error,
--- a/packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
+++ b/packages/transition-backend/src/services/transitRouting/TrAccessibilityMapBatch.ts
@@ -180,7 +180,6 @@ export const batchAccessibilityMap = async (
         Object.assign(files, resultProcessor.getFiles());
 
         return {
-            calculationName: parameters.calculationName,
             detailed: parameters.detailed,
             completed: true,
             errors: [],
@@ -190,7 +189,6 @@ export const batchAccessibilityMap = async (
     } catch (error) {
         if (Array.isArray(error)) {
             return {
-                calculationName: parameters.calculationName,
                 detailed: false,
                 completed: false,
                 errors: error,
--- a/packages/transition-backend/src/services/transitRouting/__tests__/TrAccessibilityMapBatch.test.ts
+++ b/packages/transition-backend/src/services/transitRouting/__tests__/TrAccessibilityMapBatch.test.ts
@@ -112,7 +112,6 @@ const mockJobAttributes = {
         parameters: {
             batchAccessMapAttributes: {
                 type: 'csv' as const,
-                calculationName: 'test',
                 projection: 'test',
                 idAttribute: 'id',
                 xAttribute: 'origX',
@@ -163,7 +162,6 @@ test('3 locations, all successful', async() => {
     mockedParseLocations.mockResolvedValue({ locations, errors: [] });
     const result = await batchAccessibilityMap(mockedJob, progressEmitter, isCancelledMock);
     expect(result).toEqual({
-        calculationName: mockJobAttributes.data.parameters.batchAccessMapAttributes.calculationName,
         detailed: mockJobAttributes.data.parameters.batchAccessMapAttributes.detailed,
         completed: true,
         errors: [],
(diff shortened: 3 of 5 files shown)


kaligrafy (approved):

packages/transition-backend/src/services/executableJob/ExecutableJob.ts
```diff
@@ -50,6 +51,37 @@ export class ExecutableJob<TData extends JobDataType> extends Job<TData> {
...
+     * @returns
+     */
+    static handleJobFile = async (
+        fileLocation: FileConfig,
+        userId: number,
+        importFileName: string
+    ): Promise<string | { filepath: string; renameTo: string }> => {
+        if (fileLocation.location === 'upload') {
+            return {
+                filepath: `${directoryManager.userDataDirectory}/${userId}/imports/${importFileName}`,
```
greenscientist: Since it does not operate on a job directory, I don't think this should be part of the ExecutableJob, 

ExecutableJob should just get a path at this point and not be aware of a specific imports directory.

packages/transition-common/src/services/csv/CsvFieldMapper.ts
```diff
@@ -0,0 +1,292 @@
...
+                        params: { field: fieldMappingKey }
+                    }
+            ];
+        } else if (options.isCsvField !== false && !this._csvFields.includes(mappedField)) {
+            return [{ text: 'main:errors:csv:FieldNotInCsvFile', params: { field: mappedField } }];
+        }
+        return true;
+    }
+
+    private _descriptorTypeSpecificValidation = (descriptor: CsvFieldMappingDescriptor): true | ErrorMessage[] => {
```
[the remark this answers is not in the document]
tahini: Later... when we have actual use cases for this.

packages/transition-backend/src/api/services.socketRoutes.ts
```diff
@@ -194,35 +191,18 @@ export default function (socket: EventEmitter, userId?: number) {
     if (userId !== undefined) {
         socket.on(
             TrRoutingConstants.BATCH_ROUTE,
-            async (
-                parameters: TransitBatchRoutingDemandAttributes,
-                transitRoutingAttributes: BatchCalculationParameters,
-                callback
-            ) => {
+            async (parameters: CsvFileAndMapping, transitRoutingAttributes: BatchCalculationParameters, callback) => {
```
greenscientist: I think we should have something properly defined for the CsvMapping, having a specific type which define which field we expect in the CSV file at this point. So we know we have all the right field already. 
We are losing the definition of the parameters that were in TransitBatchRoutingDemandAttributes

packages/transition-backend/src/services/odTrip/odTripProvider.ts
```diff
@@ -21,32 +21,32 @@ import Preferences from 'chaire-lib-common/lib/config/Preferences';
 import TrError, { ErrorMessage } from 'chaire-lib-common/lib/utils/TrError';
 import constants from 'chaire-lib-common/lib/config/constants';
 
-export interface OdTripOptions {
+export type OdTripCsvMapping = {
```
greenscientist: This should inherit from some type from CsvMapping

packages/transition-backend/src/services/transitRouting/BatchRoutingJob.ts
```diff
@@ -13,7 +13,7 @@ export type BatchRouteJobType = {
     name: 'batchRoute';
     data: {
         parameters: {
-            demandAttributes: TransitBatchRoutingDemandAttributes;
+            demandAttributes: CsvFileAndMapping;
```
greenscientist: Again, the type is not specific enough

tahini:
I updated the `CsvFieldMapper` to add a generic type, that defaults to `Record<string, string>` to allow to type the field mapping, which allows to not lose the type mapping of the OD demand.

packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -194,7 +194,8 @@ class TrRoutingBatch {
             this.options.progressEmitter.emit('progress', { name: 'GeneratingBatchRoutingResults', progress: 1.0 });
 
             const routingResult = {
-                calculationName: parameters.calculationName,
+                // FIXME The calculation name is never set anyway, maybe remove? or define?
```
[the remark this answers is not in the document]
greenscientist: @tahini You do not want to remove it now? Will be a later PR? Seem an easy clean up

packages/transition-common/src/services/transitDemand/TransitDemandFromCsv.ts
```diff
@@ -21,6 +21,8 @@ export interface DemandCsvAttributes extends GenericAttributes, Partial<TransitD
     /* Nothing else to add */
 }
 
+// TODO This class still exists for the accessibility map, but it should be
+// replaced with the new CSV demand importer structure
```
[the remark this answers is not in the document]
tahini: En fait, non, ce n'est pas le même TODO et ceci ne fait pas partie [encore] de l'algo génétique. C'est plutôt pour la carte d'accessibilité. En faisant cette PR, nous étions pressés par le temps en pensant qu'on devait mettre dans main tout l,algo génétique. Y allant maintenant d'un feature branche, je pourrais faire ce TODO comme 2e PR, une fois celle-ci complétée et acceptée, question d'avoir un peu plus de lignes de codes négatives. Ça ne paraît pas, mais l'approche par CsvFieldMapper devrait permettre d'enlever beaucoup de codes à plusieurs endroits et, oui, éventuellement avoir des lecteurs de mapping de fichier directement des descripteurs.

tahini:
I removed the calculation name. If we want it back someday, we'll bring it back

PR #1599, opened 2025-11-24
evolutionary algorithm: Add types for the network design job

tahini:
Make the OdTripSimulationMethod types in line with the batchRoute job

Add specific types for the transit network design based on the evolutionary algorithm


tahini:
Closing, we'll productify the network design branch from #1600, all in good times

PR #1625, opened 2025-12-03
workerpool: do not check disk usage of started jobs

### LINKED ISSUE #1616: workerPool: do not check disk quota for jobs that were already in progress
Only when a pending job is set to run should we validate that there is enough disk space for the job. Otherwise, a job that may produce large files (like the evolutionary algorithm with the cache) may be able to start fine, but hit the disk quota when resumed. If the job started, it should be allowed to complete without checking the quota.

tahini:
fixes #1616

When the job was first set in progress, the disk usage was already asserted and was correct. During the job's execution, the disk usage may reach and go over the limit, but it should not prevent the job from being resumed when the server restarts.

## Summary by CodeRabbit

* **Refactor**
  * Tasks now check disk quota before starting: if quota is insufficient the task is paused with a UserDiskQuotaReached status, saved once, and is not marked in-progress or routed for execution. When quota is sufficient, tasks are marked in-progress before continuing.

* **Tests**
  * Tests expanded to cover quota-driven flows for pending and in-progress inputs, asserting pause behavior, single save, and no routing when quota is reached.

### DIFF
--- a/packages/transition-backend/src/tasks/TransitionWorkerPool.ts
+++ b/packages/transition-backend/src/tasks/TransitionWorkerPool.ts
@@ -52,15 +52,12 @@ function taskUpdateListener() {
     return eventEmitter;
 }
 
-const assertDiskUsage = (task: ExecutableJob<JobDataType>) => {
+const isDiskSpaceSufficient = (task: ExecutableJob<JobDataType>) => {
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):

tahini:
Changed the logic to pause jobs when quota is reached at exec time

PR #1633, opened 2025-12-05
Update zoom level visibility for rail

greenscientist:
Rail often span large region, so to be able to visualize them they should be viewable at smaller zoom level.

Setting it to 5, good for multiprovincial project. Ideally could be set at 4.5, but keeping whole number for now. Needed to update the global minzoom also.

## Summary by CodeRabbit

* **New Features**
  * Transit paths and rail services are now visible at lower zoom levels, improving map discoverability when viewing at a zoomed-out perspective.

### DIFF
--- a/packages/transition-frontend/src/config/layers.config.ts
+++ b/packages/transition-frontend/src/config/layers.config.ts
@@ -277,12 +277,12 @@ const layersConfig = {
 
     transitPaths: {
         type: 'line',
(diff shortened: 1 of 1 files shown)


tahini (approved):
ça me va, je vais laisser @kaligrafy décider s'il a une opinion plus argumentée sur le sujet

kaligrafy (approved):

PR #1636, opened 2025-12-05
executableJob: do not check disk quota at creation time

### LINKED ISSUE #1635: executableJob: a job is not created if at creation time, the disk quota is reached
A bit related to #1616, which validated the the disk usage when the job is being run. But since this issue was fixed by pausing the job, there is no reason to validate the disk space when creating a job, we should simply create the job and the disk usage will be checked and the job paused, if reached, when queued.

Additionnally, currently, the reason why the job was not created does not trickle to the UI, so the user has no feedback. One more reason to follow the current flow.

tahini:
fixes #1635

Now that jobs are paused before running when the user's disk quota is reached, there is no reason to prevent the job creation if the quota is reached.

Also the `hasOutputFiles` creation parameter is not necessary anymore, so it has been removed from the job's initial data type.

## Summary by CodeRabbit

* **Changes**
  * Removed disk quota validation for job output files from the job execution system. Jobs will no longer be subject to disk usage checks when producing output files.


greenscientist (approved):
Seems to make sense

PR #1642, opened 2025-12-10
Add accessibility map POIs to API

GabrielBruno24:
The accessibility map API call can now return the amount of contained POIs within the calculated polygon. Also updates the documentation.

Tested with a python script and works as intended.

## Summary by CodeRabbit

* **New Features**
  * Optional "calculatePois" parameter added to accessibility map requests (defaults to false) to enable POI calculations for polygons.
  * Accessibility map results now include POI counts per polygon, broken down by category and by detailed category for richer insights.


tahini (commented):
ça me semble bon, mais j'ai l'impression que ça prendrait des tests en qq part, non?

PR #1671, opened 2025-12-18
dep: upgrade version of express from 4.21.2 to 5.2.1

tahini:
Also upgrade

* @types/express: 4.17.21 => 5.0.6
* supertest: 7.0.0 => 7.1.4

Wildcard paths now should use regexes instead of strings with wildcards.

Also the supertest post expects to post something, so we send empty objects in tests where nothing was sent to make sure the body exists.

## Summary by CodeRabbit

* **Chores**
  * Updated core dependencies to latest stable versions for improved compatibility and performance.
  * Refined internal routing mechanisms to enhance system stability without affecting user-facing functionality.

### DIFF
--- a/packages/chaire-lib-backend/src/api/__tests__/auth.routes.test.ts
+++ b/packages/chaire-lib-backend/src/api/__tests__/auth.routes.test.ts
@@ -251,6 +251,7 @@ test('Reset password, post without password', async () => {
     const res = await request(app)
         .post(`/reset/${token}`)
         .set('Accept', 'application/json')
+        .send({ })
         .expect('Content-Type', /json/)
         .expect(200);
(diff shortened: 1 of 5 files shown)


packages/chaire-lib-backend/src/api/__tests__/auth.routes.test.ts
```diff
@@ -251,6 +251,7 @@ test('Reset password, post without password', async () => {
     const res = await request(app)
         .post(`/reset/${token}`)
         .set('Accept', 'application/json')
+        .send({ })
```
GabrielBruno24: Is there any spot outside the tests that requires sending an empty object?

PR #1719, opened 2026-01-21
socket client: attempt to reconnect forever, every minute

### LINKED ISSUE #155: transition: Client not reconnecting to server when offline
When the server goes offline, the client displays a disconnected message. If the server comes back up fast enough, the client connects back to it, but if it takes a few minutes, then the client does not connect to it anymore. It should keep polling, even if less frequently, but it should come back at some point

tahini:
fixes #155

Do not cap the maximum number of socket connection attempts if the server if down. Default behavior of socket.io is to try at 1 second delay at first, and double that time at each failure until the max delay is reached. So if the reconnection does not happen fast, it tries every minute to not overload the page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved socket reconnection strategy to use capped exponential backoff with a maximum delay of 120 seconds for more robust connection recovery.
  * Ensured socket event listeners are reliably detached during cleanup to prevent duplicate handlers or leaks.

* **Refactor**
  * Improved socket-handling safety and typings for more stable, maintainable client behavior.

### DIFF
--- a/packages/transition-frontend/src/components/dashboard/TransitionDashboard.tsx
+++ b/packages/transition-frontend/src/components/dashboard/TransitionDashboard.tsx
@@ -6,7 +6,7 @@
  */
 import React, { PropsWithChildren } from 'react';
 import _get from 'lodash/get';
-import io from 'socket.io-client';
+import io, { Socket } from 'socket.io-client';
 import { withTranslation, WithTranslation } from 'react-i18next';
 import EventEmitter from 'events';
 import { sectionLayers } from '../../config/layers.config';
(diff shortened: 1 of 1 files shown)


greenscientist (commented):
Can we make the max higher that 1 minutes, maybe 5 ? 

Thinking of when I'm on an airplane, do not want to waste too much CPU...

greenscientist:
J'avoue que 5 ca pourrait être beaucoup, mais disons que 2 pourrait être un compromis

tahini:
any review for this PR please? @kaligrafy @greenscientist

PR #1720, opened 2026-01-21
Move memcached management outside of TrRoutingProcessManager

greenscientist:
We create a new MemcachedProcessManager, that will return a MemcachedInstance object, when we start it. We can then get the status, serverUrl and stop it from this object.

TrRoutingBatchManager is now responsible to start and stop memcached, unless the jobs have received a memcachedServer string to use when starting trRouting.

This will allow to start a memcached and have several jobs use the same one. We are planning to use this in the Genetic algorithm simulations

## Summary by CodeRabbit

* **New Features**
  * Added a Memcached process manager and support for passing an optional Memcached server to batch routing jobs; the system will use a provided server or manage one automatically. Batch start APIs and managers accept routing options to control memcached usage.
* **Tests**
  * Added extensive unit tests covering Memcached integration, batch start/stop flows, lifecycle ownership, error handling, and routing-option propagation.

### DIFF
--- a/packages/chaire-lib-backend/src/utils/processManagers/MemcachedProcessManager.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/MemcachedProcessManager.ts
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2026, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import ProcessManager from './ProcessManager';
+
+const DEFAULT_PORT = 11212; // 11211 is the default memcached port, we use +1 to not clash
+const TAG_NAME = 'memcached';
+
+const getServiceName = function (port: number) {
+    return `memcached${port}`;
+};
+
+/**
+ * Represents a running memcached instance.
+ * Use this object to get the server connection string and to stop the instance.
+ */
+export class MemcachedInstance {
+    constructor(private port: number) {
+        /* Nothing to do */
+    }
+
+    /**
+     * Get the server connection string for this memcached instance.
+     * @returns Server string in format 'localhost:port'
+     */
+    getServer(): string {
+        return `localhost:${this.port}`;
+    }
+
+    /**
+     * Check if the memcached instance is running.
+     * @returns 'running' if the process is running, 'not_running' otherwise
+     */
+    async status(): Promise<'running' | 'not_running'> {
+        const isRunning = await ProcessManager.isServiceRunning(getServiceName(this.port));
+        return isRunning ? 'running' : 'not_running';
+    }
+
+    /**
+     * Stop this memcached instance.
+     * @returns Status of the stop operation
+     */
+    async stop(): Promise<{ status: 'stopped' | 'not_running' | 'error'; error?: unknown }> {
+        return ProcessManager.stopProcess(getServiceName(this.port), TAG_NAME);
+    }
+}
+
+/**
+ * Start a memcached instance.
+ *
+ * @param options - Optional configuration
+ * @param options.port - Port to run memcached on (default: 11212)
+ * @returns A MemcachedInstance object if started successfully, null otherwise
+ */
+const start = async (options?: { port?: number }): Promise<MemcachedInstance | null> => {
+    const port = options?.port ?? DEFAULT_PORT;
+
+    const processStatus = await ProcessManager.startProcess({
+        serviceName: getServiceName(port),
+        tagName: TAG_NAME,
+        command: 'memcached',
+        commandArgs: [
+            `--port=${port}`,
+            '--user=nobody', // Memcached does not want to run as root, let's drop to nobody
+            '-vv' // Enable detailed output for logging
+        ],
+        waitString: '',
+        useShell: false,
+        attemptRestart: false
+    });
+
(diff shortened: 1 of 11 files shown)


packages/chaire-lib-backend/src/utils/processManagers/MemcachedProcessManager.ts
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright 2025, Polytechnique Montreal and contributors
```
tahini: 2026

packages/chaire-lib-backend/src/utils/processManagers/MemcachedProcessManager.ts
```diff
@@ -0,0 +1,83 @@
...
+const SERVICE_NAME = 'memcached';
+const TAG_NAME = 'memcached';
+
+/**
+ * Represents a running memcached instance.
+ * Use this object to get the server connection string and to stop the instance.
+ */
+export class MemcachedInstance {
+    constructor(private port: number) {}
+
```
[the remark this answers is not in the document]
tahini: You can also simply add a comment `/* Nothing to do */` in the constructor

packages/chaire-lib-backend/src/utils/processManagers/__tests__/MemcachedProcessManager.test.ts
```diff
@@ -0,0 +1,169 @@
+/*
+ * Copyright 2025, Polytechnique Montreal and contributors
```
tahini: 2026

packages/transition-backend/src/services/transitRouting/TrRoutingBatchManager.ts
```diff
@@ -28,12 +31,33 @@ export type TrRoutingBatchStartResult = {
...
+ *   stops it when stopBatch is called.
+ *
  * TODO: Class name is a bit confusing, we should find something better, but it works for now.
  *       We should return an instance which represent the TrRouting instance, like the OSRMMode for OSRM,
  *       Instead of just returning a port number
  */
 export class TrRoutingBatchManager {
-    constructor(private progressEmitter: EventEmitter) {
+    private memcachedInstance?: MemcachedInstance;
+    private ownsMemcached = false;
```
tahini: Do you need both properties? Could it be possible that there is a memcachedInstance, but ownsMemcached is false?

packages/transition-backend/src/services/transitRouting/__tests__/TrAccessibilityMapBatch.test.ts
```diff
@@ -47,6 +48,16 @@ jest.mock('chaire-lib-backend/lib/utils/processManagers/TrRoutingProcessManager'
...
 const mockedStopBatch = TrRoutingProcessManager.stopBatch as jest.MockedFunction<typeof TrRoutingProcessManager.stopBatch>;
 
+jest.mock('chaire-lib-backend/lib/utils/processManagers/MemcachedProcessManager', () => ({
+    start: jest.fn().mockResolvedValue({
+        getServer: () => 'localhost:11212',
+        status: jest.fn().mockResolvedValue('running'),
+        stop: jest.fn().mockResolvedValue({ status: 'stopped' })
+    })
+}));
+const mockMemcachedStart = MemcachedProcessManager.start as jest.MockedFunction<typeof MemcachedProcessManager.start>;
```
tahini: Since you added the memcached stuff in this test, I would expect some diff that includes the memcached parameters, like there is in the `TrRoutingBatch.test.ts` file.

packages/transition-backend/src/services/transitRouting/TrRoutingBatchManager.ts
```diff
@@ -43,8 +67,13 @@ export class TrRoutingBatchManager {
      * Port, debug settings, and max parallel threads are determined by
      * TrRoutingProcessManager from server configuration.
      *
+     * Memcached is handled as follows:
+     * 1. If options.memcachedServer is provided, use it
+     * 2. Else if constructor options.memcachedServer was provided, use it
```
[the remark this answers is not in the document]
tahini: Please try to remember and have a good explanation for why the options here. The way I see it, if you call it with options, after having called it without options, for example, it may override the object's `memcachedInstance` and `ownsMemcached` and you may have a rogue memcached instance lost somewhere. You should not be able to call the `startBatch` function more than once with different options. Plus with the service name without a port, you may end up with an instance running on a different port than the one you want.

greenscientist:
(still some work to do, push this version to close some of the comments)

packages/chaire-lib-backend/src/utils/processManagers/MemcachedProcessManager.ts
```diff
@@ -0,0 +1,90 @@
...
+    if (processStatus.status === 'already_running') {
+        console.error('a memcached process already exist for port:', port);
+    } else if (processStatus.status === 'error' && processStatus.error?.code === 'ENOENT') {
+        console.error('memcached executable does not exist in path');
+    } else {
+        console.error('cannot start memcached:', processStatus);
+    }
+
+    return null;
+};
```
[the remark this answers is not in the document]
greenscientist: yeah, sorry, I wanted to add a comment but forgot!

PR #1744, opened 2026-01-29
Add logging to OSRM data preprocessing and display log path on failure

### LINKED ISSUE #658: osrm: When the osrm data preparation fails, show the path to full logs
Now, if the `prepareOsmNetworkData` fails for a mode, the console just says it fails, but it should show the path where the complete log is.

But first, make sure there actually is logging for this kind of data. Looking at the OSRMServicePreparation.ts file, it is uncertain that there is such a log file. If not, it should be added, in the executable's 'data' listener. (see the ProcessManager.ts file for example of using loggers).

thomasspina:
Closes #658

## Summary

Added logging to OSRM data preprocessing and ensured the log file path is returned when preparation fails. Test file has a lot of changes because style formatter switched from 2-spaces tabs to 4-spaces tabs as is the norm in the other files.

## Changes

**Added file logging for OSRM operations:**
- Created a Winston logger that writes to `logs/osrm_prep_logs/` with separate log files for each operation and mode (e.g., `OSRM_EXTRACT_walking.log`, `OSRM_CONTRACT_cycling.log`)

**Updated `extract()` and `contract()` functions:**
- Refactored function signatures to use individual parameters instead of a generic `parameters` object
- Added logging for stdout and stderr output from OSRM processes
- Added `logFile` property to result objects so callers can display the path on failure
- Fixed function signatures in test files.

**Defined proper TypeScript interface:**
- Added `OSRMResult` interface to replace the previous `any` return types, addressing the TODO comments

## Summary by CodeRabbit

* **New Features**
  * Per-run, mode-specific OSRM preparation logs with log-file references returned in results.
  * Parameter-driven prepare/extract/contract flows for clearer per-mode operations.

* **Bug Fixes**
  * Improved error reporting to surface full preparation logs for easier troubleshooting.

* **Refactor**
  * Centralized file handling and simplified initialization to reduce setup complexity.

* **Tests**
  * Expanded, modernized, and reorganized tests covering multi-mode sequencing and error scenarios.

### DIFF
--- a/packages/chaire-lib-backend/src/utils/processManagers/__tests__/OSRMProcessManager.test.ts
+++ b/packages/chaire-lib-backend/src/utils/processManagers/__tests__/OSRMProcessManager.test.ts
@@ -5,9 +5,9 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 import path from 'path';
-var mockSpawn = require('mock-spawn');
+const mockSpawn = require('mock-spawn');
 
-var osrmSpawn = mockSpawn();
+let osrmSpawn = mockSpawn();
 require('child_process').spawn = osrmSpawn;
 
 import { directoryManager } from '../../filesystem/directoryManager';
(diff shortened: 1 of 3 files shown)


packages/chaire-lib-backend/src/utils/processManagers/__tests__/OSRMProcessManager.test.ts
```diff
@@ -30,257 +30,255 @@ jest.setTimeout(30000);
 const existingOsmFilePath = path.normalize(`${__dirname}/../../../../../../tests/files/osm_network_data.osm`);
 
 // Clone the original config to reset it after each test
-let originalTestConfig = _cloneDeep(config);
+const originalTestConfig = _cloneDeep(config);
```
tahini: The diff for this test file is hard to read, could you put a comment where you actually did modify the file, or even better next time you end up reformating a test file like this, make it in a separate commit, such that one commit is _only_ the reformating and another commit that has your actual contribution. This makes reviews easier, commit by commit

packages/chaire-lib-backend/src/utils/processManagers/OSRMServicePreparation.ts
```diff
@@ -4,32 +4,63 @@
...
+            } as any)
+        ]
+    });
+};
+
+interface OSRMResult {
+    status: 'extracted' | 'contracted' | 'prepared' | 'error';
+    service: 'osrm';
+    action: string;
+    modes: RoutingMode[];
```
tahini: Je crois que dans ce cas, c'est un seul mode par extract, non? donc juste `RoutingMode` ici?

packages/chaire-lib-backend/src/utils/processManagers/OSRMServicePreparation.ts
```diff
@@ -88,33 +127,43 @@ const extract = function (parameters: {
...
             });
         });
     });
 };
 
-//TODO set type for parameters instead of any
-const contract = function (parameters = {} as any): Promise<any> {
-    const mode = parameters.mode || 'walking';
-    const osrmDirectoryPath = getOsrmDirectoryPathForMode(mode, parameters.directoryPrefix);
+const contract = function (mode: RoutingMode | 'walking', directoryPrefix?: string): Promise<OSRMResult> {
```
tahini: Idem ici, svp garder la signature initiale des paramètres en objet, mais merci de les typer!! :)

packages/chaire-lib-backend/src/utils/processManagers/__tests__/OSRMProcessManager.test.ts
thomasspina: I split the reformatting of the files into another commit. All that changes are the linter's effects. All tests work the same otherwise

PR #1763, opened 2026-02-04
dateTimeUtils: improve the `toXXhrYYminZZsec` function

### LINKED ISSUE #1761: transitRouting: duration display drops minutes if 0
See screenshot here. It gives the false impression that it takes 1h32, but it's actually 1h and 32 seconds (as the units show), ie 61 minutes

Image

tahini:
fixes #1761

The function now receives a `t` function instead of only the translated units to allow to pluralize the units depending on the count of each value.

Pass the units and other options as a third object parameter, with named fields.

The `withSeconds` field allows to specify if we want to show the seconds in the display, or stop at the minutes.

A `zeroText` option is returned when all the values to be displayed are at 0.

Add unit tests for all the new use cases provided by this function.

Update the `DurationUnitFormatter` component to use the new function signature.

## Summary by CodeRabbit

* **Chores**
  * Enhanced internationalization support for duration formatting across the application
  * Improved localization of time unit displays and zero-value handling in duration components

### DIFF
--- a/packages/chaire-lib-common/src/utils/DateTimeUtils.ts
+++ b/packages/chaire-lib-common/src/utils/DateTimeUtils.ts
@@ -8,6 +8,7 @@ import _isFinite from 'lodash/isFinite';
 import _isString from 'lodash/isString';
 import _padStart from 'lodash/padStart';
 import { _isBlank } from './LodashExtensions';
+import type { TFunction } from 'i18next';
 
 /**
  * Formats the number of seconds since midnight to a time string (HH:MM[:ss]).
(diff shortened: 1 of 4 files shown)


packages/chaire-lib-common/src/utils/DateTimeUtils.ts
```diff
@@ -118,32 +119,65 @@ const hoursToSeconds = function (hours) {
...
- * @param {string} hourUnit - The unit for hours (e.g., "hr").
- * @param {string} minuteUnit - The unit for minutes (e.g., "min").
- * @param {string} secondsUnit - The unit for seconds (e.g., "sec").
- * @returns {string} - The formatted time string or an empty string if the input is invalid.
+ * @param {TFunction} t - The translation function to localize unit strings.
+ * @param {Object} options - Formatting options.
+ * @param {string} options.hourUnit - The unit for hours, in a localizable form. The value of the hour will be added as the 'count' option for translation.
+ * @param {string} options.minuteUnit - The unit for minutes, in a localizable form. The value of the minute will be added as the 'count' option for translation.
+ * @param {string} options.secondsUnit - The unit for seconds, in a localizable form. The value of the second will be added as the 'count' option for translation.
+ * @param {boolean} [options.withSeconds=true] - Whether to include seconds in the output.
```
tahini: Should I round the times to the next minute if there are seconds in the time and we don't want seconds? 5 minutes 2 seconds, or 5 minutes 58 seconds should both be 6 minutes? The current implementation here gives 5 in both cases

PR #1772, opened 2026-02-10
transitRouting: Add button to sort alternatives by trip length

### LINKED ISSUE #1511: transitRouting: add an option to sort alternatives by trip length
> A user requested that the alternatives could be sorted by travel time instead of using the default trRouting's return order. Would that be a good first step? We could add a sort button to the result component that would re-order the alternatives by travel time, such that the first would be the short one. 

 _Originally posted by @tahini in [#1014](https://github.com/chairemobilite/transition/issues/1014#issuecomment-3386685093)_

Ayelander:
Closes #1511

Add a toggle button to the routing results navigation that reorders alternative paths by ascending travel time. When active (green), the shortest trip appears first. Toggling off reverts to the original order.

## Summary by CodeRabbit

* **New Features**
  * Added duration-based sorting for routing alternatives with a toggle button (visual state shows active sorting and display resets to the first alternative when enabled).
  * Left/right navigation and the displayed alternative counter now follow the sorted display order.

* **Documentation**
  * Added English and French labels for the sort control ("Sort by travel time").

### DIFF
--- a/locales/en/transit.json
+++ b/locales/en/transit.json
@@ -436,6 +436,9 @@
             "horse": "Horses",
             "transit": "Transit"
         },
+        "sort": {
+            "TravelTime": "Sort by travel time"
+        },
         "Acceleration": "Acceleration [m/s²]",
         "Deceleration": "Deceleration [m/s²]",
         "DefaultAcceleration": "Default acceleration [m/s²]",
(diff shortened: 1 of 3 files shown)


greenscientist:
I have not checked the code, but is the design exensible so that we can have other way to sort? Number of transfer could be an interesting one to have. Another one I can think of is longest walk. Don't need them now, but if we can have space to easily add later.

PR #1804, opened 2026-02-20
routing: log error code instead of whole error

tahini:
A previous commit caused the whole error object to be displayed, which is very verbose in the logs, only the code is necessary for TrError to be explanatory. For other errors, having the whole error might still be useful.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error logging in routing functionality to display specific error codes instead of full error objects for better error tracking and debugging.


greenscientist (commented):
How very verbose are we talking about? Also are we losing some detailled parameters that might not be encoded in the code?

PR #1809, opened 2026-02-23
gtfs export: Add newline between batch when exporting schedule

greenscientist:
With the change in 4767e2cba gtfs: export schedules in chunks, we were missing a newline between each batch, which had the effect of putting 2 entries on the same line.

This change as the effect of always adding a newline at the end of the stop_times.txt and trips.txt files. The GTFS standard accept both files with and without a newline, so this does not changes the validity of the export.

We added new expect to catch the problem of the lack of newline between chunk. And adjust other tests to account for the new added newline at the end  of the file.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed GTFS schedule export CSV formatting to consistently append a trailing newline between consecutive export batches for trips and stop times. This ensures proper separation and correct line counts when concatenating outputs, improving file integrity and compatibility with import and third‑party processing tools.

### DIFF
--- a/packages/transition-backend/src/services/gtfsExport/__tests__/ScheduleExporter.test.ts
+++ b/packages/transition-backend/src/services/gtfsExport/__tests__/ScheduleExporter.test.ts
@@ -266,7 +266,7 @@ test('Test exporting one schedule', async () => {
         `"${lineId}","${sluggedServiceId}","${scheduleAttributes1.periods[0].trips[1].id}","${pathAttributes.name}",,0,,"${pathAttributes.id}",,`,
         `"${lineId}","${sluggedServiceId}","${scheduleAttributes1.periods[0].trips[2].id}","${pathAttributes.name}",,0,,"${pathAttributes.id}",,`,
         `"${lineId}","${sluggedServiceId}","${scheduleAttributes1.periods[1].trips[0].id}","${pathAttributes.name}",,0,,"${pathAttributes.id}",,`,
-    ].join('\n'));
+    ].join('\n') + '\n');
     expect(mockWriteStopTimeStream.write).toHaveBeenCalledTimes(1);
     // Match strings for each trip individually, to better catch errors in one trip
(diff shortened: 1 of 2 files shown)


packages/transition-backend/src/services/gtfsExport/__tests__/ScheduleExporter.test.ts
```diff
@@ -338,12 +338,12 @@ test('Test exporting multiple schedules', async () => {
     expect((response as any).pathIds).toEqual([pathAttributes.id, pathAttributes2.id]);
 
     expect(mockWriteTripStream.write).toHaveBeenCalledTimes(1);
-    // Make sure the number of lines matches (6 trips + header)
-    expect((mockWriteTripStream.write.mock.calls[0][0] as string).split('\n').length).toEqual(7);
+    // Make sure the number of lines matches (6 trips + header + last newline)
+    expect((mockWriteTripStream.write.mock.calls[0][0] as string).split('\n').length).toEqual(8);
```
GabrielBruno24: Maybe you should write the 8 as an arithmetic operation instead of directly, like you do later in the file with things like `4 * 100 + 1 + 1`

packages/transition-backend/src/services/gtfsExport/__tests__/ScheduleExporter.test.ts
```diff
@@ -338,12 +338,12 @@ test('Test exporting multiple schedules', async () => {
...
 
     expect(mockWriteStopTimeStream.write).toHaveBeenCalledTimes(1);
-    // Make sure the number of lines matches (6 trips * 4 stops + header)
-    expect((mockWriteStopTimeStream.write.mock.calls[0][0] as string).split('\n').length).toEqual(25);
+    // Make sure the number of lines matches (6 trips * 4 stops + header + last newline)
+    expect((mockWriteStopTimeStream.write.mock.calls[0][0] as string).split('\n').length).toEqual(26);
```
GabrielBruno24: Same here

tahini (approved):

Issue #29, opened 2021-02-08
UI: Add themes and refactor color palettes (including dark and light modes)


kaligrafy:

greenscientist:
@samuel-duhaime , the main issue is the color of the map. So that will need to way until we have the deck.gl refactoring

PR #1810, opened 2026-02-23
Feature: Add dark mode support and preferences

samuel-duhaime:
# Pull request
fix: #29 

## Todos

- [x] Keep the preferences even when reloading the page
- [x] Modify the header correctly (font, etc.)
- [x] Fix all the logo to put them in black and not white. Inside packages/transition-frontend/src/components/dashboard/TransitionMenuBar.tsx
- [x] Fix that the dark mode and light mode don't change everything automatically.
- [x] Fix the back button for the dark mode settings
- [x] Show a little demo
- [x] Refactor to use() and ThemeContext
- [x] Make sure it incorporate the user browser preference if not set
- [x] Coderabbit and reviewers' comments.
- [x] Save preference, back, undo, redo

## Description
See commits messages.

## Summary by CodeRabbit

* **New Features**
  * Dark mode with system-preference detection and user override; toggle added to Preferences (EN/FR)
  * Theme-aware icons with distinct light/dark variants

* **Style**
  * Updated color tokens and spacing scale to support light/dark appearances site-wide
  * Root-level color-scheme switching via data attribute

* **Refactor**
  * Theming provider integrated into the app for consistent theme handling

* **Tests**
  * Added tests covering theme, preference toggle and icon selection

### DIFF
--- a/packages/chaire-lib-frontend/src/contexts/__tests__/ThemeContext.test.tsx
+++ b/packages/chaire-lib-frontend/src/contexts/__tests__/ThemeContext.test.tsx
@@ -89,9 +89,7 @@ describe('ThemeContext', () => {
 
         test('Context updates when matchMedia change listener is invoked', () => {
             systemPrefersDark = false;
-            (Preferences.get as jest.Mock).mockImplementation((path: string) =>
-                path === 'isDarkTheme' ? undefined : undefined
-            );
+            (Preferences.get as jest.Mock).mockImplementation(() => undefined);
             Object.defineProperty(window, 'matchMedia', {
                 value: jest.fn().mockImplementation(() => ({
                     get matches() {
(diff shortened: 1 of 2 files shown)


samuel-duhaime:
This is a video of the first try. We still need to change some colors and every logo (which are always white).

https://github.com/user-attachments/assets/4486b7c5-6774-4030-bedf-bb659a2f7f37

packages/chaire-lib-common/src/config/Preferences.ts
```diff
@@ -159,6 +159,21 @@ export class PreferencesClass extends ObjectWithHistory<PreferencesModelWithIdAn
         return this._isValid;
     }
 
+    /**
+     * Apply isDarkMode preference to the document's color-scheme (for CSS light-dark()).
+     * No-op when not in a browser. Call after load() or whenever isDarkMode is set.
+     */
+    public applyThemeToDocument(): void {
```
tahini: Je suis ambivalente à mettre ça dans Preferences, qui est dans chaire-lib-common. Ceci devrait être une fonction frontend il me semble. Peut-être que c'est ça qui devrait être intégré au redux state?

samuel-duhaime:
@tahini @kaligrafy Okay, I was able to refactor to 'use' context. The application now considers the browser theme preference if no themes are set yet.

I couldn't completely remove the import from 'Preferences' because all the application logic resides there, but perhaps at some point, we should create a user preferences API to better handle that, so we can get rid of chaire-lib-common.

samuel-duhaime:
@tahini Okay, I updated the translation setup, and I verified that every callback was the same as before (it was).

I will make sure to squash the commits into one message and to simplify the message.

PR #1812, opened 2026-02-23
networkDesign: add sample line count and weight to debug

tahini:
In order to determine for large sample if the current sampling approach (where each line has x% chance of being in the sample, instead of exactly x% of the file) can cause result biases, we add to the simulation results the number of lines effectively in the sample file and the total weight that it represents. From that information, we can decide how urgent it is to change the sampling algorithm (see the FIXME in `OdTripSimulation#sampleOdTripFile` function).


packages/transition-backend/src/services/simulation/methods/OdTripSimulation.ts
```diff
@@ -62,6 +62,8 @@ export class OdTripFitnessVisitor implements BatchRouteResultVisitor<SimulationS
     private totalWaitingTimeMinutes = 0;
     private totalTravelTimeMinutes = 0;
     private countByNumberOfTransfers = new Array(6).fill(0); // max 4, last is 5+
+    private sampleFileLinesCount = 0;
+    private sampleTotalWeight = 0;
```
greenscientist: This could have a comment, took me a minute to figure out what was supposed to represent sampleTOtalWeight

Issue #1780, opened 2026-02-12
import dissemination block does not work with french version of the gml file


tahini:
From https://www12.statcan.gc.ca/census-recensement/2021/geo/sip-pis/boundary-limites/index2021-eng.cfm?Year=21, I chose the gml file for the dissemination blocks in french and I got an error that it did not find the item `children`. And it silently failed to load anything, the task completed correctly, telling it added the csv data, but in fact, there was no zone in the database.

So 2 things in this issue:

* The french version of the dissemination block does not work (I switched to english and it imported correctly)
* The task does not exit with error if the zones file does not import properly.

PR #1813, opened 2026-02-24
Import dissemination block task now works with french file

GabrielBruno24:
Due to one of the tags being translated, the import would fail with the french version of the file. This also changes error handling so that the task will fail properly is an error happens during the import.
This increases the task length by about 10 seconds, for a total ~300 seconds.
 Fix: #1780

## Summary by CodeRabbit

* **New Features**
  * Zone imports now recognize multiple tag variants for zone identifiers, with improved extraction of the selected variant's value.

* **Bug Fixes**
  * Import now fails with a clear error if a zone identifier is missing.
  * Improved error logging and propagation during parsing and stream handling; errors are reported and rejected with details.


tahini (approved):
J'imagine que tu as bien testé l'import complet dans les 2 langues et que tout fonctionne?

PR #1830, opened 2026-03-02
Implement batch DB write in TrRoutingBatch

greenscientist:
Instead of writing to the DB for each OD pairs, we accumulate the results in an array, and commit them to the database once before we do the checkpoint. This should reduce the latency in the batch job by removing many database round-trip.

We add a callback to the JobCheckpointTracker object, to be called before we commit a checkpoint. This is used in TrRoutingBatch to flush the accumulated data to database.

With this, at the moment, we accumulate 250 results entry in the array. If we realize that memory usage is too high, we could call the write to database more often.

## Summary by CodeRabbit

* **New Features**
  * Added support for optional pre-checkpoint callbacks in route processing for enhanced extensibility.

* **Performance Improvements**
  * Implemented bulk insertion of route calculation results, replacing individual writes with batched operations.
  * Added result buffering mechanism during checkpoint operations to optimize data persistence.

### DIFF
--- a/packages/transition-backend/src/services/executableJob/JobCheckpointTracker.ts
+++ b/packages/transition-backend/src/services/executableJob/JobCheckpointTracker.ts
@@ -27,11 +27,14 @@ export class CheckpointTracker {
      * chunk size * nb of terminated chunks
      * @param currentCheckpoint If resuming a task, this is the last checkpoint
      * that was registered in previous run
+     * @param beforeCheckpoint, an optional callback to call before finalizing
+     * the checkpoint
      */
     constructor(
         private chunkSize: number,
         private progressEmitter: EventEmitter,
-        currentCheckpoint = 0
+        currentCheckpoint = 0,
+        private beforeCheckpoint?: (checkpoint: number) => Promise<void>
     ) {
         this.lastCheckpointIdx = Math.floor(currentCheckpoint / chunkSize) - 1;
         // Add items to the last checkpoint, in case the chunk size is not a multiple of the current checkpoint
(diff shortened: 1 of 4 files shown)


packages/transition-backend/src/services/transitRouting/TrRoutingBatch.ts
```diff
@@ -309,8 +323,9 @@ class TrRoutingBatch {
                     }
                 });
             }
-            // FIXME Do not synchronously wait for the save (~10% time overhead). When we have checkpoint support, we can do .then/catch to handle completion instead
-            await resultsDbQueries.create({
+
+            // No await — push to buffer, let the checkpoint listener flush in bulk
+            this.resultBuffer.push({
```
tahini: Should we flush after a certain length of buffer? Say one calculation takes abnormally a long time and 3 or 4 checkpoint buckets are filled before the next in line checkpoint is completed, that will be a lot of data to send to the DB at the same time. Just how large can a batch get before being too large?

tahini (commented):
Some considerations. I would approve for the evolutionary algorithm branch, but for main, please answer or consider the comments.

packages/transition-backend/src/services/executableJob/JobCheckpointTracker.ts
```diff
@@ -74,12 +77,24 @@ export class CheckpointTracker {
...
     }
 
+    private emitCheckpoint(checkpoint: number): void {
+        const doEmit = async () => {
+            // If defined, call the beforeCheckpoint callback to do finalization before confirming checkpoint
+            if (this.beforeCheckpoint !== undefined) {
+                await this.beforeCheckpoint(checkpoint);
+            }
+            console.log('Emitting checkpoint at index ', checkpoint);
+            this.progressEmitter.emit('checkpoint', checkpoint);
```
tahini: Also, could there be a race condition for checkpoint save, where a larger batch takes longer to save the results to the DB, while a newer batch with higher checkpoint completes and updates the checkpoint, the lower checkpoint will finish later and overwrite the higher checkpoint. I don't remember the job's checkpoint handling, if it validates that the new checkpoint is greater than the registered one.

PR #1836, opened 2026-03-05
Use a global TrRouting for a whole Generation

greenscientist:
In some benchmarks, it took 30 seconds to start TrRouting and 45s to run the calculation When running 20 candidates for a generation, that's 10 minutes lost. Since we use the same capnp file for the whole generation, we can share the same trRouting. Each candidate points to a different scenario.

We use TrRoutingBatchManager to start trRouting in the Generation. We create a FakeTrRoutingBatchManager class to pass to the TrRoutingBatchExecutor, which will contains the number of threads and port information to be used as if each run would have started their own TrRouting.

### DIFF
--- a/packages/transition-backend/src/services/transitRouting/FakeTrRoutingBatchManager.ts
+++ b/packages/transition-backend/src/services/transitRouting/FakeTrRoutingBatchManager.ts
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2026, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { EventEmitter } from 'events';
+
+import { TrRoutingBatchJobParameters } from './TrRoutingBatchJobParameters';
+
+import { TrRoutingBatchManager, TrRoutingBatchStartResult } from './TrRoutingBatchManager';
+
+/**
+ * Fake the lifecycle of the TrRoutingBatchManager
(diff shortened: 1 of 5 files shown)


packages/transition-backend/src/services/transitRouting/FakeTrRoutingBatchManager.ts
```diff
@@ -0,0 +1,48 @@
...
+import { EventEmitter } from 'events';
+
+import { TrRoutingBatchJobParameters } from './TrRoutingBatchJobParameters';
+
+import { TrRoutingBatchManager, TrRoutingBatchStartResult } from './TrRoutingBatchManager';
+
+/**
+ * Fake the lifecycle of the TrRoutingBatchManager
+ * Aim to just pass information from an actual run of TrRoutingBatchManager
+ * and no nothing.
```
kaligrafy: no nothing?

packages/transition-backend/src/services/transitRouting/FakeTrRoutingBatchManager.ts
```diff
@@ -0,0 +1,48 @@
...
+     */
+    async startBatch(workloadSize: number, options?: TrRoutingBatchJobParameters): Promise<TrRoutingBatchStartResult> {
+        return {
+            threadCount: this.threadCount,
+            port: this.port
+        };
+    }
+
+    /**
+     * Noop
```
kaligrafy: Please explain why this is needed

kaligrafy (approved):

PR #1883, opened 2026-04-04
Fix nested horizontal scroll in schedule period form

iTzxMigz:
- Remove nested horizontal scroll by dropping _flex-container-row class from the period (kept on the parent list only), using inline flex styles instead.
- Constrain period width with maxWidth on the form section and selects so long path names no longer stretch each column.
- Allow clearing outbound/inbound path when only one choice is available (distinguish 'never set' vs 'user cleared').
- Hide the Generate button when no outbound path is selected, since the backend rejects schedule generation without one.
- Add optional style prop to InputSelect for per-usage width constraints.

We can see a preview here the main goal is to remove a scroll in a scroll horizontally and make generate trip without an outbound selection unavailable since it is not allowed: 
Screenshot 2026-04-04 at 12 43 41 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented schedule-generation button from appearing when outbound path is blank, avoiding invalid submissions.

* **Improvements**
  * Select inputs now accept custom inline styling for greater layout flexibility.
  * Schedule period form layout refined with explicit flex row sizing, fit-content width and tighter inner width constraints for improved alignment and visual consistency.

### DIFF
--- a/packages/chaire-lib-frontend/src/components/input/InputSelect.tsx
+++ b/packages/chaire-lib-frontend/src/components/input/InputSelect.tsx
@@ -24,6 +24,7 @@ export type InputSelectProps = {
     localePrefix?: string;
     t?: (string) => string;
     disabled?: boolean;
+    style?: React.CSSProperties;
 };
 
 type defaultInputType = {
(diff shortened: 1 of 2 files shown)


tahini (commented):
@kaligrafy, pourrais-tu reviewer, pour les styles, c'est plus dans ton domaine.

PR #1886, opened 2026-04-08
fix(frontend): avoid routing UI crash and flake while scenarios load

kaligrafy:
Guard scenarioCollection in TransitRoutingForm and ScenarioComparisonPanel effects so getById is not called before collections exist. Fix accessibility map comparison forms to validate against the freshly fetched collection after setState instead of stale state.

UI tests: allow longer wait only for the routing left-menu case so the assertion can pass after LoadingPage while scenarios load (expect retries until the panel heading appears; no fixed sleep).

## Summary by CodeRabbit

* **Bug Fixes**
  * Forms and panels now safely handle temporarily missing data collections, avoiding errors during load/resync and preserving existing behavior for clearing invalid scenario selections and population flags.

* **Tests**
  * Test suite stability improved by increasing Playwright timeouts so routing-related pages and assertions have more time to load and settle.

### DIFF
--- a/packages/transition-frontend/ui-tests/testHelpers.ts
+++ b/packages/transition-frontend/ui-tests/testHelpers.ts
@@ -199,6 +199,9 @@ export const clickLeftMenuTest: LeftMenuTest = ({ context, section }) => {
             expectedRightPanelTitle = 'Preferences';
             break;
         }
+        // The routing section keeps a full-panel LoadingPage until the scenario collection is ready, so the first
+        // real heading appears later than for agencies/preferences. Playwright retries this assert until match or
+        // expect.timeout (aligned with test timeout in playwright.config—not a fixed sleep).
         await expect(rightPanelTitle).toContainText(expectedRightPanelTitle);
     });
 };
--- a/packages/transition-frontend/playwright-example.config.ts
+++ b/packages/transition-frontend/playwright-example.config.ts
@@ -12,8 +12,11 @@ export default defineConfig({
   forbidOnly: !!process.env.CI,
   /* Retry on CI only */
   retries: process.env.CI ? 2 : 0,
-  // Each test is given 15 seconds.
-  timeout: 15000,
+  // Align test and expect timeouts (default expect is 5s); routing waits on scenario collection before headings appear.
+  timeout: 30000,
+  expect: {
+    timeout: 30000
+  },
   /* Opt out of parallel tests on CI. */
   workers: process.env.CI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
--- a/packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
+++ b/packages/transition-frontend/src/components/forms/accessibilityComparison/AccessibilityComparisonForm.tsx
@@ -170,21 +170,25 @@ class AccessibilityComparisonForm extends ChangeEventsForm<
     }
 
     onScenarioCollectionUpdate() {
-        this.setState({ scenarioCollection: serviceLocator.collectionManager.get('scenarios') });
+        const scenarioCollection = serviceLocator.collectionManager.get('scenarios');
+        this.setState({ scenarioCollection });
 
         // If a previously selected scenario was deleted, the current scenario ID will remain but the scenario itself will no longer exist, leading to an error.
         // In that case, change it to undefined.
+        if (!scenarioCollection) {
+            return;
+        }
         const routing = this.state.object;
         const alternateRouting = this.state.alternateScenarioRouting;
         const scenarioId1 = routing.attributes.scenarioId;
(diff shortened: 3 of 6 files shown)


packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -199,6 +205,12 @@ export const clickLeftMenuTest: LeftMenuTest = ({ context, section }) => {
             expectedRightPanelTitle = 'Preferences';
             break;
         }
-        await expect(rightPanelTitle).toContainText(expectedRightPanelTitle);
+        if (section === 'routing') {
+            await expect(rightPanelTitle).toContainText(expectedRightPanelTitle, {
+                timeout: routingSectionReadyTimeoutMs
```
tahini: Do you need the timeout here as well as on  line 186 above at the test level?
Also doesn't the agency section also depend on data to be loaded? in which case the timeout should also apply.

tahini (commented):
Just a question

tahini (approved):
Toujours la question, mais sinon c'est bon

packages/transition-frontend/ui-tests/testHelpers.ts
samuel-duhaime: Please, put your comment on the reason you did that in this place. (not in routingSectionReadyTimeoutMs), so we can remove it if the reason is not there in the future.

samuel-duhaime (changes requested):
So I did read the documentation of Playwright. I learned a few things.

Our application config has a global timeout of 15 seconds (we override the default right now) for test timeout, but we still have the default timeout for expect of 5 seconds. We might consider increasing it a bit. ({ expect: { timeout: 15000 } })

Image

packages/transition-frontend/ui-tests/testHelpers.ts
kaligrafy: I dropped the routing-only expect timeout and test.setTimeout. The global timeout and expect.timeout in playwright.config are aligned (30s), so every section—including agencies—gets the same retry budget for this assert, without duplicating logic. Routing was the slow path because the panel stays on LoadingPage until scenarios load; that’s documented above the expect in testHelpers.ts.

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -180,6 +183,9 @@ export const logoutTest: LogoutTest = ({ context }) => {
  */
 export const clickLeftMenuTest: LeftMenuTest = ({ context, section }) => {
     test(`Click the ${section} section of the left menu - ${getTestCounter(context, `${section}`)}`, async () => {
+        if (section === 'routing') {
+            test.setTimeout(routingSectionReadyTimeoutMs);
+        }
```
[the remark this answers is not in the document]
kaligrafy: Dropped test.setTimeout: the issue was the global test timeout being shorter than the assertion wait we needed. We aligned timeout and expect.timeout in playwright.config to 30s so the routing panel can finish loading without a special case. Documented why routing is slower in a comment on the shared expect in clickLeftMenuTest.

samuel-duhaime (approved):
Okay!

samuel-duhaime (approved):

packages/transition-frontend/ui-tests/testHelpers.ts
```diff
@@ -199,6 +199,9 @@ export const clickLeftMenuTest: LeftMenuTest = ({ context, section }) => {
             expectedRightPanelTitle = 'Preferences';
             break;
         }
+        // The routing section keeps a full-panel LoadingPage until the scenario collection is ready, so the first
```
tahini, line 202: En même temps, dans un monde idéal, toutes les sections devraient agir de la même façon. Si on met les headings et qu'on gère le load après le heading dans Agencies, il faudrait faire de même dans toutes les sections. C,est une question de cohérence.

On a donc 2 problèmes ici:
1- Celui que tu fixes et qui montrait une erreur si la collection n'était loadée (et on ne l'aurait pas nécessairemetn vu si le 2e problème n'avait pas existé, en fait vu oui, mais pas nécessairement réglé maintenant :P)
2- Le manque de cohérence entre nos sections

kaligrafy: Ok, donc j'essaie de faire la même chose partout ou on créé un issue séparé?

tahini (approved):

packages/transition-frontend/ui-tests/testHelpers.ts
tahini: > Ok, donc j'essaie de faire la même chose partout ou on créé un issue séparé?

Idéalement le fixer maintenant, sinon, l'issue, ça ira aux calendes grecques

kaligrafy: That’s closer to a small epic (ticket + incremental PRs) than a “while we’re here” change.

Issue #238, opened 2021-06-17
transition: add a way to batch delete in form


kaligrafy:
like adding a checkbox for each service/scenario/lines, etc. and delete them in batch

MaximeCoteL:
Having a way to batch delete lines would be a blessing. I imported a bunch of lines and turns out they were the wrong ones... Also, Having a button that deletes Lines with no paths would be great too.

PR #1888, opened 2026-04-09
scenarios: allow to select multiple scenarios and delete them

tahini:
part of #238

Add the `deleteByIds` function to `ScenarioCollection`, which receives
an array of ids and deletes each unfrozen scenario individually.

Add a checkbox to the `TransitScenarioButton` component, with a
callback when the selection status changes. When scenarios are selected,
a button at the bottom of the list allows to delete the selected
scenarios. A confirm modal requests confirmation to delete the scenarios
before executing the command.

## Summary by CodeRabbit

* **New Features**
  * Multi-select scenarios with checkboxes and a bulk "Delete selected scenarios" action including confirmation and progress reporting.

* **Localization**
  * Added English and French UI strings for bulk deletion labels and confirmation prompts (with count interpolation).

* **Backend & API**
  * Added support for deleting multiple scenarios in one operation and emitting update/ cache notifications after deletions.

* **Tests**
  * Added unit tests covering bulk-delete success, partial success, and error paths.

### DIFF
--- a/locales/en/transit.json
+++ b/locales/en/transit.json
@@ -110,7 +110,10 @@
         "New": "New scenario",
         "Edit": "Scenario",
         "Delete": "Delete scenario",
+        "DeleteSelected": "Delete selected scenarios",
         "ConfirmDelete": "Do you confirm deletion of this scenario?",
+        "ConfirmDeleteSelected": "Do you confirm deletion of {{count}} selected scenarios? This operation cannot be undone.",
+        "ConfirmDeleteSelected_one": "Do you confirm deletion of the selected scenario? This operation cannot be undone.",
         "DuplicateScenario": "Duplicate scenario",
         "ImportFromJson": "Import scenarios (JSON)",
         "errors": {
@@ -145,6 +148,7 @@
         "ConfirmDeleteWithSchedule": "Do you confirm deletion of this service? Schedules associated with this service will also be deleted. This operation cannot be undone.",
         "ConfirmDeleteUnused": "Do you confirm deletion of all unused services? The unused services are those with no schedules associated with it.",
         "ConfirmDeleteSelected": "Do you confirm deletion of {{count}} selected services? Schedules associated with these services will also be deleted. This operation cannot be undone.",
+        "ConfirmDeleteSelected_one": "Do you confirm deletion of the selected service? Schedules associated with this service will also be deleted. This operation cannot be undone.",
         "ImportFromJson": "Import services (JSON)",
         "DuplicateService": "Duplicate service",
         "oneLine": "Line (1)",
--- a/locales/fr/transit.json
+++ b/locales/fr/transit.json
@@ -110,7 +110,10 @@
         "Color": "Couleur",
         "Edit": "Scénario",
         "Delete": "Supprimer le scénario",
+        "DeleteSelected": "Supprimer les scénarios sélectionnés",
         "ConfirmDelete": "Confirmez-vous que vous désirez bien supprimer ce scénario?",
+        "ConfirmDeleteSelected": "Confirmez-vous que vous désirez bien supprimer les {{count}} scénarios sélectionnés? Cette opération est irréversible.",
+        "ConfirmDeleteSelected_one": "Confirmez-vous que vous désirez bien supprimer le scénario sélectionné? Cette opération est irréversible.",
         "DuplicateScenario": "Dupliquer le scénario",
         "ImportFromJson": "Importer des scénarios (JSON)",
         "errors": {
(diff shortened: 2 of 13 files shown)


packages/transition-common/src/services/scenario/ScenarioCollection.ts
```diff
@@ -54,6 +55,25 @@ class ScenarioCollection extends GenericObjectCollection<Scenario> implements Pr
         });
     }
 
+    async deleteByIds(ids: string[], socket) {
+        // Delete scenarios individually
+        // FIXME Support a batch delete job in the backend to avoid multiple long calls, but this is similar to the services delete call
```
greenscientist: I'll just say, why not do it properly now? :)

tahini: > I'll just say, why not do it properly now? :)

Why? everlasting priority vs time to implement and test. But I can put this in the evolutionaryAlgo experimental branch for now and do it properly later, along with the selectAll option...

greenscientist: Feels like doing it through the evolalgo branch is even more techdept :/

tahini: Done, I'm working on a PR that will prevent deleting frozen elements in the `deleteMultiple` call. Next, will be the backend delete of services

packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
```diff
@@ -106,7 +107,8 @@ const transitClassesConfig = {
         dbQueries: scenariosDbQueries,
         cacheQueries: scenariosCacheQueries,
         collection: new ScenarioCollection([], {}),
-        saveCollectionToCacheFct: dbToCacheQueries.loadAndSaveScenariosToCache
+        saveCollectionToCacheFct: dbToCacheQueries.loadAndSaveScenariosToCache,
+        deleteMultiple: scenariosDbQueries.deleteMultiple
```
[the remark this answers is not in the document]
tahini: Non-deletion of the frozen scenarios will be done in a separate PR, preventing deletion at the DB query level instead of relying on consumers having the check the individual status of scenarios.

packages/transition-common/src/services/scenario/ScenarioCollection.ts
```diff
@@ -54,6 +55,18 @@ class ScenarioCollection extends GenericObjectCollection<Scenario> implements Pr
         });
     }
 
+    async deleteByIds(ids: string[], socket): Promise<string[]> {
```
greenscientist, line 58: Can we add a comment that says that this function is front-end only or should be moved to the -frontend package? (So that we know were to use it, until we move the code in the right places)

packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
```diff
@@ -45,6 +45,7 @@ export interface TransitObjectDataHandler {
     read: (id: string, customCachePath: string | undefined) => Promise<Record<string, any>>;
     update: (socket: EventEmitter, id: string, attributes: GenericAttributes) => Promise<Record<string, any>>;
     delete: (socket: EventEmitter, id: string, customCachePath: string | undefined) => Promise<Record<string, any>>;
+    deleteMultiple?: (socket: EventEmitter, ids: string[]) => Promise<Status.Status<{ deletedIds: string[] }>>;
```
greenscientist, line 49: Can we have a TODO to implement it for all type. It's weird to have it optional where all the other basic operations are required. 
Tk, up to you...

packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
```diff
@@ -264,6 +266,28 @@ function createDataHandlers(): Record<string, TransitObjectDataHandler> {
...
+        // as we don't necessarily want to expose them all by default.
+        if (transitClassConfig.deleteMultiple) {
+            dataHandler.deleteMultiple = async (socket: EventEmitter, ids: string[]) => {
+                try {
+                    const deletedIds = await transitClassConfig.deleteMultiple(ids);
+                    if (deletedIds.length > 0 && isSocketIo(socket)) {
+                        // Objects were deleted, notify clients
+                        socket.broadcast.emit('data.updated');
+                        socket.emit('cache.dirty');
+                    }
```
greenscientist, line 308: Like the delete:, we should call the functions to call deleteCache if they are available.

greenscientist (changes requested):
The blocking change is managing the cache side. 
The delete function also call cache delete, we should do it in the multiple case also.

The rest is me wanting more comment TODO

@kaligrafy or @samuel-duhaime should check the commit about the translation part.

locales/en/transit.json
```diff
@@ -110,7 +110,9 @@
         "New": "New scenario",
         "Edit": "Scenario",
         "Delete": "Delete scenario",
+        "DeleteSelected": "Delete selected scenarios",
         "ConfirmDelete": "Do you confirm deletion of this scenario?",
+        "ConfirmDeleteSelected": "Do you confirm deletion of {{count}} selected scenarios? This operation cannot be undone.",
```
samuel-duhaime, line 115: This is only good for 2 or more count.
Can you have a '..._one' also?

locales/fr/transit.json
```diff
@@ -110,7 +110,9 @@
         "Color": "Couleur",
         "Edit": "Scénario",
         "Delete": "Supprimer le scénario",
+        "DeleteSelected": "Supprimer les scénarios sélectionnés",
         "ConfirmDelete": "Confirmez-vous que vous désirez bien supprimer ce scénario?",
+        "ConfirmDeleteSelected": "Confirmez-vous que vous désirez bien supprimer les {{count}} scénarios sélectionnés? Cette opération est irréversible.",
```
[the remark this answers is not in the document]
samuel-duhaime: Yeah same problem.

samuel-duhaime (commented):
Just commenting on the translations.

packages/transition-backend/src/services/transitObjects/TransitObjectsDataHandler.ts
tahini: Currently it throws an error if the object has a deleteCache function (that would be line), so when we want to implement deleteMultiple for this class, we'll have to fix.

greenscientist (approved):
I'm good with the additionnal comment and catch/throw protection for multiple delete with cache.

PR #1899, opened 2026-04-15
Refactor generateGeographyAndSegments functions to use named params

thomasspina:
Converts generateGeographyAndSegmentsFromGtfs and generateGeographyAndSegmentsFromStopTimes in PathGtfsGeographyGenerator.ts from positional parameters to a single named params object, consistent with other functions in the same file. This makes it easier to add new optional parameters without positional ambiguity or undefined holes.

This PR was made to address the growth of these functions in PR #1892 and related comments to their growth.

## Summary by CodeRabbit

* **Refactor**
  * Updated path import service functions to use clearer code patterns, improving code maintainability and readability.

* **Tests**
  * Updated corresponding tests to align with refactored service functions.

### DIFF
--- a/packages/transition-backend/src/services/path/PathGtfsGeographyGenerator.ts
+++ b/packages/transition-backend/src/services/path/PathGtfsGeographyGenerator.ts
@@ -411,16 +411,26 @@ const calculateDistances = (
     return result.status === 'success' ? result : calculateDistancesBySegments(params);
 };
 
-export const generateGeographyAndSegmentsFromGtfs = (
-    path: Path,
-    shapeCoordinatesWithDistances: GtfsTypes.Shapes[],
-    nodeIds: string[],
-    allTripsStopTimes: StopTime[][],
-    shapeGtfsId: string,
-    stopCoordinatesByStopId: { [key: string]: [number, number] },
-    defaultLayoverRatioOverTotalTravelTime = 0.1,
-    defaultMinLayoverTimeSeconds = 180
-): TranslatableMessage[] => {
+export const generateGeographyAndSegmentsFromGtfs = (params: {
+    path: Path;
+    shapeCoordinatesWithDistances: GtfsTypes.Shapes[];
+    nodeIds: string[];
+    allTripsStopTimes: StopTime[][];
+    shapeGtfsId: string;
+    stopCoordinatesByStopId: { [key: string]: [number, number] };
+    defaultLayoverRatioOverTotalTravelTime?: number;
+    defaultMinLayoverTimeSeconds?: number;
+}): TranslatableMessage[] => {
+    const {
+        path,
+        shapeCoordinatesWithDistances,
+        nodeIds,
+        allTripsStopTimes,
+        shapeGtfsId,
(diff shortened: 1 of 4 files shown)


packages/transition-backend/src/services/path/PathGtfsGeographyGenerator.ts
```diff
@@ -411,16 +411,26 @@ const calculateDistances = (
...
+export const generateGeographyAndSegmentsFromGtfs = (params: {
+    path: Path;
+    shapeCoordinatesWithDistances: GtfsTypes.Shapes[];
+    nodeIds: string[];
+    allTripsStopTimes: StopTime[][];
+    shapeGtfsId: string;
+    stopCoordinatesByStopId: { [key: string]: [number, number] };
+    defaultLayoverRatioOverTotalTravelTime?: number;
+    defaultMinLayoverTimeSeconds?: number;
+}): TranslatableMessage[] => {
```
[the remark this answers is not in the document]
thomasspina: this adds a layer of indirection for only two functions. Not valuable

tahini (approved):

packages/transition-backend/src/services/path/PathGtfsGeographyGenerator.ts
```diff
@@ -557,14 +567,22 @@ export const generateGeographyAndSegmentsFromGtfs = (
...
+    const {
+        path,
+        nodeIds,
+        allTripsStopTimes,
+        stopCoordinatesByStopId,
+        defaultLayoverRatioOverTotalTravelTime = 0.1,
+        defaultMinLayoverTimeSeconds = 180
+    } = params;
```
greenscientist, line 585: I don't see what's the point of doing an object if you just extract all the content right away. 

@tahini Is there a reason to do this, this seem pointless ?

greenscientist (dismissed):

packages/transition-backend/src/services/path/PathGtfsGeographyGenerator.ts
thomasspina: One of the main concerns that she stated in a comment on PR #1892 was, to my understanding, because calling these functions and adding "undefined" for some parameters in the function call made undefined holes in the call. See https://github.com/chairemobilite/transition/pull/1892#discussion_r3073349681

tahini: The point of doing an object, is for clarity for callers, we absolutely WANT this when we have many parameters.

Now destructuring in the signature above (instead of `params`) vs in the function is purely syntactic and has no impact at runtime or on the usage of the function. We do not have explicit preferences in the project for either, though our most recent code usually use the destructuring in signature. But I didn't feel like objecting on this.

greenscientist (commented):
So apparently @tahini prefer this that way, so I'll remove my request change. 

Don't worry @thomasspina this is a half decade long debate between me and tahini!

tahini (commented):
Ok, I was certain to have posted this comment, seems it was still pending. Anyway...

PR #1904, opened 2026-04-15
Fix division by zero and trip time rounding

thomasspina:
Guard against division-by-zero in PathGeographyGenerator.buildPathData speed calculations, which previously produced Infinity when total travel time was zero. The three speed fields now check for a positive denominator and return 0 instead. Also refactor calculateSegmentDuration to detect zero or negative OSRM durations early before they cause a divide-by-zero, and change the existing duration check from <= 0 to also include an isNaN so it also catches NaN from the physics model.

Round trip arrival and departure times to whole seconds in Schedule.generateTrip. Segment travel times can be fractional, and the accumulated tripTimeSoFar drifts away from whole seconds over many segments. Also guard against period.trips being undefined in getAssociatedPathIds, which crashes when a schedule period exists but has no generated trips.

These were pointed out by CodeRabbit as part of PR #1892.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for path duration calculations to catch zero/non-finite routing results and stop before producing invalid speeds/timings.
  * Ensured schedule period data consistently includes an empty `trips` array when periods are created as defaults.
* **Tests**
  * Added unit tests for routing/physics duration edge cases and verifying the correct error handling.
* **Documentation**
  * Clarified that trip and node time fields are expected as integer seconds; trip timing generation now rounds values.

### DIFF
--- a/packages/transition-backend/src/models/capnpCache/transitLines.cache.queries.ts
+++ b/packages/transition-backend/src/models/capnpCache/transitLines.cache.queries.ts
@@ -197,14 +197,12 @@ const exportParser = function (object: Line, cacheObject: CacheObjectClass) {
             periodCache.setNumberOfUnits(period.number_of_units || -1);
             periodCache.setIsFrozen(boolToInt8(period.is_frozen));
 
-            if (_isBlank(period.trips)) {
-                period.trips = [];
-            }
+            const trips = period.trips ?? [];
 
-            const tripsCache = periodCache.initTrips(period.trips.length);
+            const tripsCache = periodCache.initTrips(trips.length);
 
-            for (let k = 0, countK = period.trips.length; k < countK; k++) {
-                const trip = period.trips[k];
+            for (let k = 0, countK = trips.length; k < countK; k++) {
+                const trip = trips[k];
                 const tripCache = tripsCache.get(k);
                 tripCache.setUuid(trip.id);
                 tripCache.setPathUuid(trip.path_id);
--- a/packages/transition-backend/src/models/db/transitSchedules.db.queries.ts
+++ b/packages/transition-backend/src/models/db/transitSchedules.db.queries.ts
@@ -180,7 +180,10 @@ const _createPeriods = async function (
         period.integer_id = ids[idx].id;
     });
     const tripPromises = schedulePeriods
-        .filter((period) => period.trips !== undefined)
+        .filter(
+            (period): period is SchedulePeriod & { trips: Exclude<SchedulePeriod['trips'], undefined> } =>
+                period.trips !== undefined
+        )
         .map((period) => {
             period.trips.forEach((trip) => {
                 if (!trip.id) {
--- a/packages/transition-backend/src/services/gtfsImport/ScheduleImporter.ts
+++ b/packages/transition-backend/src/services/gtfsImport/ScheduleImporter.ts
@@ -359,7 +359,11 @@ const generateExactTrips = (schedule: Schedule, periods: TripAndStopTimes[][], i
                 // TODO: We ignore the block_id during import, so we set it to null here. When we do something with the block,
             };
 
-            (schedule as Schedule).attributes.periods[periodIndex].trips.push(newTrip);
+            const period = (schedule as Schedule).attributes.periods[periodIndex];
+            if (!period.trips) {
+                period.trips = [];
+            }
+            period.trips.push(newTrip);
         });
     }
 };
--- a/packages/transition-common/src/services/path/PathGeographyGenerator.ts
+++ b/packages/transition-common/src/services/path/PathGeographyGenerator.ts
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
     const routingEngine = path.getData('routingEngine') as string;
     const defaultRunningSpeedKmH = path.getData('defaultRunningSpeedKmH') as number;
 
-    const runningSpeedMps =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? segmentDistanceMeters / routedDurationSeconds
-            : kphToMps(defaultRunningSpeedKmH);
+    // When the routing engine is 'engine' (OSRM), segment duration comes from the router.
+    // Otherwise ('engineCustom'/'manual'), duration is derived from defaultRunningSpeedKmH.
+    const usesRoutedDuration = routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH);
 
-    const noDwellTimeDurationSeconds =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? routedDurationSeconds
-            : segmentDistanceMeters / runningSpeedMps;
+    // Catch zero/negative OSRM durations early to avoid dividing by zero below
+    if (usesRoutedDuration && routedDurationSeconds <= 0) {
+        throw new TrError(
+            'Error trying to generate a path geography. OSRM returned zero or negative duration for segment.',
+            'PUPDGEO0002',
+            'TransitPathCannotUpdateGeographyBecauseErrorCalculatingSegmentDuration'
+        );
+    }
+
+    const runningSpeedMps = usesRoutedDuration
+        ? segmentDistanceMeters / routedDurationSeconds
+        : kphToMps(defaultRunningSpeedKmH);
+
+    const noDwellTimeDurationSeconds = usesRoutedDuration
+        ? routedDurationSeconds
+        : segmentDistanceMeters / runningSpeedMps;
 
     const calculatedSegmentDurationSeconds = durationFromAccelerationDecelerationDistanceAndRunningSpeed(
         acceleration,
@@ -156,7 +167,7 @@ const calculateSegmentDuration = (
         runningSpeedMps
     );
 
-    if (calculatedSegmentDurationSeconds <= 0) {
+    if (calculatedSegmentDurationSeconds <= 0 || isNaN(calculatedSegmentDurationSeconds)) {
         throw new TrError(
             'Error trying to generate a path geography. There was an error while calculating segment duration.',
             'PUPDGEO0001',
--- a/packages/transition-common/src/services/path/PathGeographyGenerator.ts
+++ b/packages/transition-common/src/services/path/PathGeographyGenerator.ts
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
     const routingEngine = path.getData('routingEngine') as string;
     const defaultRunningSpeedKmH = path.getData('defaultRunningSpeedKmH') as number;
 
-    const runningSpeedMps =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? segmentDistanceMeters / routedDurationSeconds
-            : kphToMps(defaultRunningSpeedKmH);
+    // When the routing engine is 'engine', segment duration comes from the router.
+    // Otherwise ('engineCustom'/'manual'), duration is derived from defaultRunningSpeedKmH.
+    const usesRoutedDuration = routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH);
 
-    const noDwellTimeDurationSeconds =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? routedDurationSeconds
-            : segmentDistanceMeters / runningSpeedMps;
+    // Coincident nodes/waypoints produce a zero-duration leg; catch it before dividing by zero below
+    if (usesRoutedDuration && routedDurationSeconds === 0) {
+        throw new TrError(
+            'Error trying to generate a path geography. OSRM returned a zero duration for segment.',
+            'PUPDGEO0004',
+            'TransitPathCannotUpdateGeographyBecauseErrorCalculatingSegmentDuration'
+        );
+    }
+
+    const runningSpeedMps = usesRoutedDuration
+        ? segmentDistanceMeters / routedDurationSeconds
+        : kphToMps(defaultRunningSpeedKmH);
+
+    const noDwellTimeDurationSeconds = usesRoutedDuration
+        ? routedDurationSeconds
+        : segmentDistanceMeters / runningSpeedMps;
 
     const calculatedSegmentDurationSeconds = durationFromAccelerationDecelerationDistanceAndRunningSpeed(
         acceleration,
@@ -156,7 +167,7 @@ const calculateSegmentDuration = (
         runningSpeedMps
     );
 
-    if (calculatedSegmentDurationSeconds <= 0) {
+    if (calculatedSegmentDurationSeconds <= 0 || !isFinite(calculatedSegmentDurationSeconds)) {
         throw new TrError(
             'Error trying to generate a path geography. There was an error while calculating segment duration.',
             'PUPDGEO0001',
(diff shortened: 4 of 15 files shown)


packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -156,7 +164,7 @@ const calculateSegmentDuration = (
         runningSpeedMps
     );
 
-    if (calculatedSegmentDurationSeconds <= 0) {
+    if (!(calculatedSegmentDurationSeconds > 0)) {
```
greenscientist: This has the same meaning, but with a negative which is harder to understand, please don't do that.

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -139,15 +139,23 @@ const calculateSegmentDuration = (
     const routingEngine = path.getData('routingEngine') as string;
     const defaultRunningSpeedKmH = path.getData('defaultRunningSpeedKmH') as number;
 
-    const runningSpeedMps =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? segmentDistanceMeters / routedDurationSeconds
-            : kphToMps(defaultRunningSpeedKmH);
+    const usesRoutedDuration = routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH);
```
greenscientist, line 144: Please add a comment that explain this

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -365,18 +373,21 @@ const buildPathData = (
...
-        ),
-        operatingSpeedMetersPerSecond: roundToDecimals(
-            totals.totalDistance / totals.totalTravelTimeWithDwellTimesSeconds,
-            2
-        ),
-        operatingSpeedWithLayoverMetersPerSecond: roundToDecimals(
-            totals.totalDistance / (totals.totalTravelTimeWithDwellTimesSeconds + layoverTimeSeconds),
-            2
-        ),
+        averageSpeedWithoutDwellTimesMetersPerSecond:
```
greenscientist: Why would these value be 0? Is that an expected value? Are all zero together or not?

Having lot of ? : makes the code harder to read, so more comment could be needed.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -223,7 +223,7 @@ export abstract class BaseScheduleStrategy implements ScheduleGenerationStrategy
                 const segment = segments[i];
                 const dwellTimeSeconds = dwellTimes[i] || 0;
                 if (i > 0) {
-                    tripArrivalTimesSeconds.push(tripTimeSoFar);
+                    tripArrivalTimesSeconds.push(Math.round(tripTimeSoFar));
```
greenscientist, line 230: Why is fractional second bad?  This would need a comment, since it's everyone and a programmer modifying the code might not know that it's required.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1065,7 +1065,7 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
         const periods = this.attributes.periods;
         for (let i = 0, countI = periods.length; i < countI; i++) {
             const period = periods[i];
-            period.trips.forEach((trip) => (associatedPathIds[trip.path_id] = true));
+            (period.trips || []).forEach((trip) => (associatedPathIds[trip.path_id] = true));
```
greenscientist: Is an empty trip a normal condition? Should we handle this at an higher level when we create the trips?  Of not normal we would need an error message.

greenscientist (changes requested):
See comments

packages/transition-common/src/services/schedules/Schedule.ts
thomasspina: The physics model doesn't round and fractions just get truncated instead of rounded when stored which can lead to inconsistencies of at most 1 second between what is in memory and what the db would give. I suppose that justification is a bit weak all things considered. I'll remove these changes.

packages/transition-common/src/services/path/PathGeographyGenerator.ts
thomasspina: Okay. You're right on this one. These guards suggested by CodeRabbit are probably excessive considering that the scenario where this happens is nearly impossible. Code readability is probably a priority and these defensive guards are impeding more than they are of service. I'll remove them and use better judgment in the future.

packages/transition-common/src/services/schedules/Schedule.ts
thomasspina: I think this should be addressed higher. There are examples with both a guard or none in the code already.

```const trips = periods[periodIdx].trips || [];``` in *ScheduleExporter.ts*
and
```tripsCount += periods[i].trips.length;``` in *Schedule.ts*

maybe we could do something like: 
```
export interface SchedulePeriod extends GenericAttributes {
      //...
      trips?: SchedulePeriodTrip[];
 }
```
in Schedule.ts to have it consistent throughout.

Comment (thomasspina) [reply packages/transition-common/src/services/schedules/Schedule.ts]: I made a new push with a new commit with this change for you to take a look at.

File: packages/transition-backend/src/models/db/transitSchedules.db.queries.ts [hunk 6]
```diff
@@ -180,7 +180,7 @@ const _createPeriods = async function (
         period.integer_id = ids[idx].id;
     });
     const tripPromises = schedulePeriods
-        .filter((period) => period.trips !== undefined)
+        .filter((period): period is SchedulePeriod & { trips: SchedulePeriodTrip[] } => period.trips !== undefined)
```
tahini: À ce moment-ci, ce type n'est pas vrai, puisque period.trips peut être undefined, c'est plutôt dans le .map qu'il faudrati ajouter ce type. D'ailleurs, typescript permet d'exclure la valeur `undefined` d'un type avec `Exclude<SchedulePeriod['trips'], undefined>`, ce qui permet de ne pas avoir explicitement le type de `trips`, mais juste inclure la possibilité que ce soit undefined.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1065,7 +1065,7 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
         const periods = this.attributes.periods;
         for (let i = 0, countI = periods.length; i < countI; i++) {
             const period = periods[i];
-            period.trips.forEach((trip) => (associatedPathIds[trip.path_id] = true));
+            (period.trips ?? []).forEach((trip) => (associatedPathIds[trip.path_id] = true));
```
tahini: Ici, puisque c'est un forEach, ce serait plus clair de faire un `if period.trips !== undefined` plutôt que de rouler la boucle avec un array vide.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1135,7 +1135,7 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
         const periods = this.attributes.periods;
         for (let i = 0, countI = periods.length; i < countI; i++) {
             if (periods[i].trips) {
-                tripsCount += periods[i].trips.length;
+                tripsCount += (periods[i].trips ?? []).length;
             }
```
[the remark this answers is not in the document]
tahini: That is true, we already know trips is defined here

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
...
+    // When the routing engine is 'engine' (OSRM), segment duration comes from the router.
+    // Otherwise ('engineCustom'/'manual'), duration is derived from defaultRunningSpeedKmH.
+    const usesRoutedDuration = routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH);
 
-    const noDwellTimeDurationSeconds =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? routedDurationSeconds
-            : segmentDistanceMeters / runningSpeedMps;
+    // Catch zero/negative OSRM durations early to avoid dividing by zero below
+    if (usesRoutedDuration && routedDurationSeconds <= 0) {
```
tahini: Est-ce qu'il n'y aurait pas des tests unitaires à modifier pour tester ce cas par hasard?

thomasspina: C'est bon j'ai ajouté un test pour ça

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
     const routingEngine = path.getData('routingEngine') as string;
     const defaultRunningSpeedKmH = path.getData('defaultRunningSpeedKmH') as number;
 
-    const runningSpeedMps =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? segmentDistanceMeters / routedDurationSeconds
-            : kphToMps(defaultRunningSpeedKmH);
+    // When the routing engine is 'engine' (OSRM), segment duration comes from the router.
```
tahini: remove (OSRM) from the comment. Maintenant osrm est le seul engin, mais on prévoit depuis des années en essayer d'autres, donc juste ne rien mettre entre parenthèses. router -> routing engine

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -70,7 +70,7 @@ export interface SchedulePeriod extends GenericAttributes {
     // FIXME: Use seconds since midnight format instead of string, which can be anything
     custom_start_at_str?: string;
     custom_end_at_str?: string;
-    trips: SchedulePeriodTrip[];
+    trips?: SchedulePeriodTrip[];
```
tahini: Dans un monde idéal, ça devrait toujours être défini et au moins avoir un array vide (comme les periods dans ScheduleAttributes). Il faudrait plutôt fixer les endroits où les trips peuvent être `undefined`. Et si le cas est réellement valide, change le type pour rendre optionnel les `trips`. Si c'est pendant les db read, c'est assez localisé, donc simple à modifier sur place.

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
...
+    const usesRoutedDuration = routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH);
 
-    const noDwellTimeDurationSeconds =
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? routedDurationSeconds
-            : segmentDistanceMeters / runningSpeedMps;
+    // Catch zero/negative OSRM durations early to avoid dividing by zero below
+    if (usesRoutedDuration && routedDurationSeconds <= 0) {
+        throw new TrError(
+            'Error trying to generate a path geography. OSRM returned zero or negative duration for segment.',
```
greenscientist: Do we know why OSRM returned zero or negative? Negative in particular seem impossible?

packages/transition-backend/src/models/capnpCache/transitLines.cache.queries.ts
```diff
@@ -197,14 +197,12 @@ const exportParser = function (object: Line, cacheObject: CacheObjectClass) {
             periodCache.setNumberOfUnits(period.number_of_units || -1);
             periodCache.setIsFrozen(boolToInt8(period.is_frozen));
 
-            if (_isBlank(period.trips)) {
-                period.trips = [];
-            }
+            const trips = period.trips ?? [];
```
greenscientist: Should we just return (probably with a console.log) when period.trips is empty ?

packages/transition-backend/src/services/gtfsImport/ScheduleImporter.ts
```diff
@@ -359,7 +359,11 @@ const generateExactTrips = (schedule: Schedule, periods: TripAndStopTimes[][], i
                 // TODO: We ignore the block_id during import, so we set it to null here. When we do something with the block,
             };
 
-            (schedule as Schedule).attributes.periods[periodIndex].trips.push(newTrip);
+            const period = (schedule as Schedule).attributes.periods[periodIndex];
+            if (!period.trips) {
```
greenscientist: If feels like we should always initialize the period object with en empty trip.
(Do we need a class with a proper constructor for that??)

packages/transition-common/src/services/schedules/Schedule.ts
greenscientist: Have an optional field here makes it a burden for the user to do a test everytime for the object.

packages/transition-common/src/services/schedules/Schedule.ts
tahini: @thomasspina actually, I think you are right to round. Would that fix #1915 by any chance?

tahini: @thomasspina @greenscientist got the issue and confirms that the PR with the rounding fixed it, so you should revert to the rounding version here

packages/transition-common/src/services/path/PathGeographyGenerator.ts
iTzxMigz: I know it was a commentary from code Rabbit but zero could happen in theory if 2 nodes are at the same coordinate then the time would be 0. For the negative part I don't know I will juste remove it. It doesn't seem to be possible.

iTzxMigz:
@tahini @greenscientist 

I change the way the trip are handle. The optional was there because 3 branches initialise it with no trips so now we initialize it with empty trips [] (you can see the commit to see where those were not initialize).

I also add a guard that was missing for `calculatedSegmentDurationSeconds` I change it to `!IsFinite` instead of `NaN` to cover the case of Inifinity.

I modify the guard from `usesRoutedDuration && routedDurationSeconds <= 0` to `usesRoutedDuration && routedDurationSeconds === 0`

packages/transition-common/src/services/schedules/Schedule.ts
tahini: @iTzxMigz Il manquerait juste de documenter que `tripArrivalTmeSeconds` doit être des entiers. Pour éviter de faire ça à chaque round, pouvez-vous juste ajouter cette info dans la définition du `SchedulePeriodTrip` en haut de fichier, pour dire autant au arrival et departure times `/** Array of arrival times at each stops, in seconds since midnight. Values are expected to be integers */`.

tahini (approved):
Ça me va, juste un commentaire à ajouter dans le type des trips (tant qu'à..., étant donné qu'un bug a été introduit, probablement par manque d'information).

@greenscientist Prêt à re-reviewer

packages/transition-common/src/services/path/PathGeographyGenerator.ts
```diff
@@ -139,15 +139,26 @@ const calculateSegmentDuration = (
...
-        routingEngine === 'engine' || _isBlank(defaultRunningSpeedKmH)
-            ? routedDurationSeconds
-            : segmentDistanceMeters / runningSpeedMps;
+    // Coincident nodes/waypoints produce a zero-duration leg; catch it before dividing by zero below
+    if (usesRoutedDuration && routedDurationSeconds === 0) {
+        throw new TrError(
+            'Error trying to generate a path geography. OSRM returned a zero duration for segment.',
+            'PUPDGEO0002',
+            'TransitPathCannotUpdateGeographyBecauseErrorCalculatingSegmentDuration'
+        );
```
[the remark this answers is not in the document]
greenscientist: @thomasspina @iTzxMigz check this

greenscientist (approved):
I said approve, but please update the exception ID (see code rabbit comment)

PR #1905, opened 2026-04-15
Warn user when a schedule fails to regenerate after a path edit

### LINKED ISSUE #852: schedules: Warn the user if the schedule cannot be updated
For exact schedules imported from GTFS, there is no oubound/inbound, frequencies or number of units. It is thus not possible to re-generate schedules for those paths, when the path is updated for example. Just taking the current trips and updating the stop times is not a good idea, as the user cannot know which trips were using the updated paths, so cannot really understand the result.

Instead, we should just warn the user that trips using this path for service X cannot be updated and they should manually change it.

iTzxMigz:
Fixes #852 

When a path is edited and the user chooses to update affected schedules, Schedule.updateForAllPeriods used to swallow the errors returned by generateForPeriodFunction, leaving the user with no feedback that some periods (e.g. those without number of vehicles or interval set) did not actually get any trips generated.

Propagate per-service regeneration counts up through Line.updateSchedules and surface a warning modal listing every service whose schedule could not be regenerated at all. The modal points the user to the timetable section of the line to set the missing configuration.

We can see the pop up here:
Screenshot 2026-04-14 at 4 45 04 PM

## Summary by CodeRabbit

* **New Features**
  * Added English and French UI text explaining schedule-generation failures.
  * UI now shows a modal listing services whose schedules failed to regenerate.
  * Schedule period form auto-applies sensible inbound/outbound path choices.

* **Behavior Changes**
  * Regeneration process reports which services failed; UI retains path selection on failures and clears it when none.

* **Tests**
  * Expanded tests covering path association and schedule regeneration outcomes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chairemobilite/transition/pull/1905)

### DIFF
--- a/packages/transition-common/src/services/line/Line.ts
+++ b/packages/transition-common/src/services/line/Line.ts
@@ -392,12 +392,22 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable 
         delete this.attributes.scheduleByServiceId[serviceId];
     }
 
-    // update all schedules that uses the path id
-    async updateSchedulesForPathId(pathId: string, saveSchedules = false) {
+    /**
+     * Updates all schedules that use the given path id, regenerating their trips.
+     *
+     * @param pathId - The id of the path whose associated schedules must be updated
+     * @param saveSchedules - When true, persist each updated schedule via the socket
+     *                        event manager
+     * @returns The ids of the services whose schedule had generation problems
+     *          and should be raised to the user. See `updateForAllPeriods` in
+     *          `services/schedules/Schedule.ts` for the exact failure conditions.
+     */
+    async updateSchedulesForPathId(pathId: string, saveSchedules = false): Promise<string[]> {
         const associatedServiceIds = this.getScheduleServiceIdsForPathId(pathId);
         if (associatedServiceIds.length > 0) {
-            await this.updateSchedules(associatedServiceIds, saveSchedules);
+            return this.updateSchedules(associatedServiceIds, saveSchedules);
         }
+        return [];
     }
 
     // source: https://stackoverflow.com/a/41491220
@@ -427,17 +437,32 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable 
         return L > threshold ? darkColor : lightColor;
     }
 
-    async updateSchedules(serviceIds: string[] = [], saveSchedules = false) {
+    /**
+     * Regenerates trips for the given services' schedules. When no service id
+     * is provided, every schedule attached to the line is regenerated.
+     *
+     * @param serviceIds - The ids of the services to update. Pass an empty array
+     *                     (default) to update every schedule on the line.
+     * @param saveSchedules - When true, persist each updated schedule via the
+     *                        socket event manager
+     * @returns The ids of the services whose schedule had generation problems
+     *          and should be raised to the user. See `updateForAllPeriods` in
+     *          `services/schedules/Schedule.ts` for the exact failure conditions.
+     */
+    async updateSchedules(serviceIds: string[] = [], saveSchedules = false): Promise<string[]> {
         // update all services if serviceIds is empty
         const schedulesObjectsByServiceId = this.getSchedules();
         if (serviceIds.length === 0) {
             serviceIds = Object.keys(schedulesObjectsByServiceId);
         }
         const savePromises: Promise<unknown>[] = [];
+        const serviceIdsWithFailedScheduleGeneration: string[] = [];
         for (const serviceId in schedulesObjectsByServiceId) {
             if (serviceIds.includes(serviceId)) {
                 const schedule = schedulesObjectsByServiceId[serviceId];
-                schedule.updateForAllPeriods();
+                if (!schedule.updateForAllPeriods()) {
+                    serviceIdsWithFailedScheduleGeneration.push(serviceId);
+                }
                 if (saveSchedules) {
                     savePromises.push(schedule.save(serviceLocator.socketEventManager));
                 }
@@ -445,6 +470,7 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable 
             }
         }
         await Promise.all(savePromises);
+        return serviceIdsWithFailedScheduleGeneration;
     }
 
     // TODO This function does not seem to be called anywhere
--- a/packages/transition-common/src/services/schedules/Schedule.ts
+++ b/packages/transition-common/src/services/schedules/Schedule.ts
@@ -1065,7 +1065,19 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
         const periods = this.attributes.periods;
         for (let i = 0, countI = periods.length; i < countI; i++) {
             const period = periods[i];
-            period.trips.forEach((trip) => (associatedPathIds[trip.path_id] = true));
+            // Include the paths the period is configured to use, even if no trips
+            // have been generated yet. Otherwise a service that has not produced
+            // any trips would be considered unrelated to its configured paths and
+            // would silently be skipped by callers like updateSchedulesForPathId.
+            if (period.outbound_path_id) {
+                associatedPathIds[period.outbound_path_id] = true;
+            }
+            if (period.inbound_path_id) {
+                associatedPathIds[period.inbound_path_id] = true;
+            }
+            // period.trips can be undefined when the period has never been generated
+            // (no interval/units configured).
+            period.trips?.forEach((trip) => (associatedPathIds[trip.path_id] = true));
         }
 
         return Object.keys(associatedPathIds);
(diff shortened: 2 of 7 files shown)


packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -83,6 +83,13 @@ export interface ScheduleAttributes extends GenericAttributes {
     periods: SchedulePeriod[];
 }
 
+/** Result of regenerating trips for every period of a single service's schedule. */
+export type ScheduleRegenerationFailure = {
```
tahini: Ce nom de type est un peu misleading. C'est juste `ScheduleGenerationResult` (pas besoin de [re] à génération puisque cette fonction est utilisée aussi pour les créer, et ça ne contient pas que des Failures)

packages/transition-common/src/services/line/Line.ts
```diff
@@ -427,24 +429,29 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable
         return L > threshold ? darkColor : lightColor;
     }
 
-    async updateSchedules(serviceIds: string[] = [], saveSchedules = false) {
+    async updateSchedules(serviceIds: string[] = [], saveSchedules = false): Promise<ScheduleRegenerationFailure[]> {
```
tahini: Ici, vous pourriez retourner un autre type plus approprié à cette fonction: par exemple, pour chaque service, un status ok ou fail à la mise à jour

packages/transition-frontend/src/components/forms/path/TransitPathEdit.tsx
```diff
@@ -209,13 +212,31 @@ class TransitPathEdit extends SaveableObjectForm<Path, PathFormProps, PathFormSt
...
         serviceLocator.eventManager.emit('progress', { name: 'SavingLine', progress: 0.0 });
-        await line.updateSchedulesForPathId(path.getId(), true);
+        const failures = await line.updateSchedulesForPathId(path.getId(), true);
         line.refreshPaths();
         serviceLocator.eventManager.emit('progress', { name: 'SavingLine', progress: 1.0 });
         this.closeSchedulesAffectedConfirmModal(e);
+        // Only surface services where *every* period failed to regenerate. If at least one
+        // period succeeded, the service already has some useful schedule data, and we don't
+        // want to flag it as broken in the UI.
+        const fullyFailedServices = (failures || []).filter(
```
tahini: Yes we do want to flag it as broken! Typically, whenever we have a single trip for which the number of stop nodes does not match the number of nodes in the path, then trRouting completely breaks and does not start, resulting in the complete failure of all calculations in Transition forever and ever until the users figure out which path caused the problem... So yes, we want to see partial failures please!

packages/transition-frontend/src/components/forms/path/TransitPathEdit.tsx
```diff
@@ -840,6 +861,38 @@ class TransitPathEdit extends SaveableObjectForm<Path, PathFormProps, PathFormSt
                                 closeModal={this.closeDeleteConfirmModal}
                             />
                         )}
+                        {this.state.scheduleRegenerationFailures.length > 0 &&
+                            (() => {
```
tahini: You don't have to put it in a function here, or extract to a function component in this file?

packages/transition-frontend/src/components/forms/schedules/TransitSchedulePeriod.tsx
```diff
@@ -84,6 +84,19 @@ const TransitSchedulePeriod: React.FC<TransitSchedulePeriodProps> = (props) => {
     const inboundPathId =
         actualInboundPathId || (inboundPathsChoices.length === 1 ? inboundPathsChoices[0].value : undefined);
 
+    // Persist the displayed outbound/inbound path id on the schedule period.
```
tahini: Misleading comment, it does not persist the displayed path value, it updates the period object when the only possibility is not the currently selected path ID.

Aussi, c'est plus ou moins lié à l'objectif de cette PR, ça aurait pu être dans un commit séparé (dans la même PR)

packages/transition-frontend/src/components/forms/schedules/TransitSchedulePeriod.tsx
```diff
@@ -84,6 +84,19 @@ const TransitSchedulePeriod: React.FC<TransitSchedulePeriodProps> = (props) => {
...
+            onValueChange(`periods[${periodIndex}].outbound_path_id`, { value: outboundPathId });
+        }
+    }, [outboundPathId, actualOutboundPathId, periodIndex]);
+
+    React.useEffect(() => {
+        if (inboundPathId && inboundPathId !== actualInboundPathId) {
+            onValueChange(`periods[${periodIndex}].inbound_path_id`, { value: inboundPathId });
+        }
+    }, [inboundPathId, actualInboundPathId, periodIndex]);
+
```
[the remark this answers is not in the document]
tahini: The rabbit's got a point here

tahini (changes requested):

greenscientist:
I'm just wondering, is a modal like that sufficient? Should we block progress elsewhere in the app until this is resolved ? (It might be complicated to do.)  
I just worry that somebody would see the popup, maybe act on 1 or 2 lines, completely forget about which lines need to be fixed. Are the line highlighted somehow in the list?

tahini:
> I'm just wondering, is a modal like that sufficient? Should we block progress elsewhere in the app until this is resolved ? (It might be complicated to do.) I just worry that somebody would see the popup, maybe act on 1 or 2 lines, completely forget about which lines need to be fixed. Are the line highlighted somehow in the list?

It is not, but this PR is just a nice addition to inform the user of what was previously silently failing. It's not in the current scope to go beyond that for now, we should also fix the model so that it does not happen, but we're not there yet.

packages/transition-common/src/services/line/Line.ts
```diff
@@ -427,24 +429,28 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable
...
         if (serviceIds.length === 0) {
             serviceIds = Object.keys(schedulesObjectsByServiceId);
         }
         const savePromises: Promise<unknown>[] = [];
+        const serviceIdsWithFailedScheduleGeneration: string[] = [];
         for (const serviceId in schedulesObjectsByServiceId) {
             if (serviceIds.includes(serviceId)) {
                 const schedule = schedulesObjectsByServiceId[serviceId];
-                schedule.updateForAllPeriods();
+                if (schedule.updateForAllPeriods()) {
```
tahini: Si l'update retourne `true`, on l'ajoute à la liste des failed? C'est un peu contre-intuitif.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1237,13 +1237,22 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
 
         return Status.createOk(result.trips);
     }
-    updateForAllPeriods() {
-        // re-generate (after modifying path by instance)
+    updateForAllPeriods(): boolean {
+        // Returns true if at least one period failed to regenerate.
```
tahini: D'habitude, quand on retourne `true` c'est que tout va bien, il faudrait inverser la valeur de retour (et documenter sa signification dans la jsdoc de la fonction)

packages/transition-common/src/services/line/Line.ts
```diff
@@ -427,24 +429,28 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable
         return L > threshold ? darkColor : lightColor;
     }
 
-    async updateSchedules(serviceIds: string[] = [], saveSchedules = false) {
+    async updateSchedules(serviceIds: string[] = [], saveSchedules = false): Promise<string[]> {
```
tahini, line 452: svp ajouter un bloc de documentation sur cette fonction pour expliquer la valeur de retour.

tahini (changes requested):
Il faudrait juste changer la signification de la valeur de retour dela fonction `updateForAllPeriods`

iTzxMigz:
J'ai modifié la logique de détection pour pas flaguer à tout moment lorsque par exemple y'a pas de trip à la base dans une des périodes et que l'utilisateur n'en veut pas.

Donc, voici la logique qui va sortir le pop up d'erreur de génération:

- Si aucune des périodes n'a ``d'interval_seconds`` , ``inbound_interval_seconds`` ou ``number_of_units`` set (c'est comme un service fantôme, on le flag)
-  Une période qui était censée produire des trips (configurée ou qui en avait avant) échoue à régénérer
- Un service qui se ramasse avec 0 trip de généré au total à la fin.

J'ai aussi ajouté des tests pour la logique de détection

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1237,13 +1249,66 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
...
+                continue;
+            }
             // TODO period_shortname can be undefined, fix typing to avoid this or add check
-            this.generateForPeriodFunction(periods[i].period_shortname as string);
+            try {
+                const result = this.generateForPeriodFunction(period.period_shortname as string);
+                if (Status.isStatusError(result)) {
+                    success = false;
+                }
+            } catch {
```
tahini: Ici ça pourrait être intéressant de logger l'erreur au lieu de silencieusement laisser passer.

tahini (approved):
Un petit commentaire pour ajouter des logs d'erreur, mais sinon, c'est bon.

packages/transition-common/src/services/line/Line.ts
```diff
@@ -427,24 +436,39 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable
...
-    async updateSchedules(serviceIds: string[] = [], saveSchedules = false) {
+    /**
+     * Regenerates trips for the given services' schedules. When no service id
+     * is provided, every schedule attached to the line is regenerated.
+     *
+     * @param serviceIds - The ids of the services to update. Pass an empty array
+     *                     (default) to update every schedule on the line.
+     * @param saveSchedules - When true, persist each updated schedule via the
+     *                        socket event manager
+     * @returns The ids of the services whose schedule should be flagged to the
```
greenscientist: Feels like we should say something like 

ids of the services with errors/problems that should be flagged (raised?) to the user. 

Just flagged to the user it's a bit too vague. 
(I know you mention the reference, but that.s too far. Reference should probably also mention the file)

packages/transition-common/src/services/line/Line.ts
```diff
@@ -392,12 +392,21 @@ export class Line extends ObjectWithHistory<LineAttributes> implements Saveable
...
+     * @param saveSchedules - When true, persist each updated schedule via the socket
+     *                        event manager
+     * @returns The ids of the services whose schedule should be flagged to the user
+     *          (see Schedule.updateForAllPeriods for the flagging conditions)
+     */
+    async updateSchedulesForPathId(pathId: string, saveSchedules = false): Promise<string[]> {
         const associatedServiceIds = this.getScheduleServiceIdsForPathId(pathId);
         if (associatedServiceIds.length > 0) {
-            await this.updateSchedules(associatedServiceIds, saveSchedules);
+            return this.updateSchedules(associatedServiceIds, saveSchedules);
```
greenscientist, line 408: Just flagging this await to return change, I did not see a comment about this in the commit message, making use it's right and intended.

packages/transition-common/src/services/schedules/Schedule.ts
```diff
@@ -1237,13 +1249,66 @@ class Schedule extends ObjectWithHistory<ScheduleAttributes> implements Saveable
...
+    }
+
+    /**
+     * Generates trips for every period of the schedule. Periods that are
+     * intentionally empty (no configuration and no existing trips) are skipped
+     * silently and do not affect the result.
+     *
+     * @returns True when the schedule generated successfully (no flag for the user).
+     *          False when the schedule should be flagged for these reasons:
+     *          - the schedule has no period at all, or no period is configured
```
greenscientist, line 1277: Should the reason be in a ENUM type, so that we return actual precise information?

packages/transition-common/src/services/line/Line.ts
iTzxMigz: Yes it is intended, we want to propagate to ``TransitPathEdit.tsx`` the values of the service Id that fails this is why we use a return instead of an await since now this fonction return the service Id that fails. The return is catch here on ``line 247`` in ``TransitPathEdit.tsx``
```typescript
const failedServiceIds = await line.updateSchedulesForPathId(path.getId(), true);```

packages/transition-common/src/services/schedules/Schedule.ts
iTzxMigz: We could but we don't need that information for the frontend user. The failed service is basically always because `interval_seconds`, `inbound_interval_seconds` or `number_of_units` is unset or set incorrectly. So the message is always the same it's just that we don't want to flag it all the time.

By "all the time" I mean we don't want to flag the schedule just because some periods are empty. Sometimes, periods could be intentionally empty (no config + no prior trips) because the line doesn't run during that period for that service. We only flag when the schedule as a whole fails to produce a usable result, and the message to the user is the same regardless of which of the three cases triggered it.

If you want, I could still do it and regroup by reason why each service has been flagged on the frontend if you think it could still be useful information for the user even though they will need to act the same way to resolve the issue. Just let me know!

greenscientist (approved):
Seems fine,

I did not check the logic in details, I assume that tahini did that part.

PR #1916, opened 2026-04-20
gtfs export: allow to filter exported elements by service

### LINKED ISSUE #1808: gtfs export: Add button to exclude Transition specific data
We add several field specific to Transitioin in the exported GTFS. Its useful when reimporting data in transition later, but it would be nice to have the option to produce a "Clean" version of the GTFS with only the official fields. 

From the validator, unknown_column:
"agency.txt"	"tr_agency_color"	9
"agency.txt"	"tr_agency_description"	10
"calendar.txt"	"tr_service_desc"	11
"calendar.txt"	"tr_service_color"	12
"routes.txt"	"tr_route_internal_id"	13
"routes.txt"	"tr_route_row_category"	14
"routes.txt"	"tr_is_autonomous"	15
"routes.txt"	"tr_allow_same_route_transfers"	16
"shapes.txt"	"tr_shape_routing_mode"	6
"shapes.txt"	"tr_shape_routing_engine"	7
"stops.txt"	"tr_node_color"	15
"stops.txt"	"tr_routing_radius_meters"	16
"stops.txt"	"tr_default_dwell_time_seconds"	17
"stops.txt"	"tr_can_be_used_as_terminal"	18

### LINKED ISSUE #1913: gtfs export: supporting exporting for specific services only
When agencies/lines start to have many possible schedules, like for the eovlutionary algorithm, it may be useful to filter the services to export  in the GTFS

tahini:
fixes #1913 and fixes #1808

Allow to filter the elements to export in the GTFS by agencies or services. The lists are updated as the user selects new elements

Also, add an option to add the custom transition fields to the exported GTFS feed, so that by default, the exported archive is "clean", ie it does not contain any additional non-spec fields.

## Summary by CodeRabbit

* **New Features**
  * Added checkbox to include Transition custom fields in GTFS exports.
  * Added service-based selection with dynamic agency/service filtering; services list appears when relevant.

* **Bug Fixes**
  * Fixed grammar in agency selection label.
  * Clarified validation message to require selecting at least one exportable element.

* **Documentation**
  * Updated English and French translations for new export UI strings.


packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,6 +70,50 @@ const writeZipFile = async (
     });
 };
 
+type GtfsExportOptions = Pick<GtfsExportParameters, 'selectedAgencies' | 'selectedServices'>;
```
greenscientist: This feel more like  a FilterOptions and not an ExportOptions

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -96,8 +143,10 @@ export const exportGtfs = async (
         let currentStepCompleted = 0;
         fileManager.directoryManager.createDirectoryIfNotExistsAbsolute(gtfsFileDirectory);
 
+        const { agencyIds, serviceIds, lineIds } = await filterDataToExport(gtfsExportOptions);
```
greenscientist: Don't think the name filterData here is write. We are GatheringData to export, using a filter, but not really filtering data that we already have.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -128,8 +176,8 @@ export const exportGtfs = async (
             progress: (currentStepCompleted / nbExportSteps).toFixed(2)
         });
 
-        // Export the services used by the agencies' lines
-        const serviceIdsUsed = lineExportResponse.serviceIds;
+        // Export the services used by the agencies' lines, or only those selected if any
+        const serviceIdsUsed = serviceIds.length === 0 ? lineExportResponse.serviceIds : serviceIds;
```
greenscientist: It should not have been properly filed by the "data gathering" function earlier?

packages/transition-common/src/api/gtfs.ts
```diff
@@ -17,6 +17,10 @@ export interface GtfsExportParameters {
      * @memberof GtfsExportParameters
      */
     selectedAgencies: string[];
+    /**
+     * IDs of the services to export
```
greenscientist: Should add "Leave empty for all"

packages/transition-backend/src/api/__tests__/gtfs.routes.test.ts
```diff
@@ -74,10 +74,11 @@ describe('GTFS data import', () => {
 
 describe('GTFS data export', () => {
 
-    const exportData = {
+    const exportData: GtfsExportParameters = {
         gtfsExporterId: 'arbitraryId',
         selectedAgencies: ['agency1Id', 'agency2Id'],
         selectedServices: ['service1Id', 'service2Id'],
+        includeCustomFields: true,
```
greenscientist: Can we name this includeTransitionCustomField ?

if we just say custom field, I have the impression that it's data that the user have added himself.

greenscientist (commented):
(if did not check the frontend components)

packages/transition-frontend/src/components/forms/gtfs/GtfsExportForm.tsx
```diff
@@ -75,17 +94,77 @@ class GtfsExportForm extends ChangeEventsForm<WithTranslation, ChangeEventsState
...
+            if (Array.isArray(newValue.value) && newValue.value.length > 0) {
+                // Get the line ids of the selected agencies
+                const selectedLineIds =
+                    newValue.value.map((agencyId) => agencies.getById(agencyId)?.getLineIds()).flat() ?? [];
+                this.setState({
+                    servicesToDisplay: services
+                        .getFeatures()
+                        .filter((service) =>
+                            service.attributes.scheduled_lines.some((lineId) => selectedLineIds.includes(lineId))
+                        )
```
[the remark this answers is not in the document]
tahini: At this point, the scheduled_lines should not be undefined, it should be at least an empty array

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
tahini: Not if we did not specify services to export, we'll get them from the lines themselves as we load them, so only and all the used services will be exported. The data gatherer is a filtered gatherer, gatherer from filter, whatever...

packages/transition-common/src/api/gtfs.ts
tahini: There needs to be at least a service or an agency, so "leave empty for all" is not quite right either, but I'll add something.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
gvivero-mo: ✅ Fixed in [`6dc7486`](https://github.com/sonatype/seaworthy/commit/6dc7486b03c887a65fc18102b81d7f67699c0d8f) — two changes:
1. Renamed test display name to `"when removeOrganizationMembers throws (best-effort), telemetry sent with auth0Reverted=true and null errorMessage"` to accurately reflect the inner-catch path.
2. Added a new test `onOuterException_sendsTelemetryWithErrorMessage` that covers the outer catch block — mocks `getPrimaryAuth0Organization` to throw, then asserts `auth0Reverted=false` and `errorMessage` is populated with the exception message.

tahini: In the last commit, we add the `includeTransitionCustomFields` options, which is not a filter option, so these are `ExportOptions`, and they include a filter.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
tahini: So for sake of coherence, the data getter will also filter the services when only agencies are set, so the individual exporter don't have to worry about that. There's only one function taht gets the elements to export.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
greenscientist: Do we need to put all of the param in the same struct ??

tahini: > Do we need to put all of the param in the same struct ??

We don't _need_ to, but we do... We could structure it differently, with `selectedAgencies` and `selectedServices` under a single `filter` property. We may do that if/when we add a lines filter as well and/or have more than the custom field options. But for now, I'm ok with this.

kaligrafy (approved):
Je vous laisse discuter des éléments params, mais sinon c'est bon!

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+// Filter agencies and services to export based on the selected options.
+const getDataToExport = async (
+    gtfsExportOptions: GtfsExportOptions
+): Promise<{
+    agencyIds: string[];
+    serviceIds: string[];
+    lineIds: string[];
+}> => {
+    // Load agency collection and filter the selected agencies if a list if set
+    const agencies = await agenciesDbQueries.collection();
```
greenscientist: We don't have a db query with the filter already ? Would be more efficient to only get the right agencies directly than filtering them after the fact. 
But a big deal for agencies but more important for lines and other, but we should have a consistant pattern for all data type.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+        gtfsExportOptions.selectedAgencies.length > 0
+            ? agencies.filter((agency) => gtfsExportOptions.selectedAgencies.includes(agency.id))
+            : agencies;
+
+    // Get the lines of the selected agencies, a line only belongs to one agency, so no need to check for duplicates
+    const lineIdsFromAgencies = filteredAgencies.flatMap((agency) => agency.line_ids ?? []);
+
+    // Get all services, or only the selection, with their scheduled lines,
+    const services = await servicesDbQueries.collection({
+        serviceIds: gtfsExportOptions.selectedServices.length === 0 ? undefined : gtfsExportOptions.selectedServices
```
greenscientist: We cannot pass the selectedServices directly here ? Feel like we are doing an if that should be done inside the function anyway.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+    });
+    // Filter the services to export to those that have at least one line in the selected agencies, if any.
```
greenscientist: Maybe Reformulate to 
Export only the services that have at least one line in the selected agencies

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+    const serviceIdsToExport = services
+        .filter((service) => (service.scheduled_lines ?? []).some((lineId) => lineIdsFromAgencies.includes(lineId)))
+        .map((service) => service.id!);
+
+    // If there was no specific service selected, return the agencies and services filtered only by the selected agencies.
```
greenscientist: This is starting to feel that we have lot of edge cases. Wondering if this should go into a more common data manager than only live in the gtfs exporter. 
Thinking maybe of capnp file export for the genetic algoritm

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+    if (gtfsExportOptions.selectedServices.length === 0) {
```
greenscientist: Not a big fan of an if with the return in the middle of the function. 
Maybe if should be done earlier and then call to 2 different fonctions

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
```diff
@@ -67,26 +70,84 @@ const writeZipFile = async (
...
+ * @param {string} exportAbsoluteDirectory Directory where to put the files to
  * export
  * @param {EventEmitter} [progressEmitter] Optional event emitter to send
  * progress to
  * @return {*} Return the relative file path where the zip file was generated
  */
 export const exportGtfs = async (
-    selectedAgencyIds: string[],
+    gtfsExportOptions: GtfsExportOptions,
     exportAbsoluteDirectory: string,
```
greenscientist, line 190: if you create a gtfsExportOptions, why is the exportAbsDir not in it?
(I dont like big struct, but why not put it in ?  Unless the name of the struct is not right.  (I'm bringing back the idea of datafilteroptions :) )

greenscientist (changes requested):
Adding a request changes, I think there's a few structural changes that could be done. You can always argue the opposite.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
tahini: > We don't have a db query with the filter already ? Would be more efficient to only get the right agencies directly than filtering them after the fact. But a big deal for agencies but more important for lines and other, but we should have a consistant pattern for all data type.

No we don't, if we did i'd use it! I had started work on #1415 a year ago to filter on scenario, but we did not agree on a way ahead with filtered queries, so it's still there (and wouldn't have helped here anyway). I can add a fixme to filter at query level when available.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
tahini: > We cannot pass the selectedServices directly here ? Feel like we are doing an if that should be done inside the function anyway.

In the query, an empty array means no record, while undefined means all. Here, an empty array means all (no specific selection), so we need to map to undefined.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
tahini: Made 2 functions after the agencies/line load for the case with/without service selection

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
greenscientist: I guess this use case will help the discussion on that. I guess this is what I was remembering.

packages/transition-backend/src/services/gtfsExport/GtfsExporter.ts
greenscientist: ok, this might required a comment as someone who don't have all the interface in memory might not know this subtlety.

packages/transition-frontend/src/components/forms/gtfs/GtfsExportForm.tsx
```diff
@@ -77,17 +90,94 @@ class GtfsExportForm extends ChangeEventsForm<WithTranslation, ChangeEventsState
         });
     };
 
+    private filterDataForExport(
```
greenscientist, line 93: Feel like this could be some generic component that we reuse in the interface. But that's mostly a note for the future.

greenscientist (approved):
Ok, seems ok. 
Maybe just add some of the mentionned comments, but those are minor

PR #1928, opened 2026-04-29
Revert "fix missing translation in checkbox label"

tahini:
This reverts commit b8af814af070bbc041834cbc066b6fcefaf2ec24.

The label received as props is often already translated by the calling components, translating it once again just puts a translation string in the main locale prefix. It breaks the previous behavior. The label received in params should already be translated. If not set, then we should fallback to a translated "Yes".


tahini:
The actual bug to fix is #1929 which was not fixed, but introduced bugs in the original reverted commit

kaligrafy (approved):

PR #1930, opened 2026-04-29
db: avoid deleting frozen records at the DB level

tahini:
Theses commits are for the `deleteMultiple` and `delete` default DB functions respectively. It adds a parameter `hasFrozenField` to the delete queries, that are bound by the queries that extend it to either `true` or `false` depending if the corresponding table has a field named `is_frozen`. This will result in frozen records to be impossible to delete from the database. The objects can be unfrozen, then deleted though.

There is currently no parameter to override this (ie to delete frozen records). If we need/want that eventually, we can add an option parameter to the queries, or use different queries, to make sure the intent is correct.

Note that, currently, the frontend already prevents from deleting frozen records: the delete button of frozen objects is disabled and the services and scenarios list, that can be multi-selected and deleted do not allow to select the frozen objects.

But this could change if/when we add other possibilities than delete for mult-selects, in which case we'll want to select frozen records and handle them according to the task, or just for failsafe since many code paths may eventually lead to requesting a deletion. We cannot trust that all UI actions will handle this.

## Summary by CodeRabbit

* **New Features**
  * Frozen items are protected from deletion in the system.
  * Bulk delete operations now report the number of items deleted (deletedCount).

* **Bug Fixes**
  * Deletion flows and notifications updated to respect frozen protection and accurately reflect deletion outcomes.

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
+++ b/packages/chaire-lib-backend/src/models/db/default.db.queries.ts
@@ -408,28 +408,40 @@ export const deleteRecord = async (
  *
  * @param knex The database configuration object
  * @param tableName The name of the table on which to execute the operation
+ * @param hasFrozenField Whether the table has an `is_frozen` field that should
+ * prevent deletion of frozen records
  * @param {string[]|number[]} ids An array of record IDs to delete, numeric for
  * tables that have numeric primary keys, or uuid strings for tables that have
  * uuid primary keys
  * @param options Additional options parameter. `transaction` is an optional
  * transaction of which this delete is part of.
- * @returns The array of deleted IDs
+ * @returns The number of deleted records
  */
 export const deleteMultiple = async (
     knex: Knex,
     tableName: string,
+    hasFrozenField: boolean,
     ids: string[] | number[],
     options: {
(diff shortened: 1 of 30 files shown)


packages/chaire-lib-backend/src/models/db/default.db.queries.ts
```diff
@@ -408,28 +408,40 @@ export const deleteRecord = async (
...
  * @param tableName The name of the table on which to execute the operation
+ * @param hasFrozenField Whether the table has an `is_frozen` field that should
+ * prevent deletion of frozen records
  * @param {string[]|number[]} ids An array of record IDs to delete, numeric for
  * tables that have numeric primary keys, or uuid strings for tables that have
  * uuid primary keys
  * @param options Additional options parameter. `transaction` is an optional
  * transaction of which this delete is part of.
- * @returns The array of deleted IDs
+ * @returns The number of deleted records
```
greenscientist, line 418: Why are we changing the return type ?  
We can return the actual IDs that were deleted with this command.

greenscientist: Semble fesable avec le RETURNING keyword

PR #1932, opened 2026-05-04
network design: save best candidates' simulation results in the main job

tahini:

- Add a cleanup step to generations and candidates, so the job used by the candidates can be kept until the end of the generation instead of being deleted right away. This will allow to save its result at the end.
- Update the batch route file result visitor to accept a job that may not be a batch route job, but a job in which to save the files. It receives also the batch routing parameters necessary to know what results should be generated, as well as an optional suffix to append to the file names.
- The X best candidates of the Y last generation now have their results saved in the main job, and they are listed under the job's files list.

### DIFF
--- a/locales/en/transit.json
+++ b/locales/en/transit.json
@@ -1210,12 +1210,11 @@
             }
         },
         "evolutionaryTransitNetworkDesign": {
-            "files": {
-                "transitDemand": "Demand file: {{filename}}",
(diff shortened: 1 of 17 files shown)


tahini:
Il reste qq tests unitaires à mettre à jour, mais ça fonctionne

kaligrafy (approved):

PR #1949, opened 2026-05-15
auth: add a manual step to confirm user accounts

tahini:
Update the `VerifyPage` such that it loads the email corresponding to the confirmation token when the page loads, but just displays a message telling the user which email the token corresponds to and asks to click on a 'confirm' button.

Add a action in Auth.tsx to get the user data corresponding to the token and return it to the caller.

Add a `GET` route in the backend to retrieve the user information from the token. Protect both the GET and POST routes of the verify to require permissions to run the route.

## Summary by CodeRabbit

* **New Features**
  * Two-step email confirmation UI: token lookup shows associated email and enables confirm action only when ready; expanded status states and clearer messages.

* **Localization**
  * Added English and French confirmation-flow texts (loading, ownership line, ask-to-confirm, missing-token).

* **Security**
  * Account confirmation endpoints now require authorization.

* **Tests**
  * New backend tests covering token lookup and confirmation flows.


packages/chaire-lib-frontend/src/actions/Auth.tsx
```diff
@@ -158,8 +158,39 @@ export const startRegisterWithPassword = (
 
 export type ConfirmData = { token: string };
 export type ConfirmCallbackType = (body?: { status?: string } | null) => void;
+export type ConfirmInfoCallbackType = (body?: { status?: string; email?: string } | null) => void;
+
+export const startGetConfirmUserInfo = (data: ConfirmData, callback?: ConfirmInfoCallbackType) => {
```
tahini: callback probably needs to be mandatory since otherwise, this function is a no-op.

packages/chaire-lib-frontend/src/actions/Auth.tsx
```diff
@@ -158,8 +158,39 @@ export const startRegisterWithPassword = (
 
 export type ConfirmData = { token: string };
 export type ConfirmCallbackType = (body?: { status?: string } | null) => void;
+export type ConfirmInfoCallbackType = (body?: { status?: string; email?: string } | null) => void;
```
tahini: Can `body` really be undefined? I don't think so.

tahini (commented):
self-reviewing, while waiting for review...

greenscientist:
@coderabbitai review

greenscientist (commented):
We probably need to document the full flow somewhere. 

What happens when an admin click on the link and another admin have already accepted the user? Is there a message. I could not decipher that from the code.

packages/chaire-lib-backend/src/api/auth.routes.ts
```diff
@@ -133,19 +134,60 @@ export default <U extends IUserModel>(router: express.Router, authModel: IAuthMo
...
+    router.get('/verify', isAuthorized({ [UserSubject]: ['read', 'update'] }), async (req, res) => {
         try {
+            const token = typeof req.query.token === 'string' ? req.query.token : undefined;
+            if (!token) {
+                return res.status(400).json({
+                    status: 'MissingToken'
+                });
+            }
+
+            const user = await authModel.find({ confirmation_token: token });
```
tahini, line 149: @greenscientist If an admin already verified, the token will not be found here.

packages/chaire-lib-backend/src/api/auth.routes.ts
```diff
@@ -133,19 +134,60 @@ export default <U extends IUserModel>(router: express.Router, authModel: IAuthMo
...
+                });
+            }
+
             let callback: ((user: U) => void) | undefined = undefined;
             if (getConfirmEmailStrategy() === 'confirmByAdmin') {
                 callback = sendConfirmedByAdminEmail;
             }
 
-            const response = await authModel.confirmAccount(req.body.token, callback);
+            const response = await authModel.confirmAccount(token, callback);
```
tahini, line 184: And of course the confirm account will not be confirmed since the token does not exist, and the status will return 'NotFound' here. The behavior remains unchanged from the previous without confirmation behavior. This PR does _not_ add the possibility to unilaterally reject the user, though that could be an interesting additional feature.

greenscientist: @tahini Do we have an issue for that?

packages/chaire-lib-frontend/src/components/pages/VerifyPage.tsx
```diff
@@ -4,56 +4,105 @@ import { useTranslation } from 'react-i18next';
...
+    startGetConfirmUserInfo,
+    ConfirmCallbackType,
+    ConfirmInfoCallbackType
+} from '../../actions/Auth';
 import { RootState } from '../../store/configureStore';
 
-type Status = 'Confirmed' | 'In Progress' | 'NotFound' | 'Error';
+type Status = 'Loading' | 'Ready' | 'Confirming' | 'Confirmed' | 'NotFound' | 'MissingToken' | 'Error';
+
+const getStatusMessage = (status: Status): string => {
```
greenscientist, line 19: On peut presque faire un string appends au lieu d'un long switch case ici!

packages/chaire-lib-frontend/src/components/pages/VerifyPage.tsx
```diff
@@ -4,56 +4,105 @@ import { useTranslation } from 'react-i18next';
...
+                        <strong>{email}</strong>
+                    </p>
+                    <p className="apptr__form__label-standalone">{t('auth:ConfirmationTokenAskToConfirm')}</p>
+                    <p className="apptr__form__label-standalone">
+                        <Button
+                            type="button"
+                            onClick={confirmUser}
+                            label={t('auth:Confirm')}
+                            style={{ display: 'inline-block' }}
+                        />
```
greenscientist, line 101: Should we have an "Ignore" button or something like that to let user not do the confirm ?

greenscientist:
> We probably need to document the full flow somewhere.
> 

Do we create a task to document the whole thing? It should probably include move the path of the Auth config to it's own file that can we included by chaire-lib-common/src/config/shared
/project.config.ts

greenscientist (approved):

kaligrafy:
Prêt à merger? @tahini

PR #1961, opened 2026-06-09
auth: redirect to the originally requested page after login (#1946)

### LINKED ISSUE #1946: auth: login page should redirect to the originally requested page
It seems the LoginPage does not redirect to the page that was requested by the user. For example a verify account link would lead to the login page, but it does not redirect to it afterwards, so the user will have to go click on the link again in the email

kaligrafy:
When an unauthenticated user opened a protected URL (e.g. an account verification link), PrivateRoute sent them to /login but did not remember where they came from, so after logging in they always landed on the home page and had to click the original link again.

The post-login redirect logic already existed: redirectAfterLogin reads location.state.referrer. PrivateRoute now supplies it, capturing the requested path and passing it as the login navigation state, so the user is sent back to the originally requested URL after authenticating.

Adds a unit test for the redirect state and a Playwright UI test reproducing the verify-link scenario, plus a reusable goToUrlTest helper.

closes #1946

Code by Claude Opus 4.8 Low thinking, reviewed by @kaligrafy

## Summary by CodeRabbit

* **New Features**
  * Enhanced login redirect flow to preserve and return users to their originally requested page after authentication, instead of redirecting to the home page.

* **Tests**
  * Added comprehensive test coverage for authentication redirect behavior, including multiple protected URL variants and end-to-end login redirect scenarios.

### DIFF
--- a/packages/chaire-lib-frontend/src/components/routers/__tests__/PrivateRoute.test.tsx
+++ b/packages/chaire-lib-frontend/src/components/routers/__tests__/PrivateRoute.test.tsx
@@ -1,0 +1,54 @@
+/*
+ * Copyright Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import * as React from 'react';
+import { render, screen, cleanup } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { Provider } from 'react-redux';
+import { MemoryRouter, Routes, Route, useLocation } from 'react-router';
+
+import configureStore from '../../../store/configureStore';
+
+// Header pulls in ESM-only deps (react-markdown) that jest does not transform;
+// it is irrelevant to the redirect logic under test, so stub it out.
+jest.mock('../../pageParts', () => ({ Header: () => null }));
+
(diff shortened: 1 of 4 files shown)


packages/chaire-lib-frontend/src/components/routers/__tests__/PrivateRoute.test.tsx
```diff
@@ -0,0 +1,54 @@
...
+// it is irrelevant to the redirect logic under test, so stub it out.
+jest.mock('../../pageParts', () => ({ Header: () => null }));
+
+import PrivateRoute from '../PrivateRoute';
+
+// Login stand-in that exposes the referrer PrivateRoute stored in location.state,
+// so tests can assert the originally requested URL is preserved (issue #1946).
+const LoginStub = () => {
+    const location = useLocation();
+    const referrer = (location.state as { referrer?: string } | null)?.referrer;
```
tahini, line 25: As-tu besoin de caster ici ou location.state ne devrait-il pas déjà connaître le type?

tahini (approved):
Un petit commentaire, mais sinon c'est bon, test UI en prime!

greenscientist (approved):
seems ok

PR #2007, opened 2026-07-24
Add libdrogon for trRouting dependencies

greenscientist:
trRouting now use that library

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the application image setup to include a required runtime library, improving reliability when using routing binaries.


kaligrafy (approved):

greenscientist:
Confirmed tested with new and old version of trrouting

PR #2029, opened 2026-08-06
zones: add a query to return intersecting zones for all data sources

### LINKED ISSUE #2027: chaire-lib: add a query for intersecting zones for all data sources
In the zones db queries, the `getZonesContaining` only returns the zones that include the geography, possibly for a single data source, but if a data source does not cover the requested geography, it won't be present.

There's a use case (the surveys https://github.com/chairemobilite/od_mtl/issues/264) where we want to get results for every zone data source available, even if there is no zone intersecting. The `null` zone is an important information to get if there's no intersection. So there should be an additional query: `getIntersectingZonesByDatasources`, which would take an array of data source IDs as parameter and if unavailable, implies that all data sources are requested.

tahini:
fixes #2027

Add the database query `getIntersectingZonesByDatasources` in the `zones.db.queries` file to get zones intersecting a given geography for many datasets. It takes a geography as first argument, as well as a `dataSourceIds` option to get the zones for all the requested data sources. Leaving the datasource ids blank will return zones for all data zones data sources in the database.

If a zones data source does not have any feature intersecting the geography, it will still return a record with `undefined` values for the zones.

TODO
Add tests:
* With datasource ids
* With all data sources
* With one, many or no zones for a data source
* With non-zone data sources in the DB

### DIFF
--- a/packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
+++ b/packages/chaire-lib-backend/src/models/db/zones.db.queries.ts
@@ -165,6 +165,49 @@ const getZonesContaining = async (
 };
 
 /**
+ * Get the zones intersecting a given geography by data source. If there's no
+ * zone intersecting for a datasource, it will return `null`
+ * @param geography The feature for which to get the intersecting zones
+ * @param options 
+ * @param [options.dsIds=[]] The datasource IDs for which to get the
+ * intersecting zones. Leave empty to get all data sources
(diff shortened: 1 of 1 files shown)


tahini:
I'm leaving this here as draft. Needs formatting and test. It was done while working on https://github.com/chairemobilite/od_mtl/issues/264 but we'll remain in memory for now. In case we need it later, it's here.

PR #2058, opened 2026-09-18
dataSources: move socket routes to chaire-lib-backend package

### LINKED ISSUE #2057: dataSources: move the socket routes (and maybe other functionalities) to chaire-lib
While working on octavi, it seems that the data sources socket routes are in transition-backend and that requires adding a dependency to that package, which requires dependency to rust backend. The data sources classes, queries and functionalities are already in chaire-lib, there already is a dependency to socket-io in chaire-lib-backend for the upload. The data sources socket routes should go there too.

tahini:
fixes #2057

The data sources classes, queries, etc are already in the chaire-lib packages, so the socket routes should also be there.

## Summary by CodeRabbit

- **Refactor**
  - Updated internal module references to use shared implementations consistently across backend services.
  - Improved test module resolution for data source functionality.
  - No user-facing behavior or public interfaces changed.


kaligrafy (approved):

greenscientist:
Is there a missing commit that move the file, I only a change in path reference

tahini:
> Is there a missing commit that move the file, I only a change in path reference

The files are moved in this commit, there isn't much to change

greenscientist (approved):
Ok, l'interface de github ne rend vraiment pas évident le déplacement des fichiers!

